PR #4, opened 2019-04-04
Fix in bsg_mem_1rw_sync_mask_write_bit_synth for Verilator

dpetrisko:


dpetrisko:
High priority PR to fix BlackParrot build

black-parrot:
Hi, I feel like this is cart driving horse. We can not commit this unless the PPA impact is evaluated. There are many rams that are synthesized in our designs, and this could seriously hurt our PPA across all of our designs.

I suggest either we put some ifdef VERILATOR_POS in there, or create a verilator_pos hard target, or .. just fix Verilator.

dpetrisko:
True. There’s an ifdef for verilator, so that’s no problem. Will update tonight

black-parrot:
Great, Dan! Having a regression that evaluates PPA is on our todo list.

dpetrisko:
Fixed for now, with a Verilator ifdef

black-parrot:
Maybe say what Verilator version this fix is for.

PR #22, opened 2019-04-15
Iterative Multiplier with Configurable Stride

Xusine1131:


taylor-bsg:
@tommydcjung @dpetrisko Hi Tommy, can you review Xusine1131's iterative multiplier implementation?  Dan, can you review its suitability for inclusion in BlackParrot (via BaseJump STL), particularly the interface. My thinking is that we can potentially crank up how much work it does per iteration (== more area) so that it computes a multiply in just enough time before the corresponding instruction hits the end of the pipeline (four cycles?). We would only support one active multiply in the pipeline at a time, which I think is enough for most code.

@Xusine1131 can you reference the design doc in the pull request?

dpetrisko:
From a BlackParrot perspective:

1) There are 3 legal sizes of RV multiplication: 16 bit, 32 bit and 64 bit.  Of course, we can sign extend the operands to 64 bit, but that's wasteful for an iterative multiplier.  Perhaps another input which masks the input operands and sign extends the result

2) Integer writeback happens at EX3, so without adding another forwarding path, we would need it to be <=3 cycles. It then conflicts with the D$ which also has 3 cycle latency. If we can get the D$ down to two cycles by repipelining, that would be ideal

taylor-bsg:
Dan,

You may need to extend the integer pipeline if 3 cycles is not reasonable
for a multiply. Presumably this should not be too hard, just another input
to the bypass mux.
A 32-bit multiply is 1.5 cycles, so maybe it will fit in 3.

Regarding conflict with loads, you will never have them simultaneously, so
it should be fine, you would just mux between them.

M

dpetrisko:
It's not difficult, but loads are already the critical path.

taylor-bsg:
Not too worried about it, Ling is creating a PPA regression infrastructure
and we can use that to see where we are in terms of multiplier critical
paths.

Xusine1131:
This is the design doc of iterative multiplier without booth encoding:
https://docs.google.com/document/d/14xXZklWghWiHcFhxNp3a-hCWSfYd7JXTs1lWpK89bO4/edit
This is the design doc of iterative multiplier with radix-4 booth encoding:
https://docs.google.com/document/d/1lmJyuDXDwrE_pksVud3ULazEVF00o9AUybdql94Du3c/edit

If any questions please let me know! Thanks!

bsg_misc/bsg_adder_carry_save.v
```diff
@@ -0,0 +1,28 @@
...
+  ,input [width_p-1:0] opB_i
+  ,input [width_p-1:0] opC_i
+
+  ,output [width_p-1:0] res_o
+  ,output [width_p-1:0] car_o
+);
+  generate
+    for(genvar i = 0; i < width_p; i++) begin : CSA_EACH
+      assign {car_o[i], res_o[i]} = opA_i[i] + opB_i[i] + opC_i[i];
+    end
```
dpetrisko: Minor aesthetic point:
BSG style is to label 
generate begin: foo
end // foo

(Applies throughout)

bsg_misc/bsg_adder_carry_save_4_2.v
```diff
@@ -0,0 +1,39 @@
...
+This is a 4-2 carry save adder used by iterative multiplier ans 4-2 wallace tree.
+***********************************************/
+module bsg_adder_carry_save_4_2 #(parameter integer width_p = "inv")
+(
+   input [width_p-1:0] opA_i
+  ,input [width_p-1:0] opB_i
+  ,input [width_p-1:0] opC_i
+  ,input [width_p-1:0] opD_i
+
+  ,output [width_p:0] outA_o
```
dpetrisko: Just A_o, etc. the suffix indicates the direction already

bsg_misc/bsg_adder_carry_save.v
```diff
@@ -0,0 +1,28 @@
...
+3/14/2019, sqlin16@fudan.edu.cn
+==========================================================
+
+A full adder with configurable width.
+
+
+***********************************************************/
+
+
+module bsg_adder_carry_save #(parameter width_p = "inv")(
```
dpetrisko: parameter integer width_p, throughout

bsg_misc/bsg_adder_carry_save.v
```diff
@@ -0,0 +1,28 @@
+/*********************************************************
+==========================================================
+bsg_adder_carry_save.v
+3/14/2019, sqlin16@fudan.edu.cn
+==========================================================
+
+A full adder with configurable width.
```
dpetrisko: Short description of operation of a carry-save adder (i.e. how does this differ from a normal adder)

bsg_misc/bsg_adder_carry_save_4_2.v
```diff
@@ -0,0 +1,39 @@
...
+  fa2
+  (
+    .opA_i({1'b0,fa1_res_lo})
+    ,.opB_i({fa1_car_lo,1'b0})
+    ,.opC_i({1'b0,opD_i})
+    ,.res_o(outA_o[width_p:0])
+    ,.car_o(outB_o[width_p:1])
+  );
+  assign outB_o[0] = 1'b0;
+endmodule
```
dpetrisko: Minor, all files should end with newline (otherwise some Unix tools choke)

bsg_misc/bsg_adder_wallace_tree.v
```diff
@@ -0,0 +1,72 @@
...
+===================================
+
+This is a wallace balanced tree consisting of 4-2 CSA Adders. 
+And now only the iterative multiplier use this module.
+
+******************************/
+module bsg_adder_wallace_tree #(
+  parameter width_p = 32
+  ,parameter iter_step_p = 2
+  ,parameter max_out_size_lp = `BSG_SAFE_CLOG2(iter_step_p) + width_p - 1
```
dpetrisko: localparam for generated parameters

bsg_misc/bsg_adder_wallace_tree.v
```diff
@@ -0,0 +1,72 @@
...
+)
+(
+  input [iter_step_p-1:0][width_p - 1:0] op_i
```
dpetrisko: Use consistent spacing for slicing e.g. [iter_step_p-1:0][width_p-1:0]

bsg_misc/bsg_adder_wallace_tree.v
```diff
@@ -0,0 +1,72 @@
...
+      for(genvar i = 0; i < actual_iter_step; ++i) begin: CONNECTING_INPUT
+        if(i < iter_step_p)
+          assign all_wire[actual_iter_step - 2 + i] = op_i[i];
+        else
+          assign all_wire[actual_iter_step - 2 + i] = '0;
+      end
+      for(genvar i = 0; i < actual_iter_step/2 - 1; ++i) begin: GENERATE_CSA
+        if (max_out_size_lp == `BSG_SAFE_CLOG2(actual_iter_step) + width_p - 1)
+          bsg_adder_carry_save_4_2 #(.width_p(max_out_size_lp - (`BSG_SAFE_CLOG2(i + 2) - 1)))
+          //bsg_csa_adder_4_2 #(.width_p(max_out_size_lp))
```
dpetrisko: Get rid of commented param

bsg_misc/bsg_adder_wallace_tree.v
```diff
@@ -0,0 +1,72 @@
...
+          (
+            .opA_i(all_wire[4*i + 5])
+            ,.opB_i(all_wire[4*i + 4])
+            ,.opC_i(all_wire[4*i + 3])
+            ,.opD_i(all_wire[4*i + 2])
+            ,.outA_o(all_wire[2*i])
+            ,.outB_o(all_wire[2*i + 1])
+          );
+        else 
+          //bsg_csa_adder_4_2 #(.width_p(max_out_size_lp - 2*(`BSG_SAFE_CLOG2(i + 2) - 1)))
```
dpetrisko: Get rid of commented param

bsg_misc/bsg_adder_wallace_tree.v
```diff
@@ -0,0 +1,72 @@
...
+bsg_adder_wallace_tree.v
+3/16/2019 sqlin16@fudan.edu.cn
+===================================
+
+This is a wallace balanced tree consisting of 4-2 CSA Adders. 
+And now only the iterative multiplier use this module.
+
+******************************/
+module bsg_adder_wallace_tree #(
+  parameter width_p = 32
```
dpetrisko: Should be "inv"

bsg_misc/bsg_adder_wallace_tree.v
```diff
@@ -0,0 +1,72 @@
...
+  ,parameter iter_step_p = 2
```
dpetrisko: Should be "inv"

bsg_misc/bsg_adder_wallace_tree.v
```diff
@@ -0,0 +1,72 @@
+/******************************
+===================================
+bsg_adder_wallace_tree.v
+3/16/2019 sqlin16@fudan.edu.cn
+===================================
+
```
dpetrisko: Link to design docs here

bsg_misc/bsg_adder_wallace_tree.v
```diff
@@ -0,0 +1,72 @@
...
+This is a wallace balanced tree consisting of 4-2 CSA Adders. 
+And now only the iterative multiplier use this module.
```
dpetrisko: "Used in basejump_stl/bsg_misc/bsg_mul_booth_iter

bsg_misc/bsg_adder_wallace_tree.v
```diff
@@ -0,0 +1,72 @@
...
+        if (max_out_size_lp == `BSG_SAFE_CLOG2(actual_iter_step) + width_p - 1)
+          bsg_adder_carry_save_4_2 #(.width_p(max_out_size_lp - (`BSG_SAFE_CLOG2(i + 2) - 1)))
+          //bsg_csa_adder_4_2 #(.width_p(max_out_size_lp))
+          csa_4_2
+          (
+            .opA_i(all_wire[4*i + 5])
+            ,.opB_i(all_wire[4*i + 4])
+            ,.opC_i(all_wire[4*i + 3])
+            ,.opD_i(all_wire[4*i + 2])
+            ,.outA_o(all_wire[2*i])
```
dpetrisko: ,.outA_o(all_wire[2*i + 1])
,.outB_o(all_wire[2*i])

Just for visual consistency

bsg_misc/bsg_mul_booth_iter.v
```diff
@@ -0,0 +1,355 @@
...
+
+*/
+// define the booth encode
+// sel[0]: whether the bits is not zero
+// sel[1]: how many left shifts are need to do
+// sel[2]: sign
+module bsg_booth_encoder #(
+  parameter width_p = -1
+)(
+  input [2:0] ipt_i
```
dpetrisko: What is ipt?

bsg_misc/bsg_mul_iterative.v
```diff
@@ -0,0 +1,380 @@
...
+  ,input [width_p-1:0] opB_i
+  ,input opB_is_signed_i
+  // result
+  ,output [output_size_lp-1:0] result_o 
+  ,output v_o
+  ,input yumi_i
+  
+);
+
+  localparam integer iter_count_lp = (width_p - 1)/iter_step_p + 1; // calculate how many iterations shoubd be done.
```
dpetrisko: should be done

bsg_misc/bsg_mul_iterative.v
```diff
@@ -0,0 +1,380 @@
...
+bsg_mul_iterative.v
+3/14/2019, sqlin16@fudan.edu.cn
+===================================================
+
+An 32 bits integer interative multiplier, capable of singed & unsigned division,
+with configurable stide for each iteration.
+
+States:
+
+eIDNE:          Ready for input.
```
dpetrisko: eIDNE -> eIDLE? throughout

bsg_misc/bsg_mul_booth_iter.v
```diff
@@ -0,0 +1,355 @@
...
+  ,.opD_i(csa_acc_opB_r)
+
+  ,.outA_o(csa_optA)
+  ,.outB_o(csa_optB)
+);
+
+always_comb begin
+  unique case(current_state_r)
+    eSIG: begin
+      csa_opA = {width_p{opA_is_unsigned}} & opB_r;
```
dpetrisko: Maybe a comment explaining this line, it took a while to parse

bsg_misc/bsg_mul_booth_iter.v
```diff
@@ -0,0 +1,355 @@
...
+    )enc(
+      .ipt_i(opA_r[2*i+:3])
+      ,.pos_op_i(opB_r)
+      ,.neg_op_i(neg_opB_r)
+      ,.sel_op_o(sel_op_lo[i])
+    );
+  end
+endgenerate
+
+always_ff @(posedge clk_i) begin
```
dpetrisko: Hide behind debug_p

bsg_misc/bsg_mul_booth_iter.v
```diff
@@ -0,0 +1,355 @@
...
+);
+
+logic [wallace_tree_width_lp-1:0] csa_opA;
+logic [wallace_tree_width_lp-1:0] csa_opB;
+
+reg [wallace_tree_width_lp-iter_step_p-1:0] csa_acc_opA_r;
+reg [wallace_tree_width_lp-iter_step_p-1:0] csa_acc_opB_r;
+
+wire [wallace_tree_width_lp-1:0] csa_optA;
+wire [wallace_tree_width_lp-1:0] csa_optB;
```
dpetrisko: What does csa_opt mean?

bsg_misc/bsg_mul_booth_iter.v
```diff
@@ -0,0 +1,355 @@
...
+  end
+  default: begin
+    cpa_opA = '0;
+    cpa_opB = '0;
+    cpa_carry = lowbits_carry_r;
+  end
+endcase
+
+reg [width_p-1:0] result_low_r;
+generate if(iter_step_p != width_p)
```
dpetrisko: Could save a decent number of LoC by having the if statement be within the case statement, since the clock and reset conditions are the same for both cases

bsg_misc/bsg_mul_iterative.v
```diff
@@ -0,0 +1,380 @@
...
+module bsg_mul_iterative #(
+  parameter width_p = -1
+  ,parameter iter_step_p = -1
+  ,parameter full_output_p = 1
+  ,parameter output_size_lp = width_p*(full_output_p+1)
+)
+(
+  input clk_i
+  ,input reset_i
+  // shakehand signal
```
dpetrisko: handshake

bsg_misc/bsg_mul_iterative.v
```diff
@@ -0,0 +1,380 @@
...
+
+  localparam integer iter_count_lp = (width_p - 1)/iter_step_p + 1; // calculate how many iterations shoubd be done.
+  localparam cal_state_length_lp = `BSG_SAFE_CLOG2(iter_count_lp);  // calculate the length of counter used for eCAL state.
+
+  localparam actual_width_lp = iter_count_lp * iter_step_p;         // to align the width_p, which is equal to width_p providing width_p % iter_step_p == 0.
+  
+  initial assert(actual_width_lp == width_p) else $error("Alignment of stride is required for configurable iteration multiplier!");
+
+  typedef enum [2:0] {eIDNE, eCAL, eADJ, eCPA, eDONE} state_e; // FSM states.
+  state_e current_state_r, current_state_n; // curretn state and next state.
```
dpetrisko: current state

bsg_misc/bsg_mul_iterative.v
```diff
@@ -0,0 +1,380 @@
...
+            current_state_n = eDONE;
+          end
+          eDONE:  current_state_n = eDONE;
+          default : current_state_n = eIDNE;
+        endcase
+      end
+  endgenerate
+
+  assign ready_o = current_state_r == eIDNE;
+  // update curren state
```
dpetrisko: current state

bsg_misc/bsg_mul_iterative.v
```diff
@@ -0,0 +1,380 @@
...
+
+  // CSA network output.
+  logic [csa_outout_size_lp:0] csa_sum;    // sum of last CSA
+  logic [csa_outout_size_lp:0] csa_carry;  // carry of last CSA
+
+  // CPA result remnant in last addition, which should be added in next turn.
+
+  // Bits of result which can be determined in this turn are sent to CPA.
+  reg [iter_step_p-1:0] csa_sum_lowpart; //= csa_sum[iter_step_p-1:0];
+  reg [iter_step_p-1:0] csa_carry_lowpart; //= {csa_carry[iter_step_p-1:1],carry_from_cpa_in_eCAL_stage_r};
```
dpetrisko: remove commented equals

bsg_misc/bsg_mul_iterative.v
```diff
@@ -0,0 +1,380 @@
...
+  // Output of wallace tree, which will be passed to a 4-2 CSA where they are added with result_remnant.
+  wire [csa_outout_size_lp -1:0] wallace_tree_opt_1;
+  wire [csa_outout_size_lp -1:0] wallace_tree_opt_2;
+
+  // Determine cpa operands.
+  always_comb begin
+    unique case(current_state_r) 
+      eIDNE: begin
+        cpa_opA = '0;
+        cpa_opB = '0;//opB_i & {actual_width_lp{opA_neg_n}};
```
dpetrisko: remove comment

bsg_misc/bsg_mul_iterative.v
```diff
@@ -0,0 +1,380 @@
...
+    ,.opB_i(csa_opB)
+    ,.opC_i(csa_outout_size_lp'(result_remnant_sum_r))
+    ,.opD_i(csa_outout_size_lp'(result_remnant_carry_r))
+
+    ,.outA_o(csa_sum)
+    ,.outB_o(csa_carry)
+  );
+  // Update the result and remant result.
+  generate 
+    if(actual_width_lp != iter_step_p) begin: FULL
```
dpetrisko: Same comment about injecting if statement

bsg_misc/bsg_mul_iterative.v
```diff
@@ -0,0 +1,380 @@
+`timescale 1ps/1ps
+/*************************************************************
+==================================================
+bsg_mul_iterative.v
+3/14/2019, sqlin16@fudan.edu.cn
+===================================================
+
+An 32 bits integer interative multiplier, capable of singed & unsigned division,
```
dpetrisko: signed & unsigned

bsg_misc/bsg_mul_iterative.v
```diff
@@ -0,0 +1,380 @@
+`timescale 1ps/1ps
```
dpetrisko: @proftaylor we shouldn't put timescale in the modules, right?

bsg_misc/bsg_mul_iterative.v
```diff
@@ -0,0 +1,380 @@
...
+full_output_p:  Type of results. A 64bit result will be generated with full_output_p = 1.
+output_size_lp: Length of the results, which is defined as width_p*(full_output_p+1) by default.
+
+TODO:
+  1. Removing unnecessary state eADJ from the half version of multiplier.
+  2. Using booth recode to eliminate prefix 1 from sign extension
+
+************************************************************/
+module bsg_mul_iterative #(
+  parameter width_p = -1
```
dpetrisko: "inv", not -1

bsg_misc/bsg_mul_iterative.v
```diff
@@ -0,0 +1,380 @@
...
+  ,parameter iter_step_p = -1
+  ,parameter full_output_p = 1
+  ,parameter output_size_lp = width_p*(full_output_p+1)
```
dpetrisko: localparam

bsg_misc/bsg_mul_iterative.v
dpetrisko: Should add assertion to width if we've only tested 32 bit configuration

bsg_misc/bsg_mul_iterative.v
```diff
@@ -0,0 +1,380 @@
...
+
+eIDNE:          Ready for input. 
+eCAL:           Add the partial products using CSA. CPA is also used to calculate the lower bits of the result.
+eADJ:           Adjust the high part by adding a -opA if opB is negative.
+eCPA:           Recover the result by CPA.
+eDONE:          Output the result
+
+Parameters:
+
+width_p:        Length of opreands
```
dpetrisko: operands

bsg_misc/bsg_mul_iterative.v
```diff
@@ -0,0 +1,380 @@
...
+iter_step_p:    Stride of each iteration of adding.(How many partial products are added in one iteration)
+full_output_p:  Type of results. A 64bit result will be generated with full_output_p = 1.
+output_size_lp: Length of the results, which is defined as width_p*(full_output_p+1) by default.
```
dpetrisko: full_output seems like the wrong name here, but I can't think of a better one.

bsg_misc/bsg_mul_iterative.v
dpetrisko: Also, this equation takes a bit to parse.

full_output_p ? 2 * width_p : width_p

may be more clear

bsg_misc/bsg_mul_iterative.v
```diff
@@ -0,0 +1,380 @@
...
+  wire reset_internal = reset_i | (yumi_i & current_state_r == eDONE); 
+  // counter used for jumping out of eCAL
+  logic [cal_state_length_lp-1:0] cal_state_counter_r;
+  // counter update
+  always_ff @(posedge clk_i) begin
+    if(reset_internal) cal_state_counter_r <= '0;
+    else cal_state_counter_r <= (current_state_r == eCAL) ? cal_state_counter_r + 1 : cal_state_counter_r;
+  end
+  
+  // FSM. And the full output version and half output version contain different states.
```
dpetrisko: Seems like the only difference is 

eCAL : begin
            if(cal_state_counter_r == '1)
                current_state_n = eADJ;
            else
                current_state_n = eCAL;
          end

which could be folded into the first case's if statement

bsg_misc/bsg_mul_iterative.v
```diff
@@ -0,0 +1,380 @@
...
+        cpa_opB = '0;
+      end
+    endcase
+  end
+
+  generate begin: ADDER_TREE_BLOCK
+    if(iter_step_p == 1) begin: NO_CSA
+      bsg_adder_carry_save #(.width_p(actual_width_lp))
+      csa(
+        .opA_i(opA_r & {actual_width_lp{opB_r[0]}}) // sign extension to 64bit
```
dpetrisko: The sign bit is the LSB? Did I miss that in the docs?

bsg_misc/bsg_mul_iterative.v
```diff
@@ -0,0 +1,380 @@
...
+      default: begin
+        csa_opA = '0;
+        csa_opB = '0;
+      end
+    endcase
+  end
+  // 4-2 CSA
+
+  bsg_adder_carry_save_4_2 #(
+    .width_p(csa_outout_size_lp)
```
dpetrisko: output, not output

throughout

testing/bsg_misc/bsg_mul_booth_iter/bsg_test.v
```diff
@@ -0,0 +1,84 @@
...
+
+  logic [2*width_p-1:0] result_lo;
+  wire [2*width_p-1:0] actual_result = signed_li ? opA_cal * opB_cal : opA_li * opB_li;
+  logic v_lo;
+
+  bsg_mul_booth_iter #(
+    .width_p(width_p)
+    ,.iter_step_p(width_p)
+  ) mul(
+    clk_li
```
dpetrisko: use fully qualified names, not positional arguments

bsg_misc/bsg_adder_wallace_tree.v
Xusine1131: For unsigned number, this parameter provides the maximum size of result, while as for the signed addition, all operands must be sign extended, and it's a waste of CSAs if we still follow this pattern to calculate maximum size. With this parameter, we could control how many CSAs are generated to meet the requirement exactly. I'll put a regarding comment nearby.

bsg_misc/bsg_mul_booth_iter.v
Xusine1131: Bits to encode. I could not find a suitable name so I gave ipt (means input). Now I've changed it to source bits.

bsg_misc/bsg_mul_booth_iter.v
Xusine1131: csa_opt means the output of CSA.

bsg_misc/bsg_mul_booth_iter.v
Xusine1131: I'll check remove these debug codes before I upload my code next time!

bsg_misc/bsg_mul_booth_iter.v
Xusine1131: if I put the judgement in always block, it will be synthesized right? I think this seems to be a waste of resources?

testing/bsg_misc/bsg_mul_booth_iter/bsg_test.v
Xusine1131: I know using full qualified name is the rule but the argument's name is exactly the corresponding qualified name except the suffix, and it seems unnecessary to use qualified names. 
I'll correct my code anyway!

bsg_misc/bsg_mul_iterative.v
Xusine1131: Sorry!  At first I used Vivado as my environment and this was created automatically. I have removed all timescales in my module.

bsg_misc/bsg_mul_iterative.v
Xusine1131: I'm sorry for this misleading comment. It was a TODO which denoted a sign extension should be done here, but finally I found another method to prevent extension but I forgot to remove it.

bsg_misc/bsg_mul_iterative.v
Xusine1131: I think full-sized or full-scaled is a better choice?

Xusine1131:
@dpetrisko Thanks for your patience and effort! And I'm so sorry for so many typos and grammar mistakes, I'll pay more attention next time I submit my code!
 I've adjust my code to meet your requirements, but there are several points making me confused and I put my thought under your comment and await your reply. I'll change my code as long as my question is solved.

taylor-bsg:
Thank you Dan — great feedback.

bsg_misc/bsg_mul_iterative.v
Xusine1131: I tried to injecting the if statement but an error provided by VCS indicates that result_low_r[actual_width_lp-1:iter_step_p] is an illegal expression because of invalid range.( actual_width_p is equal to iter_step_p)

Xusine1131:
@tanglingshu Thanks for your testing! These errors seem not to appear in the latest version. (I use VCS to compile my design and no error occurs.)
You can try the latest version, and if any problems please let me know!

tanglingshu:
@Xusine1131 May you point me to your latest commit?

Xusine1131:
@tanglingshu 
Here: https://github.com/bespoke-silicon-group/basejump_stl/blob/c8c70efa97f97db6e2cdbca601db6677c4a921b3/bsg_misc/bsg_mul_iterative.v
And I think the line number you provide is wrong? I do find the regarding line "generate begin: ADDER_TREE_BLOCK" in 213 instead of 238.
I'm not sure whether local parameters can be defined at the port list but an I/O port needs this parameter in my design, and VCS seems to accept this syntax. 
Thanks! :)

tanglingshu:
@Xusine1131 Yea, I thought that would be the latest commit. So yea, if that's your latest commit, then the line number would be a little different.
1) The localparam in the port list seems to raise an error on my end. I'll check with someone first before getting back to you on that
2) Even with the latest commit, it complains around (231: Syntax error at or near token 'assign') again, maybe need a begin/end for the FOR loop?
3) Other complaints are same 
213: Syntax error on an obsolete Verilog 2001 construct standalone generate block 'ADDER_TREE_BLOCK'
286: Syntax error on an obsolete Verilog 2001 construct standalone generate block 'RESULT_UPDATE'
349: Syntax error on an obsolete Verilog 2001 construct standalone generate block 'RESULT_O' 
could be related to this
image

I'll double check 1) and 3) with someone first and get back to you on those. They could be tool related.

dpetrisko:
Hi -- localparam in port list is fine for DC; we're using it for BlackParrot.

The issue is generate begin : LABEL.  This syntax is not supported, you cannot label generate blocks.  Instead, you label ifs and fors in the generate.

For instance,
generate
for (...)
begin : rof1
end // rof1

generate 
if ()
begin : fi1
end // fi1

dpetrisko:
@tanglingshu

Which tool / version causes issue on localparam for you?

tanglingshu:
@dpetrisko most likely the same 90nm flow we're using for ee477, but will need to talk to Scott to confirm this. I'm just running some of our basejump stl designs (trying this one out first) through our regression flow. For now, it's showing some errors when synthesizing.

dpetrisko:
Interesting, @stdavids , which DC version is the PPA regression working with? We're using localparams in port lists all over in BP, so I'm guessing it's fine as of some version of DC (whatever is default on our server).

My suspicion for line 231 is that you're running synthesis on a verilog compiler (not SV-2005), which does not support ++i syntax.  Either way, @Xusine1131, it would be good to add a label to the for loop.

Xusine1131:
@dpetrisko Thanks for your feedback, I've removed all tags caused error.
@tanglingshu Hi, can you try this version(https://github.com/bespoke-silicon-group/basejump_stl/pull/22/commits/0c35838a31e6c79eca335cd9b169ed3c5f7efab7) and let know whether any problem exists in my code? Thanks for your effort! :)

tanglingshu:
@Xusine1131 
Hmm.. Few things I'm not so sure about. Let's take a look one of your generate case (at line 213 of bsg_mul_iterative.v)

1) you have "generate begin" and then at the closing part (line 250-251) you have "end endgenerate". I don't think you need the "begin" and "end" (they could just be generate and then endgenerate)
2) In verilog-2005, I think even generate/endgenerate are optional with the use of genvar, so technically you can remove that as well. Just have some like

for (...)
begin : rof1
end // rof1

if ()
begin : fi1
end // fi1

without generate or endgenerate. Here is an example from our very own basejump_stl [bsg_mul_array.v](https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_misc/bsg_mul_array.v) Maybe someone who's expert on verilog can comment on this. But yea, with what you gave me, I still see similar error above. I have a tool expert checking out our tools now. Hopefully what I said is helpful.

Xusine1131:
@tanglingshu Thanks for your feedback! You're right, but I think this is a style problem instead of a logic or syntax error. I can remove them and you can try whether it is related the error as present.

stdavids:
Regarding localparams in the parameter list, @dpetrisko was right we were accidentally using a 2013 version of DC, I fixed the issue and using a 2016 version there are no problems with the localparam in the parameter list. Rasies an interesting question about tool compatibility though...

Regarding generate statements and @tanglingshu notes:

> 1. you have "generate begin" and then at the closing part (line 250-251) you have "end endgenerate". I don't think you need the "begin" and "end" (they could just be generate and then endgenerate)

This is correct and is considered a syntax bug from DC so this needs to be changed. I found this issue in the following files:

bsg_misc/bsg_adder_carry_save.v
bsg_misc/bsg_adder_wallace_tree.v
bsg_misc/bsg_mul_iterative.v

> 2. In verilog-2005, I think even generate/endgenerate are optional with the use of genvar, so technically you can remove that as well. Just have some like

This is also correct but as @Xusine1131 pointed out this is not a syntax error. I don't think the BSG coding guidelines mention the use (or lack there of) of the generate statement. We tend not to use them, there is really no benefit to them and less code is usually better :).

After fixing those, I'm now seeing a port width mismatch inside bsg_adder_wallace_tree on port A_o of the bsg_adder_carry_save_4_2 instances.

Xusine1131:
@stdavids Thanks! I've solved the problem you mentioned! It's my fault to neglect the warning raised by VCS.
@tanglingshu Thanks again! I can use design compiler but I don't have nay PDK at present. I wonder whether any public or generic PDK I can use for synthesis only?

taylor-bsg:
Hi,

You can just do elaboration in DC — no PDK needed. Our bsg_sv2v does this.

Xusine1131:
@taylor-bsg Thanks! I'll check it out.

dpetrisko:
@stdavids are the synthesis issues cleared up now? I'll approve if they are

tanglingshu:
@dpetrisko The synthesis issues are cleared up now. The bsg_mul_iterative.v was successfully synthesized. 
@Xusine1131 What is the smallest stride (iter_step_p) possible? 
What are some of the more common use cases, for PPA purpose

Xusine1131:
@tanglingshu Thanks! The smallest stride is 2 and the maximum is width_p. I think a typical value is width_p/2 or width_p/4.

tanglingshu:
@Xusine1131, the attached spreadsheet shows some of the numbers gathered from your design. The numbers show possible performance limitation based on static timing report, vectorless power, area of your design, and the critical path. There is also a baseline number based on a regular multiplier using (*). As you can see, your design performs much better. You can also use these number to optimize your design if you wish.  
[bsg_mul_iterative_ppa.xlsx](https://github.com/bespoke-silicon-group/basejump_stl/files/3175609/bsg_mul_iterative_ppa.xlsx)

Xusine1131:
@tanglingshu Thanks for your effort! I'll carefully analyze your feedback!

taylor-bsg:
What were your conclusions Xusine?  How many cycles does your 32-bit
multiplier take?  In terms of latency (in ns), frequency, and area, what
are the pareto optimal points across the configurations of your multiplier
and the synthesized one?

M

Xusine1131:
@taylor-bsg I think the best area/power configuration is stride = 8. It takes 8 cycles to perform a multiplication, with frequency equal to 300MHz and the latency is 26.7ns.  
Another configuration is stride = 16. The latency is 24ns and the frequency is 250MHz. But its area is 1/3 larger.
Almost all the critical paths in my design are from state machine to result[31]. So I think the performance is restricted by the carry propagate adder.

tanglingshu:
Hi @Xusine1131 , to provide you with more information. The attached spreadsheet contains PPA information regarding to the current iterative multiplier (bsg_imul_iterative.v) in the basejump_stl repository. You can use this information to compare it to your design. 

Comparing 8-bit multiplication, our bsg_imul_iterative takes around 12 cycles from the start of the calculation to when v_o is asserted. I haven't had the chance to dive deeper into either of the multipliers so it would be interesting for you to compare your design to bsg_imul_iterative. Roughly speaking, the multiplier in our repo seems to perform better from power and area's perspective. But please check if the statement is still true if both multipliers are operating at an equivalent performance standpoint. Good luck!

[bsg_imul_iterative_ppa.xlsx](https://github.com/bespoke-silicon-group/basejump_stl/files/3192643/bsg_imul_iterative_ppa.xlsx)

Xusine1131:
@tanglingshu Thanks for your PPA regression! 
For area, additional components like carry saved adders are included in my design. Also, the result in my design is 16 bit while in bsg_imul_iterative the result is lower/higher 8 bits.  
As for power, I think a more persuasive index is power/area. So it's not quite accurate to compare power without considering area. I have calculate the power per area of bsg_imul_iterative(300MHz, 8bit) and the value is 0.1832uW/um^2, while the power per area in my design(300MHz, 8bit ,stride = 2) is 0.1338.

taylor-bsg:
Hi all,

@tanglingshu <https://github.com/tanglingshu> @Xusine1131
<https://github.com/Xusine1131>

We want to consider throughput, latency (= cycle_time * number of cycles),
and area independently.  Power, we need vectors to really have anything
meaningful to say so we will ignore for now.

In order to accept the code into the code base, we need to show that the
code is on the pareto curve for common configurations -- namely 16, 32, and
64 bit.  This means that, for the area, there is no other design in
BaseJump or through simple synthesis of operators + pipelining, that has
better latency and throughput.   Can you guys work to converge on this?
Ling, you need to provide the data Xusine needs, and Xusine, you need to
iterate on your design and compare.  The 90nm SAED process that link was
using may not be representative, so we are going to give you numbers in
normalized numbers from another commercially representative flow, which we
will not disclose. The numbers will be FO4 for delay, and another
normalized unit for area.  In terms of iterative multiply, it looks like
from  Ling's runs, that a 32-bit multiplier can attain 5ns latency and 30K
um^2 in the fake 90nm node. It can be easily pipelined to attain 1 multiply
per cycle. (Ling, is this multiply with 64-bits of output data?)  So this
is a very aggressive alternative to beat, and may require some iteration.

M

tanglingshu:
@taylor-bsg, @stdavids and I will work to add the "undisclosed" flow into our regression infrastructure. We're aiming to get this done within this week.

To answer your question regarding the PPA numbers for @Xusine1131 's 32-bit iterative multiplier, the inputs are 32-bits and the output is 64-bits. (This is the default setup for his/her design and I didn't modify the design parameter to choose between 32/64 bit output).

In contrast to Xusine's design, it seems that the current iterative multiplier in basejump_stl (bsg_imul_iterative.v) outputs 32-bits, given 32-bits inputs. There is a parameter that selects between outputting the lower 32-bits vs the upper 32-bits. Without further analysis on how both multipliers, I left it up to Xusine to determine the cycle count/throughput between the two designs.

Because one design outputs 64-bits and the other outputs 32-bits, it may be hard to compare area precisely. 

@Xusine1131 if I say full_sized_p=0, 32-bits inputs will result in a 32-bit output. How should that 32-bit output be interpreted (because ideally 32-bits by 32-bits should give 64-bits result). Also, always feel free to ask me for more data, test different iterations of your design, or ask me how the data is interpreted.

Xusine1131:
@tanglingshu If full_sized_p is zero, the output is lower bits of 64 bits result.
And I'm wondering whether DC can tell me the complete critical path instead of its begin and end. I feel it very hard to optimize my design without its synthesis RTL circuit.  If so, please update it! Thanks!

tanglingshu:
@Xusine1131 Sure the DC can tell us the complete critical path, but due to the sensitivity of the information, I'm not sure if I can just provide you the complete timing report. I have gathered some critical path for you in the attached spreadsheet (check tab bsg_mul_iterative_critical_path). Due to currently limited capability of exporting critical path, I was only able to grab a few critical path manualy across timing reports with various data widths. I hope the data is useful even though the naming of the path can be obscure without the synthesized gate-list RTL circuit. 

In general there seems to be two types of critical paths. One goes through the synthesized adder for your addition, and the second goes through the wallace tree. 

Hopefully this information is conducive, good luck!
[bsg_mul_iterative_ppa_updated_5-21.xlsx](https://github.com/bespoke-silicon-group/basejump_stl/files/3201542/bsg_mul_iterative_ppa_updated_5-21.xlsx)

Xusine1131:
@tanglingshu Thanks, I'll check it out!

Xusine1131:
@taylor-bsg Sorry to bother you professor Taylor. 
I wonder whether there are modules using python script to generate sub-modules before synthesizing. If so, how can I pass the parameter defined in v file to script?
For example, to shrink the height of Wallace Tree, a python script is written to generate Wallace Tree for specific number of operands, using 3-2 CSA instead of 4-2 CSA. If the stride in my design is 8, then a 8-capacity Wallace Tree is generated. But this parameter is defined in SystemVerilog so how can I use this parameter to call the script before synthesizing? Thanks!

stdavids:
@Xusine1131 Regarding using python to generate the wallace tree, we do a similar thing with bsg_round_robin_arb inside of the bsg_misc folder that you can look at for a reference. The idea is that the python script should actually generate a bunch of different wallace trees for many capacities you would expect are commonly needed and place them all inside a module that uses an if-statement to select the implementation based on the capacity. Ideally, there would be an else-case for when the capacity has no generated implementation that prints an error saying that we are missing the wallace tree implementation. Push both the python script and the wallace tree module for your chosen capacities to github and then put clear instructions at the top of the python script for how to run it so that we can generate a new wallace tree implementation that includes a new capacity should we need it!

Xusine1131:
@stdavids Thanks! I'l try it out!

dpetrisko:
I wonder if there’s a step in FuseSoc that could be used to generate the correctly parameterized Verilog file before compilation...

proftaylor:
We have a slightly alternative take on machine-generated code to what we
would usually do in software, which is that we view the checked in
generator output as the “true” (read-only) copy and the python generator as
the “reconstruction / provenance / update code.” This way the user is never
hunting around trying to figure out what exactly what code was generated.
This is the classic problem with generate statements— you can’t easily see
what was generated, and with makefile generated files — Make is always
deleting them and sometimes you want to look at them.

The output files are big since they contain all parameterizations we think
anybody would use. So far we have not seen any tool scalability issues. But
if we did have some, then what you propose makes a lot of sense!

M

Xusine1131:
@tanglingshu Sorry to bother you. May I have your email or Wechat? I think I need so many synthesis results that it's not very convenient for you. So I want to learn the process and I hope I can do it on my own! Thank you!

tanglingshu:
@Xusine1131 No worries at all. You can add me on Wechat. my WC ID is ta1992118. My email is lingshu@uw.edu

Xusine1131:
@taylor-bsg Hi professor Taylor, Here is the PPA regression result of ALL integer multipliers currently in Basejump STL. The awesome data is from @tanglingshu and I have a great gratitude towards him! 

https://docs.google.com/spreadsheets/d/153VAgxY73i40jKyiOYy6s0p_SKkJzZHyb9c1GWvO83A/edit

Optimal configurations are highlighted in the sheet, for both pipelined and unpipelined version. In conclusion, my design is the best, with great advantages on timing compared with SystemVerilog * operator and bsg_mul_pipelined.

proftaylor:
Thanks Xusine!

For the CPA are you building your own or letting Synopsys synthesize?
Synopsys has really good adders.

Note: the occupancy of a pipelined multiplier is 1. Also, maybe you need
another column, which is latency (cycles)? Can you update the table?

I think we should discard any multipliers that run slower than say 275 MHz
since they are cycle time inhibiting. Then discard any adder that has the
same latency and occupancy of some other adder but worse area and frequency?

So which ones in your mind are the good configs?

Xusine1131:
@proftaylor Hi professor Taylor. I've read the resource reports generated by design compiler and I'm sure the CPA is synthesized into Synopsys adder.

Occupancy in the table of pipelined version might be a typo I forgot to fix when I copied the table from the unpipelined version :) Now I've renamed the column to Latency(cycles), which is the original meaning I wish to express.

Meanwhile I've updated the table according to your proposal. And the configs I think is suitable have been marked in purple such that you can easily find them. 

https://docs.google.com/spreadsheets/d/153VAgxY73i40jKyiOYy6s0p_SKkJzZHyb9c1GWvO83A/edit

Xusine1131:
@taylor-bsg I added a new column in the spreadsheet where I put my suggestion for multiplier selection. Hope this information is useful for you. If you're satisfied with the result, can we merge this PR?

PR #30, opened 2019-05-08
Add hardened 1rw memory with bit write mask for Xilinx ultrascale arc…

leonardxiang:
This hardened version solves the tag memory synthesis bug on HammerBlade Manycore project.


stdavids:
The general format for the hard directory is hard/<process>/<mirrored directories and files>, so this file should be moved to hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit.v

The hard/bsg_mem directory is tsmc180/tsmc250 before we started support multiple process nodes. Those should be moved.

leonardxiang:
Sure, thanks Scott~

drichmond:
Can we merge?

PR #37, opened 2019-05-16
Adding explicit widths to NOC direction for Verilator compatibility

dpetrisko:
No functional change, but if you leave enum values unsized, it causes issues in concatenations involving them


proftaylor:
Add a comment to source above modified line as to why widths are explicitly
given including Verilator version number.

Issue #38, opened 2019-05-16
Dlatch fixes


taylor-bsg:
https://github.com/bespoke-silicon-group/basejump_stl/blob/0a9b3a9ee26c233e126da93d3235db7ce8d598c9/bsg_misc/bsg_dlatch.v#L2

- Have input be clk_i and not en_i; we want these used in very specific cases (usually a negative clock dlatch, will grab and pass through the new value at the second half of the cycle).
- Use always_latch and not always_ff
- Add %m warning if it is instantiated; and i_know_this_is_a_bad_idea_p which disables the warning.

stdavids:
Attempted fix in PR #39

stdavids:
PR #39 accepted and merged.

Issue #21, opened 2019-04-15
bsg_dlatch should use always_latch


dpetrisko:
...to match design intent

https://github.com/bespoke-silicon-group/basejump_stl/blob/3585a9e120a163986d21f2dc5a67a82e7ace253a/bsg_misc/bsg_dlatch.v#L7

black-parrot:
And Scott promised to have this signal an error unless somebody uses the parameter i_know_this_is_usually_a_bad_idea_p.

M

stdavids:
Attempted fix in PR #39

stdavids:
PR #39 accepted and merged.

PR #39, opened 2019-05-16
Update to bsg_dlatch

stdavids:
Addressing issues #38 and #21.

Adding the i_know_this_is_a_bad_idea_p parameter and warning. Changed en_i port to clk_i. Refactored to use always_latch rather than always_ff.

### DIFF
--- a/bsg_misc/bsg_dlatch.v
+++ b/bsg_misc/bsg_dlatch.v
@@ -6,12 +6,10 @@ module bsg_dlatch #(parameter width_p="inv"
   ,output logic [width_p-1:0] data_o
   );
 
-  // Do not surround with translate off tags so that a warning is issued in DC
-  // if the user has not admitted that this is a bad idea.
-  if (i_know_this_is_a_bad_idea_p == 0)
-    begin
-      initial $warning("# WARNING (%m) using bsg_dlatch without admitting that this is a bad idea!");
-    end
+  // synopsys translate_off
+  initial assert (i_know_this_is_a_bad_idea_p == 1)
+    else $error("# Error (%m) using bsg_dlatch without admitting that this is a bad idea!");
+  // synopsys translate_on
 
   always_latch
     begin
--- a/bsg_misc/bsg_dlatch.v
+++ b/bsg_misc/bsg_dlatch.v
@@ -6,10 +6,8 @@ module bsg_dlatch #(parameter width_p="inv"
   ,output logic [width_p-1:0] data_o
   );
 
-  // synopsys translate_off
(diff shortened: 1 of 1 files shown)


stdavids:
Not sure I follow, there is just a single initial statement for the warning...

taylor-bsg:
Ah got it .. it was how the app was showing it. Probably better to use
assert instead of if statement?

stdavids:
Done, moved to using assert. I also surrounded it in translate_off / translate_on pragmas because with asserts there are no parameterizations that would remove the warning of initial statements during synthesis analysis.

taylor-bsg:
Sorry I did not understand this last part?

dpetrisko:
Oh interesting. Are you saying you can't conditionally generate asserts?

stdavids:
Initial statements and assert statements throw warnings in DC. Before, I used a generate if statement surrounding an initial statement so if that if statement resolves to false the initial statement will not be instantiated and DC is happy thus I didn't use translate pragmas because if you are not consenting that this is a bad idea, DC will throw a warning which is nice.

Moving to assert, if the statement resolves to true or false, DC still throws a warning. There is no parameterization that stops DC throwing a warning therefore I surrounded it with translate pragmas.

dpetrisko:
Came across this about "elaboration system tasks": 

image

https://www.veripool.org/issues/1429-Verilator-Feature-request-elaboration-tasks#change-4871

stdavids:
Hey Dan, I'm assuming that your suggestion is to move error to fatal? I like that idea, that will kill vcs if you do not acknowledge that the latch is a bad idea.

However this will have no effect on DC's elaboration and has no affect the aforementioned translate off pragmas, you'll simply get a VER-281 saying that the initial block is ignored.

dpetrisko:
Admittedly I haven’t tested it out, but according to this it sounds like if you put the code in the generate if _instead_ of an initial block, it flags it at elaboration instead of simulation. So DC might do the right thing / not complain since there’s no initial block

stdavids:
Ah I see, very interesting. I just tested in DC and VCS and this works well however we cannot use the %m in the message, presumably because it doesn't know the full hierarchical path yet. With that said, VCS does actually print the hier path just below the error and you can figure out the path via the DC log albeit not as clear as VCS. I'm in favor of this over the previous two implementations!

stdavids:
Oh, I should also note that to get this to work in VCS you need to add the `-assert svaext` flag. If you don't have that flag, it errors as a syntax error regardless of parameterization.

dpetrisko:
Nice! I just checked and Verilator seems fine with the syntax with no additional flags

taylor-bsg:
Going to merge...

PR #48, opened 2019-05-20
Cache lock2

tommydcjung:


bsg_cache/bsg_cache.v
```diff
@@ -342,9 +345,15 @@ module bsg_cache
...
+
+  // this is the signal that activates bsg_cache_miss. the conditions are:
+  // 1) it's load or store, and there is no tag hit
+  // 2) it's tagfl, and the set is valid and not locked.
+  // 3) it's afl, aflinv, or ainv, and the there is tag hit with set that is
+  // not locked.
   assign miss_v = v_v_r & (((ld_op_v_r | st_op_v_r) & ~(tag_hit_v[1] | tag_hit_v[0]))
-    | (tagfl_op_v_r & valid_v_r[addr_set_v])
-    | ((afl_op_v_r | aflinv_op_v_r | ainv_op_v_r) & (tag_hit_v[1] | tag_hit_v[0])));
+    | (tagfl_op_v_r & valid_v_r[addr_set_v] & ~lock_v_r[addr_set_v])
```
taylor-bsg: Tagfl copies the contents of the cache back out to DRAM, without removing it from the cache. This is still a reasonable thing to do even if the line is locked, so I think this needs to be fixed.

bsg_cache/bsg_cache.v
```diff
@@ -342,9 +345,15 @@ module bsg_cache
 
   assign tag_hit_v[1] = (addr_tag_v == tag_v_r[1]) & valid_v_r[1];
   assign tag_hit_v[0] = (addr_tag_v == tag_v_r[0]) & valid_v_r[0];
+
+  // this is the signal that activates bsg_cache_miss. the conditions are:
+  // 1) it's load or store, and there is no tag hit
+  // 2) it's tagfl, and the set is valid and not locked.
+  // 3) it's afl, aflinv, or ainv, and the there is tag hit with set that is
```
taylor-bsg: afl also should still write things back if it is locked. once again, it will not be invalidated.

what about aflinv, and ainv? I think they should probably ignore if the line is locked and do the action anyways. Otherwise, there will be unexpected behavior when the user is trying to do software coherence.  So a*inv would also unlock the line.

I think we should probably add an ALOCK and AUNLOCK instruction. The first will bring the line into the cache and lock it. The second will unlock the line but not evict it.

bsg_cache/bsg_cache_dma.v
```diff
@@ -212,31 +201,16 @@ module bsg_cache_dma
...
-        data_mem_v_n = data_mem_v_o;
-
-        data_flopped_n = data_mem_v_r
-          ? ~out_fifo_ready_lo
-          : data_flopped_r;
-
-        data_buf_n = (data_mem_v_r & ~out_fifo_ready_lo)
-          ? data_way_selected
-          : data_buf_r;
+        dma_state_n = (counter_r == block_size_in_words_p) & out_fifo_ready_lo
```
taylor-bsg: what happened here again?

bsg_cache/bsg_cache_miss.v
```diff
@@ -143,9 +146,19 @@ module bsg_cache_miss
...
+        // lock[0]  lock[1] valid[1]  valid[0]  mru_n  |  chosen_set_n
+        // -----------------------------------------------------------
+        // 1        X       X         X         X      |  1
+        // 0        1       X         X         X      |  0
+        // 0        0       0         X         X      |  0
+        // 0        0       1         0         X      |  1
+        // 0        0       1         1         0      |  1
+        // 0        0       1         1         1      |  0
+        // -----------------------------------------------------------
+        chosen_set_n = lock_v_i[0]
```
taylor-bsg: you can also put in the comments =)

bsg_cache/bsg_cache_dma.v
tommydcjung: data_mem output had to be buffered before latch_last_read_p changes, but now we don't need this extra buffer anymore. out_fifo is somewhat orthogonal change, I thought we should have this bsg_two_fifo, since the distance between bsg_cache_dma and bsg_cache_to_dram_ctrl can be long.

bsg_cache/bsg_cache.v
tommydcjung: https://github.com/bespoke-silicon-group/basejump_stl/issues/51

bsg_cache/bsg_cache_dma.v
```diff
@@ -206,37 +195,22 @@ module bsg_cache_dma
...
-          : data_way_selected;
- 
-        counter_n = out_fifo_ready_lo
-          ? counter_r + 1
-          : counter_r;
         dma_state_n = (counter_r == block_size_in_words_p) & out_fifo_ready_lo
           ? IDLE
           : SEND_EVICT_DATA;
+
+        counter_n = out_fifo_ready_lo
```
taylor-bsg: Seems like there is a BaseJump STL counter we should be able to use?

taylor-bsg:
This one is basically ready to go modulo my suggestion about using a BaseJump STL counter (and the future addition of a ALOCK and AUNLOCK instruction?

tommydcjung:
it's ready to go, except that we need to have that latch_last_read_p changes

tommydcjung:
counter added

bsg_cache/bsg_cache_miss.v
```diff
@@ -161,11 +176,9 @@ module bsg_cache_miss
         tag_mem_v_o = dma_done_i;
         tag_mem_w_o = dma_done_i;
         tag_mem_addr_o = addr_index_v;
-        tag_mem_data_o = {2{1'b1, addr_tag_v}};
-        tag_mem_w_mask_o = {
-          {(1+tag_width_lp){chosen_set_n}},
-          {(1+tag_width_lp){~chosen_set_n}}
-        };
+        tag_mem_data_o = {2{1'b1, 1'b0, addr_tag_v}};
```
taylor-bsg: Strange behavior? You are bring in a new line, overwriting the tags, but leaving the locked bit untouched, by masking it? Seems like clearing the lock makes more sense here?

bsg_cache/bsg_cache_miss.v
```diff
@@ -189,11 +202,9 @@ module bsg_cache_miss
         tag_mem_v_o = 1'b1;
         tag_mem_w_o = 1'b1;
         tag_mem_addr_o = addr_index_v;
-        tag_mem_data_o = {2{~(ainv_op_v_i | aflinv_op_v_i), {tag_width_lp{1'b0}}}};
-        tag_mem_w_mask_o = {
-          chosen_set_n, {tag_width_lp{1'b0}},
-          ~chosen_set_n, {tag_width_lp{1'b0}}
-        };
+        tag_mem_data_o = {2{1'b0, 1'b0, {tag_width_lp{1'b0}}}};
```
taylor-bsg: Maybe a comment here? Generally you would prefer to put the complexity in the data rather than in the mask, since sometimes you cannot mask at a bit granularity and have to waste ram space.  Is the issue that you do not know what the lock bits are, and just want to keep them the same?

bsg_cache/bsg_cache_dma.v
```diff
@@ -205,38 +213,26 @@ module bsg_cache_dma
         data_mem_v_o = in_fifo_v_lo;
         data_mem_w_o = in_fifo_v_lo;
         in_fifo_yumi_li = in_fifo_v_lo;
-        counter_n = in_fifo_v_lo
-          ? counter_r + 1
-          : counter_r;
+
+        counter_en = in_fifo_v_lo & (counter_r != (block_size_in_words_p-1));
```
taylor-bsg: factor these comparison signals into two reused variables: (counter_r == block_size_in_words_p-1) and  (counter_r == block_size_in_words_p)

bsg_cache/bsg_cache_miss.v
taylor-bsg: What kind of tests have you written for the lock bits?

tommydcjung: I thought load and store miss shouldn't touch the lock bit

tommydcjung: https://github.com/bespoke-silicon-group/basejump_stl/blob/cache_lock2/testing/bsg_cache/regression/trace_master_0.tr#L859

basically lock mechanism for every possible scenarios

bsg_cache/bsg_cache_miss.v
tommydcjung: this actually fixes the bug. previously, tagfl would have set the valid bit to 1'b1.

tommydcjung: actually, that would never happen, because if the set was invalid, tagfl would not have triggered the miss handler

bsg_cache/bsg_cache_miss.v
tommydcjung: also randomized test with some lines locked.

https://github.com/bespoke-silicon-group/basejump_stl/blob/cache_lock2/testing/bsg_cache/dmc/bsg_trace_rom.py#L82

bsg_cache/bsg_cache.v
```diff
@@ -11,22 +11,21 @@
...
     ,parameter sets_p="inv"
 
-    ,localparam lg_sets_lp=`BSG_SAFE_CLOG2(sets_p)
-    ,localparam data_mask_width_lp=(data_width_p>>3)
-    ,localparam lg_data_mask_width_lp=`BSG_SAFE_CLOG2(data_mask_width_lp)
-    ,localparam lg_block_size_in_words_lp=`BSG_SAFE_CLOG2(block_size_in_words_p)
-    ,localparam tag_width_lp=(addr_width_p-lg_data_mask_width_lp-lg_sets_lp-lg_block_size_in_words_lp)
-    ,localparam bsg_cache_pkt_width_lp=`bsg_cache_pkt_width(addr_width_p,data_width_p)
-    ,localparam bsg_cache_dma_pkt_width_lp=`bsg_cache_dma_pkt_width(addr_width_p)
+    ,parameter stat_mem_width_p=4 // sometimes we cannot generate SRAM macro with 3-bit width, so we set it to some minimal value.
```
taylor-bsg: We should remove this parameter and just have the input parameter to the size of the stat mem be the actual number of bits that is required. It is up to the hardened SRAM implementations to round up to whatever is most efficient at that level. RAM implementation is handled at the base_jump_stl/hard level.

bsg_cache/bsg_cache.v
```diff
@@ -393,8 +406,8 @@ module bsg_cache
   logic miss_tag_mem_v_lo;
   logic miss_tag_mem_w_lo;
   logic [lg_sets_lp-1:0] miss_tag_mem_addr_lo;
-  logic [2*(tag_width_lp+1)-1:0] miss_tag_mem_data_lo;
-  logic [2*(tag_width_lp+1)-1:0] miss_tag_mem_w_mask_lo;
+  logic [2*(tag_width_lp+2)-1:0] miss_tag_mem_data_lo;
```
taylor-bsg: add:

// TODO STYLE: use a struct

bsg_cache/bsg_cache_miss.v
taylor-bsg: Just thinking this through -- the only time you will miss on a locked line is if both lines are locked. But the effect will be that a new line is brought in, overwriting the locked line, but still leaving the locked line locked. Which seems a little strange?

In terms of testing, how are you verifying that the cache is correctly respecting the locking bits? Would it make sense to add some counters and have a command to read them? for example, # loads, # stores, # cache misses, # of evictions of dirty data? Then you could check them in your traces.

bsg_cache/bsg_cache_miss.v
```diff
@@ -143,9 +146,21 @@ module bsg_cache_miss
           ? dirty_r
           : dirty_i;
         dma_send_fill_addr_o = 1'b1;
-        chosen_set_n = valid_v_i[0]
-          ? (valid_v_i[1] ? ~mru_n : 1'b1)
-          : 1'b0;
+
```
taylor-bsg: add negedge assertion that prints a warning if both lines are locked during a cache miss?

bsg_cache/bsg_cache_miss.v
taylor-bsg: Maybe a more detailed comment then?

bsg_cache/bsg_cache.v
tommydcjung: https://github.com/bespoke-silicon-group/basejump_stl/issues/74

bsg_cache/bsg_cache_miss.v
tommydcjung: if the way is locked, it will never be chosen as a way to replace.
if neither way is locked, then it prioritizes the invalid way, then lru.

taylor-bsg:
@tommydcjung is working on merging this in

tommydcjung:
closing this for the other feature branch

PR #52, opened 2019-05-21
Changing wormhole router to use bsg_noc_links instead of ready-valid

dpetrisko:
This PR changes the wormhole_router and _adapters to use bsg_noc links on the network side. The testbenches have been updated to work with the new interface and I've tested the whole system with new links on BlackParrot in a separate branch there.

### DIFF
--- a/bsg_noc/bsg_wormhole_router_adapter_in.v
+++ b/bsg_noc/bsg_wormhole_router_adapter_in.v
@@ -4,6 +4,8 @@
  *  packet = {payload, length, y_cord, x_cord}
  */
 
+`include "bsg_noc_links.vh"
+
 module bsg_wormhole_router_adapter_in
   #(parameter max_num_flit_p="inv"
     , parameter max_payload_width_p="inv"
@@ -20,6 +22,8 @@ module bsg_wormhole_router_adapter_in
     , localparam padded_packet_width_lp=
       max_packet_width_lp+padding_width_lp
     , localparam len_offset_lp=(x_cord_width_p+y_cord_width_p)
+
+    , localparam bsg_ready_and_link_sif_width_lp=`bsg_ready_and_link_sif_width(width_lp)
   )
   (
     input clk_i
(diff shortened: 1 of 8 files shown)


dpetrisko:
Good idea, fixed!

tommydcjung:
looks fine. if you plan to buffer adapter in with bsg_two_fifo, it should probably go between adapter and router, not between LCE and adapter

dpetrisko:
Good call, I’ll move it there for consistency

dpetrisko:
Can we merge this in?

bsg_noc/bsg_wormhole_router_adapter_in.v
```diff
@@ -55,14 +75,14 @@ module bsg_wormhole_router_adapter_in
...
 
   always_comb begin
     state_n = state_r;
     count_n = count_r;
     ready_o = 1'b0;
-    v_o = 1'b0;
+    v_lo = 1'b0;
 
     case (state_r) 
       WAIT: begin
```
farzamgl: I might be mistaken, but shouldn't `ready_o` be 1 in `WAIT`?

testing/bsg_noc/bsg_wormhole_router_adapter_in/Makefile
```diff
@@ -1,4 +1,9 @@
+export BSG_IP_CORES_DIR=/home/petrisko/bitbucket/pre-alpha-release/basejump_stl
```
farzamgl: I don't think we want these 2 paths in basejump.
Also in the adapter_out testbench

dpetrisko: Good catch :)

bsg_noc/bsg_wormhole_router_adapter_in.v
dpetrisko: No, it's actually a late yumi. Essentially what happens is the adapter reads the input piecemeal and then asserts ready when it's done to dequeue the input

farzamgl: In that case, the input interface is not r&v anymore right?

dpetrisko: It’s debatably yumi. But in that last cycle, the ready does not depend on valid, so it’s technically still a helpful consumer

farzamgl: But the provider should keep track of when it can put the data on the bus externally. Possibly by registering the ready_o signal and clearing it on sent data.

dpetrisko: No, because the producer is producing regardless of whether the router is ready. All the producer needs to know is the moment when the data is consumed

farzamgl: Got it

PR #55, opened 2019-05-24
set default max_step_p of bsg_counter_up_down in bsg_flow_counter to 1

muwyse:
This change sets the max_step_p parameter to have a value of 1 for bsg_counter_up_down inside bsg_flow_counter. This eliminates a width mismatch warning from VCS and DC.

The bsg_flow_counter only ever increments or decrements by 1 since it is used to count tokens or credits of a single module between a r&v and valid-yumi input/output pair.


taylor-bsg:
bsg_flow_counter depends on bsg_counter_up_down, and the interface to bsg_counter_up_down changed slightly when bsg_counter_up_down_variable was merged in.

PR #58, opened 2019-05-29
Radix-4 FPU Divider

Xusine1131:


Xusine1131:
Exhaustive test based of float16 based on my own version is completed. And Berkeley FP Test is running now. I push the testing file as long as it's done.
I wonder if VCS supports multi-threading? Performing a exhaustive test really costs.

taylor-bsg:
The float16 exhaustive testing is for fp divider? Very cool! I think rather than implement more code, the next step is to focus on PPA comparison to our other modules, to characterize exactly where you are winning and how much.

Xusine1131:
@taylor-bsg Thanks for your suggestion, I'll organize a series of reports for all the modules I've designed regarding the PPA regression and how many speed-up after the Berkeley FP Test is finished!

PR #60, opened 2019-05-29
Add wormhole network building blocks

gaozihou:
Modifying 3 current files:
**bsg_parallel_in_serial_out.v** and **bsg_serial_in_parallel_out_full.v**: Add parameter "msb_then_lsb_p" to enable reversed serialization and de-serialization. The default value is 0 so all other modules are not affected.
**bsg_noc_links.vh**: Add wormhole packet data structure.

Adding 9 files:
**bsg_serial_in_parallel_out_full_buffered.v**: This de-serializer has 0 bubble in transition. It utilizes bsg_two_fifos inside as memory.
**bsg_channel_tunnel_wormhole.v**: A channel tunnel that accepts wormhole packets.
**bsg_link_source_sync_downstream.v** and **bsg_link_source_sync_upstream.v**: The new "common-link" modules, utilizing async-reset on token.
**bsg_link_iddr_phy.v** and **bsg_link_oddr_phy.v**: The DDR PHY for ASIC.
**bsg_link_ddr_downstream.v** and **bsg_link_ddr_upstream.v**: The integration of serializer, source_sync transceiver and DDR PHY.
**bsg_link_ddr.v**: Wrapper of link upstream and downstream.

The corresponding testbench is on bsg_manycore repo: bsg_manycore/testbenches/manycore_loopback_ddr_link


bsg_link/bsg_link_iddr_phy.v
```diff
@@ -0,0 +1,21 @@
...
+  
+  logic [2*width_p-1:0] data_r;
+  logic [width_p-1:0] data_p;
+  assign data_o = data_r;
+
+  always @(posedge clk_i)
+    data_p <= data_i;
+  
+  always @(negedge clk_i)
+    data_r <= {data_i, data_p};
```
stdavids: No need for data_p: `data_r = {data_i, data_r[2*width_p-1:width_p]};`

bsg_link/bsg_link_oddr_phy.v
```diff
@@ -0,0 +1,103 @@
+
+
```
stdavids: Description, and what are phy_180 and phy_90?

bsg_link/bsg_link_oddr_phy.v
```diff
@@ -0,0 +1,103 @@
...
+  (input reset_i
+  ,input clk_2x_i
+  ,input [2*width_p-1:0] data_i
+  ,output logic [width_p-1:0] data_r_o
+  ,output logic clk_r_o);
+  
+  logic odd, clk;
+  
+  always @(negedge clk_2x_i) begin
```
stdavids: always_ff everywhere

bsg_link/bsg_link_oddr_phy.v
```diff
@@ -0,0 +1,103 @@
...
+
+  always @(posedge clk_2x_i)
+    if(odd) 
+      data_r_o <= data_180[0+:width_p];
+    else
+      data_r_o <= data_180[width_p+:width_p];
+
+endmodule
+
+
```
stdavids: remove empty lines

bsg_link/bsg_link_oddr_phy.v
```diff
@@ -0,0 +1,103 @@
...
+  (input reset_i
+  ,input clk_2x_i
+  ,input [2*width_p-1:0] data_i
+  ,output logic [width_p-1:0] data_r_o
+  ,output logic clk_r_o);
+  
+  logic odd, clk;
+  
+  always @(negedge clk_2x_i) begin
+    clk <= ~reset_i & ~clk;
```
stdavids: Use if statement for reset idiom everywhere.

bsg_link/bsg_link_source_sync_downstream.v
```diff
@@ -0,0 +1,266 @@
...
+    ,.w_ptr_binary_r_o() // not needed
+    ,.w_ptr_gray_r_o()   // not needed
+    ,.w_ptr_gray_r_rsync_o(core_credits_gray_r_iosync)
+    );
+
+   // this logic allows us to return two credits at a time
+   // note: generally relies on power-of-twoness of io_credits_sent_r
+   // to do correct wrap around.
+
+   always_comb io_credits_sent_r_p1 = io_credits_sent_r+1;
```
stdavids: assign for one liners

bsg_dataflow/bsg_parallel_in_serial_out.v
```diff
@@ -27,6 +27,7 @@
 
 module bsg_parallel_in_serial_out #( parameter width_p = -1
                                    , parameter els_p   = -1
+                                   , parameter msb_then_lsb_p = 0
```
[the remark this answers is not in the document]
gaozihou: Done!

bsg_dataflow/bsg_serial_in_parallel_out_full.v
```diff
@@ -10,8 +10,11 @@
 module bsg_serial_in_parallel_out_full
   #(parameter width_p="inv"
     , parameter els_p="inv"
+    , parameter msb_then_lsb_p = 0
```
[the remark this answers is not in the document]
gaozihou: Done!

bsg_dataflow/bsg_parallel_in_serial_out.v
```diff
@@ -41,6 +42,10 @@ module bsg_parallel_in_serial_out #( parameter width_p = -1
     , output [width_p-1:0] data_o
     , input                yumi_i
     );
+    
+    
+    localparam init_cnt_lp = (msb_then_lsb_p == 0)? 0 : els_p-1;
+    localparam end_cnt_lp = (msb_then_lsb_p == 0)? els_p-1 : 0;
```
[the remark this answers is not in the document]
gaozihou: Done!

bsg_dataflow/bsg_serial_in_parallel_out_full.v
```diff
@@ -10,8 +10,11 @@
...
     , localparam counter_width_lp=`BSG_SAFE_CLOG2(els_p+1)
     , localparam lg_els_lp=`BSG_SAFE_CLOG2(els_p)
+    , localparam terminate_cnt_lp = (msb_then_lsb_p == 0)? els_p : (1<<counter_width_lp)-1
```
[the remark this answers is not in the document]
gaozihou: Done

bsg_dataflow/bsg_parallel_in_serial_out.v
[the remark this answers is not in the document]
gaozihou: Done

bsg_dataflow/bsg_serial_in_parallel_out_full.v
```diff
@@ -10,8 +10,11 @@
...
+    , localparam init_cnt_lp = (msb_then_lsb_p == 0)? 0 : els_p-1
```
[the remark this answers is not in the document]
gaozihou: Done

bsg_link/bsg_link_oddr_phy.v
gaozihou: Those two are 90-degree phase shift and 180-degree phase shift version, now they are in different files.

bsg_link/bsg_link_ddr.v
```diff
@@ -0,0 +1,91 @@
...
+  (// chip logic clock
+   input clk_i
+   
+   // IO clocks
+  ,input clk_1x_i
+  ,input clk_2x_i
+  
+  // all control signals synchronous to clk_i
+  ,input reset_i
+  ,input chip_reset_i
```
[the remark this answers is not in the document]
gaozihou: Thanks Scott! I changed reset_i to link_reset_i to be more descriptive. Also comments are added to the module!

bsg_dataflow/bsg_serial_in_parallel_out_full_buffered.v
```diff
@@ -0,0 +1,76 @@
...
+    ,.data_o(data_o[i])
+    ,.yumi_i(fifo_yumi_i[i]));
+    
+  end
+  
+  always @(posedge clk_i) begin
+    counter_r <= (reset_i)? init_cnt_lp : counter_n;
+  end
+  
+  always_comb begin
```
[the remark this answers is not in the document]
gaozihou: Thanks! Moving to bsg_counter_up_down.

bsg_link/bsg_link_iddr_phy.v
gaozihou: Thanks for the kind suggestion!

Typically there are 4 working modes for an IDDR (like the IDDR on Xilinx FPGA):
1. Output data valid for half cycle, posedge aligned
2. Output data valid for half cycle, negedge aligned
3. Output data valid for one cycle, posedge aligned
4. Output data valid for one cycle, negedge aligned

Mode 1 does not fit our need, because the negedge data is newer than posedge data (we want to receive posedge data first)
Mode 2 and 4 will work, but have higher requirement on signal integrity
Mode 3 should be the best

The original one was mode 4 and you suggested mode 2. But I guess we may go for mode 3 just to be safe!

bsg_link/bsg_link_ddr.v
```diff
@@ -0,0 +1,91 @@
...
+  ,parameter channel_width_p = 8
+  ,parameter num_channel_p = 1
+  ,parameter lg_fifo_depth_p = 6
+  ,parameter lg_credit_to_token_decimation_p = 3)
+
+  (// chip logic clock
+   input clk_i
+   
+   // IO clocks
+  ,input clk_1x_i
```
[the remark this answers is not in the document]
gaozihou: This is a good question! Generating 1x clock inside bsg_link_ddr will have 2 impacts:

1. If we instantiate two bsg_link_ddr on single chip, then we will generate two 1x clocks (which is fine... but is this what we want?)

2. The create_generated_clock constraint will be problematic. The syntax of this constraint is like :
**create_generated_clock –divide_by 2 –source [get_ports {CLK}]  U1/reg1:Q**
In order to find the actual generated clock, we may need the exact name of instantiated chip_top, wormhole_network_wrapper, bsg_link_ddr and bsg_link_ddr_upstream...

So my opinion is to put it right next to the clk_gen, what do you think?

gaozihou:
Thanks a lot for the detailed suggestions @stdavids ! 

Everything is fixed, except for the generated 1x clock (see comment within the conversation above).

stdavids:
Hey Paul, I know we just talked about the 1x and 2x clock ports, I am putting our conversation here so @taylor-bsg and @tommydcjung can chime in if they want.

So the main downside we came up with for pushing the 1x clock inside the comm link is that it basically hides a timing constraint (need to set the output of the downsample as a generated clock of the original 2x clock). However, there are other modules where this is true, particularly with CDC. We have talked a bit about how we should potentially start including some timing constraint frameworks for certain blocks this might be a good candidate for such a thing if we do remove the 1x port? Also we would need to put a pretty big warning in the comments and the toplevel of bsg_link is quite basic (just two instances).

The main up-side is that it reflects the physical design better (we want to put the downsample quite close to the DDR so we don't have to worry about phases over long distances).

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,397 @@
...
+
+ #(// Wormhole packet configurations
+   parameter width_p = "inv"
+  ,parameter x_cord_width_p = "inv"
+  ,parameter y_cord_width_p = "inv"
+  ,parameter len_width_p = "inv"
+  ,parameter reserved_width_p = "inv"
+  
+  // Total number of channels
+  ,parameter num_in_p = "inv"
```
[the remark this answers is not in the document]
gaozihou: This is NOT the same parameter as the num_channel_p on bsg_link_ddr. The num_channel_p is the physical IO channels we have on IO padring. num_in_p represents how many stream of traffics we want to merge in channel tunnel.

"num_in_p" name should be consistent with the original channel tunnel, so maybe I need to change the comment here for channel tunnel wormhole!

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,397 @@
...
+  
+  // Channel Tunnel Data Output
+  logic [mux_num_in_lp-1:0] ofifo_valid_o, ofifo_yumi_i;
+  logic [mux_num_in_lp-1:0][width_p-1:0] ofifo_data_o;
+  
+  for (i = 0; i < num_in_p; i++) begin
+  
+    logic ofifo_data_ready_o, ofifo_header_ready_o;
+  
+    // Header to CT
```
taylor-bsg: comment that describes what this complex expression is accomplishing.

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,397 @@
...
+    logic [len_width_p-1:0] ocount_r, ocount_n;
+    assign ocount_n = (v_i[i] & ready_o[i])? 
+        (ocount_r==0)? data_i[i][len_offset_lp+:len_width_p] : ocount_r-1 : ocount_r;
+    
+    always_ff @(posedge clk_i) begin
+        if (reset_i)
+            ocount_r <= 0;
```
taylor-bsg: Use '0;  0 is actually 32'b0

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,397 @@
...
+    assign link_o_cast[i].ready_and_rev = ready_o[i];
+
+    assign link_o_cast[i].v = v_o[i];
+    assign link_o_cast[i].data = data_o[i];
+    assign yumi_i[i] = v_o[i] & link_i_cast[i].ready_and_rev;
+  
+  end
+  
+  
+  // Channel Tunnel Data Output
```
taylor-bsg: More detailed comment that says what this block of code is doing.

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,397 @@
...
+        if (reset_i)
+            ocount_r <= 0;
+        else
+            ocount_r <= ocount_n;
+    end
+    
+    // Data fifo
+    bsg_fifo_1r1w_small 
+   #(.width_p(width_p)
+    ,.els_p(4))
```
taylor-bsg: Comment justifying why this fifo needs to be four elements.

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,397 @@
...
+        (ocount_r==0)? data_i[i][len_offset_lp+:len_width_p] : ocount_r-1 : ocount_r;
+    
+    always_ff @(posedge clk_i) begin
+        if (reset_i)
+            ocount_r <= 0;
+        else
+            ocount_r <= ocount_n;
+    end
+    
+    // Data fifo
```
taylor-bsg: more detailed comment explaining exactly what this fifo is for

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,397 @@
...
+
+    ,.ready_o(ofifo_data_ready_o)
+    ,.data_i(data_i[i])
+    ,.v_i(~(ocount_r==0) & v_i[i])
+
+    ,.v_o(ofifo_valid_o[i])
+    ,.data_o(ofifo_data_o[i])
+    ,.yumi_i(ofifo_yumi_i[i]));
+    
+    // Header fifo
```
taylor-bsg: more detailed comment explaining purpose of this fifo

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,397 @@
...
+    ,.v_o(inside_valid_li[i])
+    ,.data_o(inside_data_li[i])
+    ,.yumi_i(inside_yumi_lo[i]));
+    
+    assign ready_o[i] = (ocount_r==0)? ofifo_header_ready_o : ofifo_data_ready_o;
+
+  end
+  
+  
+  // Header out of CT
```
taylor-bsg: more detailed comment explaining what this FIFO is for.

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,397 @@
...
+  ,.ready_o(headerout_ready_o)
+  ,.data_i(outside_data_lo)
+  ,.v_i(outside_valid_lo)
+
+  ,.v_o(ofifo_valid_o[num_in_p])
+  ,.data_o(ofifo_data_o[num_in_p])
+  ,.yumi_i(ofifo_yumi_i[num_in_p]));
+  
+  
+  // Channel Tunnel Output Select
```
taylor-bsg: more detailed comment explaining what heuristic or algorithm it is using to select, and also what channel tunnel output select means.

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,397 @@
...
+  logic [mux_num_in_lp-1:0] ififo_valid_i, ififo_ready_o;
+  
+  for (i = 0; i < num_in_p; i++) begin
+  
+    logic ififo_valid_o, ififo_yumi_i;
+    logic [width_p-1:0] ififo_data_o;
+    
+    bsg_fifo_1r1w_large 
+   #(.width_p(width_p)
+    ,.els_p(remote_credits_p*max_len_p))
```
taylor-bsg: Comment explaining why this is the correct expression for the size of this fifo

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,397 @@
...
+
+    ,.ready_o(ififo_ready_o[i])
+    ,.data_i(multi_data_i)
+    ,.v_i(ififo_valid_i[i])
+
+    ,.v_o(ififo_valid_o)
+    ,.data_o(ififo_data_o)
+    ,.yumi_i(ififo_yumi_i));
+    
+    // dummy data out of CT
```
taylor-bsg: more detailed comment

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,397 @@
...
+        (icount_r==0)? inside_data_lo[i][len_offset_lp+:len_width_p] : icount_r-1 : icount_r;
+    
+    always_ff @(posedge clk_i) begin
+        if (reset_i)
+            icount_r <= 0;
+        else
+            icount_r <= icount_n;
+    end
+    
+    assign v_o[i] = (icount_r==0)? inside_valid_lo[i] : ififo_valid_o;
```
taylor-bsg: align (icount_r==0) to enhance code symmetry and readability

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,397 @@
...
+// and need to merge them together to the IO channel (n >= 2)
+// All wormhole packet parameters should match corresponding wormhole network
+//
+// Refer to bsg_channel_tunnel for more information on flow control
+//
+//
+
+`include "bsg_noc_links.vh"
+
+module  bsg_channel_tunnel_wormhole
```
taylor-bsg: Please review use of 0 versus '0. In most cases you mean '0.

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,397 @@
...
+    ,.v_i(ififo_valid_i[i])
+
+    ,.v_o(ififo_valid_o)
+    ,.data_o(ififo_data_o)
+    ,.yumi_i(ififo_yumi_i));
+    
+    // dummy data out of CT
+    logic [len_width_p-1:0] icount_r, icount_n;
+    assign icount_n = (yumi_i[i])? 
+        (icount_r==0)? inside_data_lo[i][len_offset_lp+:len_width_p] : icount_r-1 : icount_r;
```
taylor-bsg: factor icount_r=='0 into icount_r_is_zero, since the expression is used a lot

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,397 @@
...
+  in_ready_mux
+  (.data_i(ififo_ready_o)
+  ,.sel_i(in_sel_r)
+  ,.data_o(multi_ready_o));
+  
+  for (i = 0; i < mux_num_in_lp; i++) begin
+    assign ififo_valid_i[i] = (i==in_sel_r)? multi_v_i : 0;
+  end
+  
+  always_comb begin
```
taylor-bsg: comment describing algorithm being used here

bsg_dataflow/bsg_serial_in_parallel_out.v
```diff
@@ -34,8 +34,8 @@ module bsg_serial_in_parallel_out #(parameter   width_p   = -1
```
taylor-bsg: unnecessary code churn -- revert.

bsg_dataflow/bsg_serial_in_parallel_out_full.v
```diff
@@ -10,8 +10,11 @@
 module bsg_serial_in_parallel_out_full
```
taylor-bsg: Why not merge with bsg_serial_in_parallel_out to preserve symmetry with bsg_parallel_in_serial_out?

bsg_dataflow/bsg_serial_in_parallel_out_full.v
```diff
@@ -36,9 +39,11 @@ module bsg_serial_in_parallel_out_full
   logic clear_li;
   logic up_li;
 
+  if (msb_first_p == 0) begin: lsb_first
```
taylor-bsg: begin should be on next line.

bsg_dataflow/bsg_serial_in_parallel_out_full.v
```diff
@@ -48,9 +53,18 @@ module bsg_serial_in_parallel_out_full
 
     ,.count_o(count_lo)
   ); 
+  
+  end else begin: msb_first
```
taylor-bsg: a. The semantics of this counter do not match the semantics of the previous counter.  Can clear and up be asserted on the same cycle? In this case, the up_li signal would be ignored.
b. does it make sense to generalize bsg_counter_clear_up to handle both cases?

bsg_dataflow/bsg_serial_in_parallel_out_full_buffered.v
```diff
@@ -0,0 +1,86 @@
...
+    count_down_lo = 0;
+    if (v_i & ready_o)
+        if (counter_r == end_count_lp)
+            end_count_lo = 1;
+        else
+            if (msb_first_p == 0) count_up_lo = 1;
+            else count_down_lo = 1;
+  end
+  
+  bsg_counter_up_down
```
taylor-bsg: Reset is generally reserved for resetting; we use clear if we want to zero things. Can up or down be asserted at the same time as clear and what is the expected outcome?

bsg_link/bsg_link_ddr.v
```diff
@@ -0,0 +1,106 @@
...
+// for more information on flow control
+//
+//
+
+module bsg_link_ddr
+
+ #(parameter width_p = "inv"
+  ,parameter channel_width_p = 8
+  ,parameter num_channel_p = 1
+  ,parameter lg_fifo_depth_p = 6
```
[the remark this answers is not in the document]
taylor-bsg: This is a bottom-up requirement, which comes from the need to gray code pointers to cross a clock domain. We could add an comment to this effect, but it's okay not to get into it, since the underlying module demands it, and it violates the principle of information hiding and abstraction.

taylor-bsg: comment saying on how the user would compute the correct number for a particular system. When I have done similar calculations, I have seen 7 as potentially being the more correct number for FPGA.

bsg_link/bsg_link_ddr.v
```diff
@@ -0,0 +1,106 @@
...
+  ,parameter lg_credit_to_token_decimation_p = 3)
+
+  (// chip logic clock
+   input clk_i
```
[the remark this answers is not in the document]
taylor-bsg: from bsg style guide:

https://docs.google.com/document/d/1xA5XUzBtz_D6aSyIBQUwFk_kSUdckrfxa2uzGjMgmCU/edit#

"For modules that involve the interaction of multiple clocks, then every wire should indicate in the beginning of the name which clock it is synchronous to. "

Thus, suggest that you name the clock core_clk_i, and then rename the signals in this module according to what clock relates to them. 

This helps us maintain correct clock discipline.

bsg_link/bsg_link_ddr.v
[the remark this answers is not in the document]
taylor-bsg: Let's come back to this one after the other items have been resolved.

bsg_link/bsg_link_ddr_downstream.v
```diff
@@ -0,0 +1,127 @@
...
+
+ #(parameter width_p = "inv"
+  ,parameter channel_width_p = 8
+  ,parameter num_channel_p = 1
+  ,parameter lg_fifo_depth_p = 6
+  ,parameter lg_credit_to_token_decimation_p = 3
+  ,localparam ddr_width_lp = channel_width_p*2
+  ,localparam piso_ratio_lp = width_p/(ddr_width_lp*num_channel_p))
+
+  (input clk_i
```
taylor-bsg: from bsg style guide:

https://docs.google.com/document/d/1xA5XUzBtz_D6aSyIBQUwFk_kSUdckrfxa2uzGjMgmCU/edit#

"For modules that involve the interaction of multiple clocks, then every wire should indicate in the beginning of the name which clock it is synchronous to. "

Thus, suggest that you name the clock core_clk_i, and then rename the signals in this module according to what clock relates to them.

This helps us maintain correct clock discipline.

bsg_link/bsg_link_ddr_upstream.v
```diff
@@ -0,0 +1,137 @@
...
+ #(parameter width_p = "inv"
+  ,parameter channel_width_p = 8
+  ,parameter num_channel_p = 1
+  ,parameter lg_fifo_depth_p = 6
+  ,parameter lg_credit_to_token_decimation_p = 3
+  ,localparam ddr_width_lp = channel_width_p*2
+  ,localparam posi_ratio_lp = width_p/(ddr_width_lp*num_channel_p))
+
+  (input clk_i
+  ,input clk_1x_i
```
taylor-bsg: stdavids 3 days ago  Member
Why do we expose the clk_1x_i as a port rather than down sampling the 2x clock inside this module?

  @gaozihou
gaozihou 3 days ago  Author Member
This is a good question! Generating 1x clock inside bsg_link_ddr will have 2 impacts:

If we instantiate two bsg_link_ddr on single chip, then we will generate two 1x clocks (which is fine... but is this what we want?)

The create_generated_clock constraint will be problematic. The syntax of this constraint is like :
create_generated_clock –divide_by 2 –source [get_ports {CLK}] U1/reg1:Q
In order to find the actual generated clock, we may need the exact name of instantiated chip_top, wormhole_network_wrapper, bsg_link_ddr and bsg_link_ddr_upstream...

So my opinion is to put it right next to the clk_gen, what do you think?

bsg_link/bsg_link_ddr.v
```diff
@@ -0,0 +1,106 @@
+
```
taylor-bsg: I suggest deleting this module. It just instantiates two non-interacting modules; making the code base larger.

taylor-bsg: See newly added "Justify your wrappers " in BSG SystemVerilog coding guidelines

bsg_link/bsg_link_ddr_upstream.v
stdavids: From a physical design perspective we want to generate the 1x clock close to the sinks to prevent the tools from having to create two large clock trees (original 2x and generated 1x) that are kept in phase over long distances... that is tough. Because the downsampler here is 2 gates I think it is fine that we have multiple generated clocks for all the bsg_links.

For the timing constraint, I don't think it would be too bad to push inside the bsg_link_ddr. We can use a bsg_dff module to give us a consistent port name for the new generated clock source. The constraint example you provided looks like the pin is post-mapped but we can do this pre-compile and use module ports and it will propagate to the mapped pin. So then the tcl proc to constrain this would take the master clock object (so we can query the source pin) and the bsg_link_ddr instance path (so we can build the path to the double sampler bsg_dff module). We could also use the bsg_counter_clock_downsample module but that is more for variable downsampling, I'd be curious what it looks like when the downsample value is static.

bsg_link/bsg_link_ddr.v
stdavids: Here would be the reasons to keep the wrapper:
1. If we don't always want a 1:1 ratio of upstream and downstream links (ie. would anyone ever want just upstream or just downstream? I honestly don't know)
2. Physical design -- so we can use placement bounds to place the streams near their respective IO drivers where the upstream and downstream drivers are not near each other.

taylor-bsg: Those are reasons not to keep the wrapper, right?  I think the main reason to have this module is just if we expect to write a lot of code that always instantiates both, and it saves LOC. But my guess is that we will have some other module that will instantiate these as a pair anyways.

bsg_link/bsg_link_ddr_upstream.v
```diff
@@ -0,0 +1,137 @@
...
+    ,.core_valid_i(out_ps_yumi_li)
+    ,.core_ready_o(out_ps_ready_li[i])
+
+    // source synchronous output channel; going to chip edge
+    ,.io_data_r_o(out_ddr_data_o)
+    ,.io_valid_r_o(out_ddr_valid_o)
+    ,.token_clk_i(io_token_i[i]));
+
+
+    bsg_link_oddr_phy
```
taylor-bsg: merging these two bsg_link_oddr_phy's will better convey (and potentially, achieve) the intent that the delays are matched; and will also shorten the code.

bsg_link/bsg_link_oddr_phy_180.v
```diff
@@ -0,0 +1,42 @@
+//
```
taylor-bsg: Is this module even used?

bsg_link/bsg_link_oddr_phy_90.v
```diff
@@ -0,0 +1,42 @@
+//
```
taylor-bsg: Is this module used?

bsg_link/bsg_link_source_sync_downstream.v
```diff
@@ -0,0 +1,266 @@
...
+    , output [channel_width_p-1:0] core_data_o
+    , output                       core_valid_o
+    , input                        core_yumi_i
+   );
+  
+  
+  // local reset
+  logic io_reset_lo;
+  
+  bsg_launch_sync_sync
```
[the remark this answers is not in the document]
taylor-bsg: bsg_launch_sync_sync is much safer. bsg_sync_sync is only used when there is no launch clock (i.e. it is a purely asynchronous signal -- usually a reset coming from off chip.)

bsg_link/bsg_link_source_sync_downstream.v
```diff
@@ -0,0 +1,266 @@
+// MBT 7/24/2014
```
taylor-bsg: what is the relationship of these modules to the original SSI modules? Isn't there a way into git to copy those files and then make the diffs so we know what the differences are?

taylor-bsg:
> Hey Paul, I know we just talked about the 1x and 2x clock ports, I am putting our conversation here so @taylor-bsg and @tommydcjung can chime in if they want.
> 
> So the main downside we came up with for pushing the 1x clock inside the comm link is that it basically hides a timing constraint (need to set the output of the downsample as a generated clock of the original 2x clock). However, there are other modules where this is true, particularly with CDC. We have talked a bit about how we should potentially start including some timing constraint frameworks for certain blocks this might be a good candidate for such a thing if we do remove the 1x port? Also we would need to put a pretty big warning in the comments and the toplevel of bsg_link is quite basic (just two instances).
> 
> The main up-side is that it reflects the physical design better (we want to put the downsample quite close to the DDR so we don't have to worry about phases over long distances).

Yeah we can discuss after Paul iterates on some of the existing suggestions

bsg_dataflow/bsg_parallel_in_serial_out.v
```diff
@@ -27,6 +27,9 @@
 
 module bsg_parallel_in_serial_out #( parameter width_p = -1
                                    , parameter els_p   = -1
+                                   , parameter msb_first_p = 0
+                                   , localparam init_count_lp = (msb_first_p == 0)? 0 : els_p-1
```
taylor-bsg: Going to push back on this change:
@stdavids 
1) Some synthesis tools don't like localparam in parameter list.
2) You should only put _lp's in the parameter list if they are needed for the module interface, which is not the case here; or, more rarely, it is something that shouldn't be overridden by a layer user, but would make compelling sense for an expert in the know to modify. You usually would only do this if you actually needed to modify the parameter in practice to test something.

bsg_link/bsg_link_ddr.v
stdavids: Yeah, sorry... I don't know why but for some reasons I was thinking that the suggestion was to flatten bsg_link_ddr. I agree, I think this modules is not needed.

bsg_dataflow/bsg_parallel_in_serial_out.v
dpetrisko: Eek, which synthesis tools have problems with this? We use them all over the place for our interface structs in BlackParrot

stdavids: I'm fine with the push back, I know that older versions of DC (like 5 years ago) don't like this but modern DC is totally fine with it. It was actually the heavy usage in BP that influenced my original suggestion.

taylor-bsg: Sorry, simulation tools; Cadence Incisive 15.20 (circa 2017):

irun: 15.20-s038: (c) Copyright 1995-2017 Cadence Design Systems, Inc.
module foo #(localparam bar=1);
                      |
ncvlog: *E,LOCALP (testbench.sv,3|22): illegal localparam in list of parameters [12.2(IEEE-2001)].

taylor-bsg: Don't worry about updating BlackParrot for now..

dpetrisko: It's a SV-2012 construct, does it still not work with irun/ncvlog -sv?

I just checked and icarus verilog doesn't support it either (even with 2012 flags), although Verilator is obviously fine with it.

Aside: Elaboration time seems to blow up using parameters vs localparams (for BP's excessive use case), so I have a strong preference for keeping them there.  Most likely we could just s/localparam/parameter, if we're dying to support something else. 

For BaseJump, it makes sense to support the LCD

taylor-bsg:
Interesting point! Presumably this is only for localparams that are placed
in the module signature, so moving them out if they are not needed also
helps with this.

bsg_dataflow/bsg_channel_tunnel_wormhole.v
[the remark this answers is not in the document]
gaozihou: Hi Scott, this parameter is not the same as num_channels_p on bsg_link_ddr. I've updated the comments to eliminate misunderstanding!

taylor-bsg:
Checkout edaplayground.com for some old tools, including commercial ones

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,397 @@
...
+  ,parameter reserved_width_p = "inv"
+  
+  // Total number of channels
+  ,parameter num_in_p = "inv"
+  
+  // Max number of wormhole packets buffer can store
+  ,parameter remote_credits_p = "inv"
+  
+  // Max possible number of wormhole packet flits
+  ,parameter max_len_p = "inv"
```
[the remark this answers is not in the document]
gaozihou: Hi Scott, this piece comment is wrong... It is not number of wormhole flits, but num_flits-1. I've updated the comments to match what this parameter is doing!

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,397 @@
...
+  
+  // How often (every ? wormhole packets) does channel tunnel return credits
+  ,parameter lg_credit_decimation_p = "inv"
```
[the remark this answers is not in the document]
gaozihou: Default value updated!

bsg_dataflow/bsg_channel_tunnel_wormhole.v
gaozihou: Thanks prof! This "0" is the initial value of wormhole flit counter, so it do not have to be 0. In the first version channel tunnel (two months ago) this value was "1"... Then Tommy suggested we may represent wormhole length by "num_flits-1", so the init value for counter was changed to "0".

I have push the modification, in which all "0"s are replaced by "counter_value_min_lp". In this way it looks much better than before.

bsg_dataflow/bsg_parallel_in_serial_out.v
stdavids: Oh that is an interesting data point, do you have numbers? (Time to elaborate pre and post convert and the number of localparams converted to parameters)?

I think the answer is not to convert all localparams to parameters though but to simply put the localparams in the module unless they are needed in the port sizes.

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,459 @@
...
+
+`include "bsg_noc_links.vh"
+
+module  bsg_channel_tunnel_wormhole
+
+ #(// Wormhole packet configurations
+   parameter width_p = "inv"
+  ,parameter x_cord_width_p = "inv"
+  ,parameter y_cord_width_p = "inv"
+  ,parameter len_width_p = "inv"
```
stdavids: Isn't this always the $clog2(max_len_p) ?

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,459 @@
...
+  ,parameter num_in_p = "inv"
+  
+  // Max number of wormhole packets buffer can store
+  // This is independent of how many flits each wormhole packet has
+  ,parameter remote_credits_p = "inv"
+  
+  // Max possible "wormhole packet length" setting
+  // Note that the length of wormhole packet is represented by (num_flits - 1)
+  // If max possible number of flits is n, then max_len_p should be set to (n-1)
+  ,parameter max_len_p = "inv"
```
stdavids: So if my max length is 0 then I can set at most 1 flit? If that is true, I think the naming of this parameter is confusing, I'm not sure how a length of 0 correlates to 1 flit. Perhaps we just specify the max number of flits per packet?

bsg_dataflow/bsg_parallel_in_serial_out.v
dpetrisko: Unfortunately, I don’t. I only noticed the difference after changing from passing all of the BP parameters down the chain to the config enum index into config array approach.

Definitely. But there’s not a one-liner to convert the port sizes to parameters. 

Actually, the sv2v conversion could take care of this

bsg_dataflow/bsg_channel_tunnel_wormhole.v
gaozihou: Hi Scott, len_width_p may not be $clog2(max_len_p). len_width_p is part of "wormhole packet configurations", which is the properties of wormhole network. max_len_p is property of channel_tunnel. We cannot assume that all kinds of packets in network will go through channel tunnel to IO interface.

Here is an example, in black parrot wormhole network the max possible len is 8 (cached load / store). But in the future we may push all those packets to on-ASIC DRAM controller, then the max_len_p on channel tunnel is 1, but len_width_p should still be 4 (same network, same set of configurations).

bsg_dataflow/bsg_channel_tunnel_wormhole.v
gaozihou: The "len" here represents "number of payload flits", which does not include header flit. If the len=0 in a wormhole packet, then it has only one header flit, zero payload flit.

This naming is consistent to "len" parameter in wormhole packet definition. We deliberately made it this way because if we have at most 8 flits in a packet, we only need 3 bits to represent it (max len is 7).

bsg_dataflow/bsg_channel_tunnel_wormhole.v
dpetrisko: Does it make sense to have something like wh_len and ch_len to differentiate?

bsg_dataflow/bsg_channel_tunnel_wormhole.v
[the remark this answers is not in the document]
taylor-bsg: Yep, also using num_in_p makes it symmetric with bsg_channel_tunnel.

bsg_dataflow/bsg_channel_tunnel_wormhole.v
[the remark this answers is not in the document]
taylor-bsg: I think you guys are just searching for the right term -- perhaps you should say "max_payload_flits_p"?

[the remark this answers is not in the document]
taylor-bsg: You can define payload flits are the number of flits in the wormhole packet excluding the header. I think this will make it less ambiguous and also allow you to remove some of the comments which are also confusion.

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -22,19 +22,30 @@ module  bsg_channel_tunnel_wormhole
...
+  // Note that the length of wormhole packet is represented by (num_flits - 1)
+  // If max possible number of flits is n, then max_len_p should be set to (n-1)
   ,parameter max_len_p = "inv"
   
-  // How often (every ? wormhole packets) does channel tunnel return credits
-  ,parameter lg_credit_decimation_p = "inv"
+  // How often does channel tunnel return credits to sender
+  // Channel tunnel return credits to sender every (2^lg_credit_decimation_p) packets
+  // This is independent of how many flits each wormhole packet has
+  ,parameter lg_credit_decimation_p = `BSG_MIN($clog2(remote_credits_p+1),4)
```
taylor-bsg: Comment: match default value in child module.

Aside:  is clog2(remote_credits_p+1) actually right?  I think it might not ever send over any credits in this case.  For example,  remote_credits_p=4 -> $clog2(4+1) = 3. It would only send a credit packet over when more than 2^3 credits had accumulated which would never happen? (But at least it will fail in simulation pretty fast.)

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -131,26 +142,44 @@ module  bsg_channel_tunnel_wormhole
...
+  // go into original bsg_channel_tunnel.
+  // Remaining data packet flits are buffered in ofifo.
+  // A counter is used to handle the wormhole packet flits
+  
   for (i = 0; i < num_in_p; i++) begin
   
     logic ofifo_data_ready_o, ofifo_header_ready_o;
   
     // Header to CT
     logic [len_width_p-1:0] ocount_r, ocount_n;
```
taylor-bsg: This is a counter and should be in basejump_stl/bsg_misc. I suggest creating bsg_counter_dynamic_start_en_down.v, based on bsg_counter_dynamic_limit_en.v; and have a parameter called last_val_p, which is the last value the counter will take.

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -293,25 +341,31 @@ module  bsg_channel_tunnel_wormhole
     
     // dummy data out of CT
     logic [len_width_p-1:0] icount_r, icount_n;
-    assign icount_n = (yumi_i[i])? 
-        (icount_r==0)? inside_data_lo[i][len_offset_lp+:len_width_p] : icount_r-1 : icount_r;
+    logic icount_r_is_min;
```
taylor-bsg: I believe this is the same counter and can be factored out per my comments above?

bsg_dataflow/bsg_serial_in_parallel_out_full.v
gaozihou: Hi prof,
a. It is true that the counter is not the same as bsg_counter_clear_up (It is actually bsg_counter_up_down). For this module, up_li and clear_li will never be 1 at the same time (see link below), so it is safe here. [https://github.com/bespoke-silicon-group/basejump_stl/blob/2540ceb0a90d001a3a97c387a776e74cf83ba78a/bsg_dataflow/bsg_serial_in_parallel_out_full.v#L66](url)

b. Yeah we can definitely add a "bsg_counter_clear_up_down". For this specific usage, bsg_counter_up_down should be good enough!

bsg_dataflow/bsg_channel_tunnel_wormhole.v
taylor-bsg: Thanks Paul -- see my comments about using "payload_num_flits" to make this all more clear. Regarding 0, my suggestion generally is just to use '0 instead of 0.

bsg_dataflow/bsg_serial_in_parallel_out_full_buffered.v
gaozihou: Yeah you are right, if reset and up/down are non-zero at the same time, the up/down value will be ignored. 
For this specific SIPO, the up/down and clear will never be asserted at the same time (if/else relationship, only one of them can be asserted), so using bsg_counter_up_down will make hardware a little bit simpler!

bsg_dataflow/bsg_parallel_in_serial_out.v
```diff
@@ -27,6 +27,9 @@
 
 module bsg_parallel_in_serial_out #( parameter width_p = -1
                                    , parameter els_p   = -1
+                                   , parameter msb_first_p = 0
```
taylor-bsg: MSB first creates a lot of complexity in the module.  Instead of implementing it the two ways; I suggest instead you instead create a module bsg_array_reverse in bsg_misc, which uses a generate for loop to reorder all of the elements in an array. Then for msb_first_p, you can just reverse the input of this module (or the output for serial to parallel) using bsg_array_reverse and only have the implementation for one direction.

taylor-bsg: I also suggest placing bsg_parallel_in_serial_out and bsg_parallel_out_serial_in in a separate pull request.

taylor-bsg:
See my high-level comments about how to implement this module -- using
bsg_array_reverse will significantly reduce the complexity and code size of
what you are doing.

M

bsg_misc/bsg_counter_dynamic_start_en_down.v
```diff
@@ -0,0 +1,34 @@
...
+  ,output                     is_last_val_o);
+
+  assign is_last_val_o = ( counter_o == last_val_p );
+  
+  always_ff @ (posedge clk_i)
+    if (reset_i)
+      counter_o <= last_val_p;
+    else if (en_i)
+      if (is_last_val_o) counter_o <= start_i;
+      else               counter_o <= counter_o - 1;
```
dpetrisko: This should be width_p'(1)

dpetrisko: Either 1'b1 or width_p'(1)

gaozihou: 32'b1 should be good in my opinion.

dpetrisko: It's not misunderstanding, it's to avoid width_mismatch warnings

gaozihou: In VCS, the only mismatching warning happens on ports.

taylor-bsg:
Still, why is 32’b1 the right answer here?

M

bsg_link/bsg_link_iddr_phy.v
```diff
@@ -0,0 +1,35 @@
...
+  ,output [2*width_p-1:0] data_o);
+  
+  logic [2*width_p-1:0] data_r;
+  logic [width_p-1:0] data_n_r, data_p_r;
+  
+  assign data_o = data_r;
+
+  always_ff @(posedge clk_i) begin
+    data_p_r <= data_i;
+    data_r <= {data_n_r, data_p_r};
```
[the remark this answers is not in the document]
gaozihou: Thanks prof, I have updated the comments here. data_p_r is actually earlier in time than data_n_r.

bsg_link/bsg_link_oddr_phy.v
```diff
@@ -0,0 +1,42 @@
...
+  ,input [2*width_p-1:0] data_i
+  ,output logic [width_p-1:0] data_r_o
+  ,output logic clk_r_o);
+  
+  logic odd, clk;
+  
+  always_ff @(negedge clk_2x_i) begin
+    if (reset_i) clk <= 0;
+    else clk <= ~clk;
+    clk_r_o <= clk;
```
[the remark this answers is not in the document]
gaozihou: We need this for register packing in FPGA. Also, it is good for delay matching on ASIC (keeping output flop 0-extra-fanout).

bsg_link/bsg_link_oddr_phy.v
```diff
@@ -0,0 +1,42 @@
...
+  end
+    
+  always_ff @(posedge clk_2x_i)
```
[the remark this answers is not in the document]
gaozihou: Thanks prof, clk_r is negedge-aligned while odd_r is posedge-aligned, so they are different.

bsg_link/bsg_link_ddr.v
gaozihou: Thanks prof! I have updated the comment for this parameter. 6 is the experimental number on Lattice FPGA (minimum number to keep sending without stalling) @400MHz. I think we might need 7 if it runs faster than 400M?

bsg_misc/bsg_counter_dynamic_start_en_down.v
dpetrisko: I’ve seen the VCS/Verilator linters complain about it

gaozihou: Okay, I will fix it.

gaozihou:
32'b1 is wrong answer which causes warning in lint. I fixed it!

bsg_link/bsg_link_oddr_phy.v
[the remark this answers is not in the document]
gaozihou: Some reference on this: https://docs.google.com/document/d/1lmkOxvlAvxrk_MM5W8xv3ho2DS26xbOMTCqUIyS6di8/edit# Page 4

bsg_link/bsg_link_oddr_phy.v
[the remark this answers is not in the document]
gaozihou: Some reference on this: https://docs.google.com/document/d/1lmkOxvlAvxrk_MM5W8xv3ho2DS26xbOMTCqUIyS6di8/edit# Page 4

bsg_link/bsg_link_source_sync_downstream.v
gaozihou: Here is a comparison of SSI vs Downstream and SSO vs Upstream: https://github.com/gaozihou/basejump_stl/commit/25a81d423d09eea3d90ba4bc3174f46915fe50da#diff-0477b7643b576e366d6daf0a5d7a5439

gaozihou:
Most of the issues are resolved, except for the x1 x2 clock ports and bsg_parallel_in_serial_out and bsg_parallel_out_serial_in (will be in another pull request).

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,475 @@
...
+  ,input multi_v_i
+  ,output multi_ready_o
+
+  // outgoing multiplexed data
+  ,output [width_p-1:0] multi_data_o
+  ,output multi_v_o
+  ,input multi_yumi_i
+
+  // demultiplexed data
+  ,input [num_in_p-1:0][bsg_ready_and_link_sif_width_lp-1:0] link_i
```
taylor-bsg: align things that are vertically the same to improve code symmetry and readability.

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,475 @@
...
+ #(.width_p(raw_width_lp)
+  ,.num_in_p(num_in_p)
+  ,.remote_credits_p(remote_credits_p)
+  ,.use_pseudo_large_fifo_p(1)
+  ,.lg_credit_decimation_p(lg_credit_decimation_p))
+  channel_tunnel
+  (.clk_i(clk_i)
+  ,.reset_i(reset_i)
+  // outside
+  ,.multi_data_i(outside_data_li)
```
taylor-bsg: vertical alignment of (outside ...)'s and (inside..)'s

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,475 @@
...
+  `declare_bsg_ready_and_link_sif_s(width_p,bsg_ready_and_link_sif_s);
+  bsg_ready_and_link_sif_s [num_in_p-1:0] link_i_cast, link_o_cast;
+  
+  for (i = 0; i < num_in_p; i++) 
+  begin
+  
+    assign link_i_cast[i] = link_i[i];
+    assign link_o[i] = link_o_cast[i];
+
+    assign v_i[i] = link_i_cast[i].v;
```
taylor-bsg: align

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,475 @@
...
+  bsg_mux 
+ #(.width_p(1)
+  ,.els_p(mux_num_in_lp))
+  out_v_mux
+  (.data_i(ofifo_valid_o)
+  ,.sel_i(mux_sel_r)
+  ,.data_o(multi_v_o));
+  
+  for (i = 0; i < mux_num_in_lp; i++) 
+  begin
```
taylor-bsg: generate statement label

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,475 @@
...
+    assign ofifo_yumi_i[i] = (i==mux_sel_r)? multi_yumi_i : 0;
+  end
+  
+  always_comb 
+  begin
+    
+    ostate_n = ostate_r;
+    mux_sel_n = mux_sel_r;
+    
+    if (multi_yumi_i) begin
```
taylor-bsg: begins on newline, so they align with end's

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,475 @@
...
+  (.data_i(ofifo_valid_o)
+  ,.sel_i(mux_sel_r)
+  ,.data_o(multi_v_o));
+  
+  for (i = 0; i < mux_num_in_lp; i++) 
+  begin
+    assign ofifo_yumi_i[i] = (i==mux_sel_r)? multi_yumi_i : 0;
+  end
+  
+  always_comb
```
taylor-bsg: comment describing what this block does.

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,475 @@
...
+  // Wormhole header flits and credit returning flit should go into channel tunnel.
+  // Data flits should go into corresponding ififos.
+  // Destination is selected by "reserved bits" generated by sender.
+  
+  logic [tag_width_lp-1:0] in_sel_r, in_sel_n;
+  logic [len_width_p-1:0] istate_r, istate_n;
+  
+  always_ff @(posedge clk_i) 
+  begin
+    if (reset_i) begin
```
taylor-bsg: align begin and end

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,475 @@
+//
+// Paul Gao 03/2019
+// 
+// This module is a special version bsg_channel_tunnel that accepts wormhole packet
+//
+// Typical usage: When there are n wormhole routers (n wormhole networks)
+// and need to merge them together to the IO channel (n >= 2)
+// All wormhole packet parameters should match corresponding wormhole network
+//
```
taylor-bsg: Describe what algorithm the module is using to interleave the wormhole packets. How does it track credits, and how does it design how to interleave packets?

gaozihou:
Done.

PR #62, opened 2019-06-02
(Deferred) Adding dynamic load counter

dpetrisko:
I was going to use this as the BP RTC, but added the down counter for genericism

### DIFF
--- a/bsg_misc/bsg_counter_dynamic_load.v
+++ b/bsg_misc/bsg_counter_dynamic_load.v
@@ -1,0 +1,33 @@
+//
+// This is a counter which can count up or down and supports loading 
+//   the current counter value. Upon a load, the value in the
+//   next cycle will be the loaded value (not +1)
+//
+
+module bsg_counter_dynamic_load
+ #(parameter width_p = "inv"
+
+  // Up is default
+  , parameter down_not_up_p = 0
+  )
+ (input                        clk_i
+  , input                      reset_i
+
+  , input                      w_v_i
+  , input        [width_p-1:0] data_i
+
+  , output logic [width_p-1:0] counter_o
+  );
+
+  always_ff @(posedge clk_i)
+    if (reset_i)
+      counter_o <= '0;
+    else if (w_v_i)
+      counter_o <= data_i;
+    else if (down_not_up_p)
+      counter_o <= counter_o - 1'b1;
+    else
+      counter_o <= counter_o + 1'b1;
+
+endmodule
+


stdavids:
https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_misc/bsg_counter_overflow_set_en.v

Seems very similar to bsg_counter_overflow_set_en except it has the up_down capability. Excluding the up down, I believe this counter is the same as:

```
module bsg_counter_dynamic_load
 #(parameter width_p = "inv"

  // Up is default
  , parameter down_not_up_p = 0
  )
 (input                        clk_i
  , input                      reset_i

  , input                      w_v_i
  , input        [width_p-1:0] data_i

  , output logic [width_p-1:0] counter_o
  );

bsg_counter_overflow_set_en #(.max_val_p( (2**width_p)-1 ))
  foobar
    (.clk_i(clk_i)
    ,.en_i(1'b1)
    ,.set_i( w_v_i | reset_i )
    ,.val_i( reset_i ? '0 : data_i )
    ,.count_o(counter_o)
    ,.overflow_o()
    );

endmodule
```

Someone double check this, but if that is correct maybe we just add the up/down to that counter?

dpetrisko:
Oh, I missed that one.  So the differences are:

- Minus enable
- Plus reset
- Minus overflow 
- Up/down capability

I think your solution should work.  It's kind of messy.  I think the problem is that we're approaching adding counters as sampling the cross-product.  We should be able to make a generator which creates the full taxonomy of counters and then select based on parameters.

Another approach would be to add the functionality to this:
https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_misc/bsg_cycle_counter.v
which is more related to what I'm trying to do, anyway.

taylor-bsg:
There does seem to be a big taxonomy of counters. I don't think we need a generator per se; mostly just a consistent naming system. What is this counter used for in this BP?

dpetrisko:
RTC. The only difference between this and the bsg_cycle_counter is that we need a way to dynamically set it.

taylor-bsg:
The path forward on this is more or less to do what dan suggested but in a slightly more consistent way with the other counters. We currently have:

https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_misc/bsg_counter_set_down.v

We can create bsg_counter_set_up.v, and factor the code out of the two modules into bsg_counter_set_up_or_down which takes a parameter down_not_up or the similar.

An important difference is that these modules allow simultaneous set and up assertion. This is aligned with the philosophy that to the extent that two input transactions are exclusive (or that one command is ignored if another command is asserted), an enum should be used so that it is impossible to express the unsupported simultaneous activity.

M

PR #63, opened 2019-06-04
Wormhole Network Standalone

gaozihou:
Pull request for standalone modules:

Modifying 1 current files:
bsg_noc_links.vh: Add wormhole packet data structure.

Adding 6 files:
bsg_serial_in_parallel_out_full_buffered.v: This de-serializer has 0 bubble in transition. It utilizes bsg_two_fifos inside as memory.
bsg_link_source_sync_downstream.v and bsg_link_source_sync_upstream.v: The new "common-link" modules, utilizing async-reset on token.
bsg_link_iddr_phy.v and bsg_link_oddr_phy.v: The DDR PHY for ASIC.
bsg_counter_dynamic_start_en_down.v: New count-down counter that has dynamic start value.

The corresponding testbench is on bsg_manycore repo: bsg_manycore/testbenches/manycore_loopback_ddr_link

### DIFF
--- a/bsg_dataflow/bsg_serial_in_parallel_out_full_buffered.v
+++ b/bsg_dataflow/bsg_serial_in_parallel_out_full_buffered.v
@@ -1,0 +1,103 @@
+/**
+ *  bsg_serial_in_parallel_out_full_buffered.v
+ *
+ *  This is a simpler version of bsg_serial_in_parallel_out.
+ *  Output is only valid, when the output vector is fully assembled.
+ *  This version has zero bubble.
+ *
+ */
+
+module bsg_serial_in_parallel_out_full_buffered
+
+ #(parameter width_p     ="inv"
+  ,parameter els_p       ="inv"
+  ,parameter msb_first_p = 0
+  )
+  
+  (input clk_i
+  ,input reset_i
+    
+  ,input                                 v_i
+  ,output logic                          ready_o
+  ,input [width_p-1:0]                   data_i
+
+  ,output logic [els_p-1:0][width_p-1:0] data_o
+  ,output logic                          v_o
+  ,input                                 yumi_i
+  );
+  
+  localparam lg_els_lp     = `BSG_SAFE_CLOG2(els_p);
+  localparam end_count_lp  = (msb_first_p == 0)? els_p-1 : 0;
+  localparam init_count_lp = (msb_first_p == 0)? 0 : els_p-1;
+  
+  genvar i;
+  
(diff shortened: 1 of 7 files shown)


bsg_dataflow/bsg_serial_in_parallel_out_full_buffered.v
```diff
@@ -0,0 +1,94 @@
...
+  logic [els_p-1:0] fifo_valid_li, fifo_ready_lo;
+  logic [els_p-1:0] fifo_valid_lo, fifo_yumi_li;
+  
+  assign ready_o = fifo_ready_lo[counter_r];
+  assign v_o     = & fifo_valid_lo;
+  
+  for (i = 0; i < els_p; i++) 
+  begin: fifos
+    
+    assign fifo_valid_li[i] = (i == counter_r) & v_i;
```
tommydcjung: (i == counter_r) & v_i is a pretty common expression, and bsg_decode_with_v can be used for that.

bsg_dataflow/bsg_serial_in_parallel_out_full_buffered.v
```diff
@@ -0,0 +1,94 @@
...
+    ,.yumi_i (fifo_yumi_li[i])
+    );
+    
+  end
+  
+  logic end_count_lo, count_up_lo, count_down_lo;
+  
+  always_comb 
+  begin
+    end_count_lo  = 0;
```
tommydcjung: 1'b0 instead of 0

bsg_misc/bsg_counter_dynamic_start_en_down.v
```diff
@@ -0,0 +1,38 @@
...
+  ,input        [width_p-1:0] start_i
+  ,output logic [width_p-1:0] counter_o
+  ,output                     is_last_val_o
+  );
+
+  assign is_last_val_o = (counter_o == last_val_p);
+  
+  always_ff @ (posedge clk_i)
+    if (reset_i)
+        counter_o <= last_val_p;
```
tommydcjung: seems weird to have it start at the last val. Shouldn't it reset to start_val? but then the start_i is dynamic..

stdavids:
I suggest closing this PR. There is still too many different kinds of modules here. We should really have a PR for the SIPO, a PR for the counter, a PR for the phys, and a PR for the upstream/downstream. Even those two can be split. Some of these are really complicated and we are going to run into a PR #60 mess again with hundreds of comments.

PR #65, opened 2019-06-04
Add Generic Counter

gaozihou:
This is developed based on bsg_counter_dynamic_limit_en.v (dynamic-end up-counter). It starts counting from user-specified start value (start_i), ends at fixed value (last_val_p).


dpetrisko:
Can we add up_not_down_p to this counter? Just to be more general with no real overhead?

gaozihou:
@dpetrisko interesting suggestion! Then this will be "dynamic start fixed end" counter that supports both up and down. The module name "bsg_counter_dynamic_start_en_down" may need to change then...

dpetrisko:
dynamic_start_fixed_end sounds good to me!

taylor-bsg:
I think the issue is that all of these counters have slightly different behavior, and it's hard to tell from the module name what the differences in the behavior are, and from the user of a module, it's not clear whether they needed the strange functionality, or whether they just used the first one they found.  For example, in Paul's module, he doesn't actually use the output value of the counter, and the lower limit is zero; it doesn't really need to be a settable value. I am reminded a little bit of for loops in C, which have their own issues (<=, !=, ==); never seem to get it right if you are counting down, etc.

Here is my attempt at a minimally ambiguous interface. It assumes that the synthesizer will do okay optimizing constants across module boundaries (true these days with DC). The only state stored inside the module is the active counter. It is up to the user to place flops to hold the stop value in place with a register if it is not a constant.

bsg_counter_cmd_s --> copy_start_to_cur_val
bsg_counter_cmd_s --> step
bsg_counter_cmd_s --> noop

module bsg_counter_generic #(width_p, up_not_down_p, never_stop_p)
(
input clk_i
,input bsg_counter_cmd_s cmd_i
,input [width_p-1] start_val_i
,output[width_p-1] cur_val_o
,input [width_p-1] stop_val_i
,output cur_val_at_stop_val_o
);

So then Dan's dynamic load counter is:

bsg_counter_generic #(.up_not_down_p(1),.never_stop_p(1)) dan_ctr
(.clk_i, .cmd_i(w_en? copy_start_to_cur_val : step), .start_val_i(GPR[]), ,. stop_val_i(), .cur_val(curVal), cur_val_at_stop_val_o());

bsg_counter_generic #(.up_not_down_p(0)) paul_ctr
(.clk_i, .cmd_i(new_packet ? copy_start_to_cur_val : sending_packet & ready_o ? step : nop ),. start_val_i(top_val),. stop_val_i('0),. cur_val(), .cur_val_at_stop_val_o(done_lo));

dpetrisko:
Looks good to me. If we embrace parameter-based port stubbing, I think we can clean up quite a few module sets in this way.

Possibly it makes sense to add a file containing wrappers for common configurations.

gaozihou:
Agree prof! Inspired by your idea I come up with a simpler one. The purpose is to make it as lightweight as possible, as easy-to-use as possible. This one can potentially substitute most of the counters in bsg_misc directory.

module bsg_counter_generic
 #(parameter max_val_p      = "inv"
  ,parameter max_step_p     = "inv"
  ,parameter init_val_p     = "inv"
  ,parameter count_on_set_p = 0 // Just to minimize hardware
  ,localparam val_width_lp  = BSG_SAFE_CLOG2(max_val_p) 
  ,localparam step_width_lp = BSG_SAFE_CLOG2(max_step_p) 
  )
  (input                      clk_i
  ,input                      reset_i
  ,input                      en_i
  ,input                      set_i
  ,input  [step_width_lp-1:0] up_i
  ,input  [step_width_lp-1:0] down_i
  ,input  [val_width_lp-1:0]  set_val_i
  ,output [val_width_lp-1:0]  cur_val_o
  );

Dan's counter will be like:
bsg_counter_generic
 #(.max_val_p(?),.max_step_p(1),.init_val_p(0))
dan_counter
  (.clk_i(),. reset_i()
  ,.en_i(1'b1)
  ,.set_i(write_en)
  ,.up_i(1'b1)
  ,.down_i(1'b0)
  ,.set_val_i(data_i)
  ,.cur_val_o(counter_o)
  );

My counter will be like:
bsg_counter_generic
 #(.max_val_p(?),.max_step_p(1),.init_val_p(end_val))
dan_counter
  (.clk_i(),. reset_i()
  ,.en_i(v_i&ready_o)
  ,.set_i(counter_o==end_val)
  ,.up_i(1'b0)
  ,.down_i(1'b1)
  ,.set_val_i(new_val_lo)
  ,.cur_val_o(counter_o)
  );

PR #66, opened 2019-06-04
Add bsg_link source sync input / output

gaozihou:
This pair of modules are developed based on bsg_comm_link/bsg_source_sync_input and bsg_comm_link/bsg_source_sync_output.

### DIFF
--- a/bsg_link/bsg_link_source_sync_downstream.v
+++ b/bsg_link/bsg_link_source_sync_downstream.v
@@ -1,4 +1,9 @@
-// MBT 7/24/2014
+//
+// Prof. Taylor   7/24/2014
+// <prof.taylor@gmail.com>
+//
+// Updated by Paul Gao 02/2019
+//
 // DDR or center/edge-aligned SDR source synchronous input channel
 //
 // this implements:
@@ -6,10 +11,11 @@
 //     async fifo to go from source-synchronous domain to core domain
 //     outgoing token channel to go from core domain deque to out of chip
 //     outgoing source-synchronous launch flops for token
-//     programmable capture on either or both edges of the clock
 //
-// note, the default FIFO depth is set to 2^5
-// based on the following calculation:
+// note, the default FIFO depth is set to 2^6 based on experiments on FPGA
+// FIXME: update these numbers based on clocks in each clock domain and from actual waveforms.
+//
+// Below is a rough calculation:
 //
 // 2 clks for channel crossing
 // 3 clks for receive fifo crossing
@@ -22,7 +28,7 @@
 // -----------
 // 19 clks
 //
-// This leaves us with 13 elements of margin
+// This leaves us with 45 elements of margin
 // for FPGA inefficiency. Since the FPGA may run
 // at 4X slower, this is equivalent to 3 FPGA cycles.
 //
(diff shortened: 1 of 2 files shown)


gaozihou:
I've updated the branch to "diff" between files. 
Same parts:
1. Token send / receive, credit counter.
2. Async fifo / twofer fifo
Newly added:
1. Core / IO reset control (previously reset is controlled by comm-link-kernel)
2. Token async reset
3. interface to ODDR PHY (upstream), interface to IDDR PHY (downstream)

bsg_link/bsg_link_source_sync_downstream.v
```diff
@@ -222,109 +132,68 @@ module bsg_source_sync_input #(parameter lg_fifo_depth_p=5
...
+
+  ,.v_o    (core_valid_o)
+  ,.data_o (core_data_o)
+  ,.yumi_i (core_yumi_i)
+  );
 
 
    // a word was transferred to the two input fifo if ...
    wire core_transfer_success = core_valid_o_tmp & core_twofer_ready;
+   assign core_actual_deque = core_transfer_success;
```
taylor-bsg: Unnecessary?

bsg_link/bsg_link_source_sync_downstream.v
```diff
@@ -32,188 +34,96 @@
...
+  
+  (// control signals
+   input                        core_clk_i
+  ,input                        core_link_reset_i
+
+  // source synchronous input channel; coming from chip edge
+  ,input                        io_clk_i     // sdi_sclk
+  ,input  [channel_width_p-1:0] io_data_i    // sdi_data
+  ,input                        io_valid_i   // sdi_valid
+  ,output                       io_token_r_o // sdi_token; output registered
```
taylor-bsg: No need to transfer this signal back to the io clock domain from the core clock domain. We should compute token toggles in core.

bsg_link/bsg_link_source_sync_downstream.v
```diff
@@ -222,109 +132,68 @@ module bsg_source_sync_input #(parameter lg_fifo_depth_p=5
...
-                                           : core_sent_0_want_to_send_1_r;
-     end
 
 // **********************************************
 // credit return
 //
 // these are credits coming from the receive end of the async fifo in the core clk
 //  domain and passing to the io clk domain and out of the chip.
 //
+
```
taylor-bsg: After this point code is unnecessary. Just move the credit decimation counter into this module and drive off of the core clock domain.

taylor-bsg: Can use bsg cycle counter and key off of the top bit.

bsg_link/bsg_link_source_sync_downstream.v
```diff
@@ -32,188 +34,96 @@
...
-              io_trigger_count_r <= 2**(lg_io_trigger_words_lp);
-       end
 
+module bsg_link_source_sync_downstream 
+
+ #(parameter channel_width_p                 = 16
+  ,parameter lg_fifo_depth_p                 = 6
+  ,parameter lg_credit_to_token_decimation_p = 3
+  
+  // Reset pattern is 0xF*
```
taylor-bsg: I thought we were just going remove these reset pattern things and use bsg_tag sequences resets?

bsg_link/bsg_link_source_sync_downstream.v
```diff
@@ -32,188 +34,96 @@
...
+  );
+  
+  //  
+  // remote reset signal coming from bsg_link_source_sync_upstream
+  // remote reset is asserted when source_sync_upstream in reset OR not enabled
+  //
+  logic io_remote_reset_lo, core_remote_reset_lo;
+  assign io_remote_reset_lo = (io_data_i==reset_pattern_p && ~io_valid_i);
+  
+  // remote reset signal from io clock to core clock
```
taylor-bsg: I suggest removing all synchronizers from this module, and specifying that there are two input reset signals, one for core and one for i/o that have already been synchronized? This will ensure that we only have one reset signal in each clock domain.

gaozihou: Hi prof, it is possible to have "one reset signal per clock domain", but we will need to add a separate "token_async_reset_i". This is because the "async reset" and ""token counter reset" are different signals.

gaozihou: Also, we must have "one sets of reset signals per bsg_link_ddr", since we may want to disable one-pair of them. (Imagine our chip can do chained-communication but we only connect one side)

bsg_link/bsg_link_source_sync_downstream.v
gaozihou: How about using the newly-proposed bsg_counter_generic?

bsg_link/bsg_link_source_sync_downstream.v
gaozihou: We might want one reset synchronizer in this module: from core_domain to io_domain. This io_domain reset is only used within this module, so it should be better to generate locally, what do you think prof?

taylor-bsg: It is a logical argument indeed, but snaking resets through clock domains is one of the trickiest part about verifying a design, because you have to consider the effects of multiple clock domains on the reset signal, and all of the relative frequencies they could be running at. if you snake I/O reset from core, then the core clock needs to be running in order for reset to occur. It also prevents us from controlling the precise ordering of the I/O reset signal relative to other resets.

bsg_link/bsg_link_source_sync_downstream.v
taylor-bsg: bsg_counter_clear_up is custom made for this use case..

bsg_link/bsg_link_source_sync_upstream.v
```diff
@@ -38,247 +34,216 @@ module bsg_source_sync_output
...
-
-    // token reset rules:
-    //    token_clk_i should be 010'd while token_reset_i is asserted
-    //    io_master_clk_i should have > 4 posedges after that.
-    //    token_reset_i is not actually synchronous to token_clk_i
-    //    so it should be asserted well before token_clk_i is asserted
-    //    and de-asserted well afterwards to avoid metastability
-
-    , input                         token_reset_i    // token fifo reset
+    // source synchronous output channel to chip edge
```
taylor-bsg: output channel to ODDR_PHY

bsg_link/bsg_link_source_sync_upstream.v
```diff
@@ -38,247 +34,216 @@ module bsg_source_sync_output
...
+  // w.c. is cycleTimeB == cycleTimeA
+  //
+  // --> 6 elements
+  //
+  // however, for the path from A to B and B to A
+  // we need to clear not only the cycle time of A/B
+  // but the two setup times, which are guaranteed to
+  // be less than a cycle each. So we get 8 elements total.
+  //
+
```
taylor-bsg: Add: 
// TODO: if we use 3-cycle synchronizers, then the async fifo would have to be larger.

bsg_link/bsg_link_source_sync_downstream.v
gaozihou: Thanks prof! Do you think we should do the same thing on Upstream? Removing all synchronizers, expose core_reset_i / io_reset_i / token_async_reset_i to the outside world.

bsg_link/bsg_link_source_sync_downstream.v
gaozihou: Thanks prof. I tried bsg_counter_clear_up but gets huge amount of overflow warnings. Seems that it does not allow overflowing.

I think we should allow overflow to minimize hardware? (if reset counter periodically, we need one more comparison between max value and current value)

bsg_link/bsg_link_source_sync_downstream.v
gaozihou: I removed all synchronizers inside Source Sync Upstream, now io_master_reset and core_reset all are exposed to outside world (bsg_tag). 

For Source Sync Downstream, I find out that the only place uses io_reset is async fifo... So it makes more sense to leave one synchronizer there?
1. When we reset async fifo, we must have both io clock and core clock running
2. This "io_clk" is unique for each IO channel, so it is okay to be generated locally?

bsg_link/bsg_link_source_sync_downstream.v
```diff
@@ -326,86 +157,22 @@ module bsg_source_sync_input #(parameter lg_fifo_depth_p=5
...
-          // we need this to rate match the incoming data
-
-	  // code is written like this because empty_1 is late relative to empty_0
-          io_credits_sent_r <= (empty_1
-                                ? (empty_0 ? io_credits_sent_r_p2 : io_credits_sent_r)
-                                : (empty_0 ? io_credits_sent_r_p1 : io_credits_sent_r));
-     end
-
-endmodule // bsg_source_sync_input
+  logic [lg_credit_to_token_decimation_p+1-1:0] core_credits_sent_r;
```
taylor-bsg: bsg_counter_clear_up

gaozihou: I tried bsg_counter_clear_up but gets huge amount of overflow warnings. Seems that it does not allow overflowing.

I think we should allow overflow to minimize hardware? (if clear counter periodically, we need one more comparison between max value and current value)

bsg_link/bsg_link_source_sync_upstream.v
```diff
@@ -37,248 +34,185 @@ module bsg_source_sync_output
...
+   
+   // when fifo has valid data and token credit is available
+   logic io_fifo_valid_credit_avail;
+   
+   always_comb
      begin
-        io_reset_r   <= io_reset_i;
-        if (io_override_en_i)
-          { io_valid_r_o, io_data_r_o }   <= io_override_valid_data_i;
+        if (io_master_link_reset_i)
```
taylor-bsg: pull io_data_o = inactive_pattern_p out of if statement.

bsg_link/bsg_link_source_sync_downstream.v
taylor-bsg: okay I added a parameter to disable the warning.

bsg_link/bsg_link_source_sync_downstream.v
```diff
@@ -32,292 +34,121 @@
...
+  ,output                       core_token_r_o // sdi_token; output registered
+
+  // going into core; uses core clock
+  ,output [channel_width_p-1:0] core_data_o
+  ,output                       core_valid_o
+  ,input                        core_yumi_i
+  );
+  
+  
+  // io_reset only for async-fifo reset purposes
```
taylor-bsg: yes, remove this and make io_reset_ be an input

gaozihou: Sure!

bsg_link/bsg_link_source_sync_downstream.v
```diff
@@ -32,292 +34,121 @@
...
+  ,.w_reset_i(io_reset_lo)
+  
+  ,.w_enq_i  (io_valid_i)
+  ,.w_data_i (io_data_i)
+  ,.w_full_o (io_async_fifo_full)
 
-    ,.r_clk_i(core_clk_i)
-    ,.r_reset_i(core_reset_i)
+  ,.r_clk_i  (core_clk_i)
+  ,.r_reset_i(core_reset_lo)
```
taylor-bsg: core_reset_i

bsg_link/bsg_link_source_sync_downstream.v
```diff
@@ -32,292 +34,121 @@
...
+  // going into core; uses core clock
+  ,output [channel_width_p-1:0] core_data_o
+  ,output                       core_valid_o
+  ,input                        core_yumi_i
+  );
+  
+  
+  // io_reset only for async-fifo reset purposes
+  // should be asserted or deasserted together with core_reset
+  logic core_reset_lo, io_reset_lo;
```
taylor-bsg: remove core_reset_lo everywhere and just use core_reset_i.  this is not an intended usage of the iclk_data_o port because we are supposed to minimize the load on the synchronizer launch flop to avoid timing issues -- it is more for feedback with the iclk_data_i input.

bsg_link/bsg_link_source_sync_upstream.v
```diff
@@ -37,248 +34,184 @@ module bsg_source_sync_output
        // and the ratio of signal to I/O V33 and VZZ is
        // 4:1:1.
        //
-
+       
        // fixme: an alternative might be to tri-state
        // the output, but further analysis is required
        // as to whether this is a good idea.
 
        // This has implications for calibration;
```
taylor-bsg: delete comments about calibration since we don't support that

bsg_link/bsg_link_source_sync_upstream.v
```diff
@@ -37,248 +34,184 @@ module bsg_source_sync_output
...
+  ,.data_i (core_data_i)
+  ,.v_i    (core_valid_i)
+
+  ,.v_o    (core_fifo_valid)
+  ,.data_o (core_fifo_data)
+  ,.yumi_i (core_fifo_yumi)
+  );
+  
+  
+  logic core_fifo_full;
```
taylor-bsg: should be core_async_fifo_full

bsg_link/bsg_link_source_sync_upstream.v
```diff
@@ -37,248 +34,184 @@ module bsg_source_sync_output
...
-   (
-    // going out of core; uses core clock
-    input                           core_clk_i
-    , input                         core_reset_i
+       
+   (// control signals  
+      input                         core_clk_i
+    , input                         core_link_reset_i
+    
+    , input                         io_master_clk_i
```
taylor-bsg: for consistency, rename io_master_* to io_*

bsg_link/bsg_link_source_sync_upstream.v
```diff
@@ -37,248 +34,184 @@ module bsg_source_sync_output
...
        // This has implications for calibration;
-       // specifically v=0,d=5* should not be used.
+       // specifically v=0,d=A* should not be used.
        // an alternative to this inactive pattern is to
        // invert alternating bits or to apply a scramble
-       //  pattern. We keep it simple.
-
-       , parameter inactive_pattern_p = {channel_width_p { 2'b01 } }
+       // pattern. We keep it simple.
+
```
taylor-bsg: // this default is only safe because we assume that only data bits are 
// being specified in the channel and no control bits, and that these data bits
// are otherwise ignored by the receiver control logic if the output of this module is indicated
// as invalid.

bsg_link/bsg_link_source_sync_upstream.v
```diff
@@ -318,47 +246,49 @@ module bsg_source_sync_output
...
+        ,.w_reset_i(token_reset_lo)
 
         // the I/O clock domain is responsible for tabulating tokens
-        ,.r_clk_i  (io_master_clk_i)
-        ,.r_reset_i(io_reset_i     )
+        ,.r_clk_i             (io_clk_i                )
+        ,.r_reset_i           (io_link_reset_i         )
         ,.r_dec_credit_i      (io_posedge_credits_deque)
-        ,.r_infinite_credits_i(io_infinite_credits_i)
+        ,.r_infinite_credits_i(io_link_reset_i         )
```
taylor-bsg: r_infinite_credits_i should be 1'b0?

bsg_link/bsg_link_source_sync_upstream.v
```diff
@@ -318,47 +246,49 @@ module bsg_source_sync_output
...
 
         // the I/O clock domain is responsible for tabulating tokens
-        ,.r_clk_i             (io_master_clk_i         )
-        ,.r_reset_i           (io_reset_i              )
+        ,.r_clk_i             (io_clk_i                )
+        ,.r_reset_i           (io_link_reset_i         )
         ,.r_dec_credit_i      (io_negedge_credits_deque)
-        ,.r_infinite_credits_i(io_infinite_credits_i)
-        ,.r_credits_avail_o(io_negedge_credits_avail)
+        ,.r_infinite_credits_i(io_link_reset_i         )
```
taylor-bsg: r_infinite_credits_i should be 1'b0?

bsg_link/bsg_link_source_sync_upstream.v
```diff
@@ -37,248 +34,177 @@ module bsg_source_sync_output
...
-    // there is data in the FIFO
-    ,.r_valid_o(io_data_avail)
-    );
-
-   // this can easily happen if the sending core clock
-   // is higher than the I/O clock
-   // always @(negedge core_clk_i)
-   //  assert (~(core_fifo_full===1))
-   //   else $error("source synchronous output FIFO unexpectedly full");
+    //
```
taylor-bsg: move details of reset to the top of the file. also, this description seems less precise than this:

https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_async/bsg_async_credit_counter.v

// ASYNC RESET PROCEDURE:
// Step 1: Assert r_ reset.
// Step 2: w_ reset must be posedge/negedge toggled (0->1->0) at least once.
//         w_ clock cannot toggle during this step.
// Step 3: r_ clock posedge toggled at least four times after that.
// Step 4: Deassert r_ reset.

bsg_link/bsg_link_source_sync_upstream.v
```diff
@@ -37,248 +34,177 @@ module bsg_source_sync_output
...
+    //    Initial value       0                  0                1
+    //    Step 1              1                  1                1
+    //    Step 2              0                  0                1
+    //    Step 3              0                  0                0
+    // *******************************************************************
+    // As shown above, token_reset_lo signal has a 0->1->0 pattern, where the
+    // posedge (0->1) resets the token counter.
+    //
+    
+    logic token_reset_lo;
```
taylor-bsg: replace this signal with async_token_reset_i, which is an input parameter to the module

bsg_link/bsg_link_source_sync_downstream.v
```diff
@@ -1,15 +1,17 @@
...
 //     async fifo to go from source-synchronous domain to core domain
 //     outgoing token channel to go from core domain deque to out of chip
 //     outgoing source-synchronous launch flops for token
-//     programmable capture on either or both edges of the clock
 //
-// note, the default FIFO depth is set to 2^5
-// based on the following calculation:
+// note, the default FIFO depth is set to 2^6 based on experiments on FPGA
+// Below is a rough calculation:
 //
```
taylor-bsg: Add 

// FIXME: update these numbers based on clocks in each clock domain and from actual waveforms.

bsg_link/bsg_link_source_sync_downstream.v
```diff
@@ -32,292 +34,105 @@
...
-    .w_clk_i(io_clk_i)
-    ,.w_reset_i(io_reset_i)
-    ,.w_enq_i(io_async_fifo_enq)
-    ,.w_data_i(io_trigger_mode_en_i
-               ? { 1'b1                       , 1'b1
-                   , io_data_1_r_swizzle      , io_data_0_r_swizzle}
-
-               : { io_valid_1_r & io_edge_i[1], io_valid_0_r & io_edge_i[0]
-                  , io_data_1_r               , io_data_0_r                })
+   wire  core_actual_deque, core_valid_lo;
```
taylor-bsg: Sorry for the nitpickiness. Since we have two fifos, it's a little confusing. I propose:

core_async_fifo_deque, core_async_fifo_valid_lo, core_async_fifo_data_lo

This will be easier to follow.

gaozihou:
Thanks prof! information added to commit message.

PR #68, opened 2019-06-04
Add serial in parallel out full --- buffered version

gaozihou:
Q: Why do we want SIPO_Full?
A: 1. Power might be better, no shifting-based fifo 2. Hardware might be simpler.

Q: Why do we want **buffered version** SIPOF?
A: Original SIPOF has one cycle bubble after reassembling each packet. The **buffered version** SIPOF has two-element fifo inside, so no more stalling.


gaozihou:
Thanks for the kind suggestions and help prof!

1. Seems that life gets much easier with bsg_round_robin_1_to_n...
2. I agree, we only need the lowest word fifo to be two-element. This will save a lot of area!
3. From my understanding yes, the behavior and interface is exactly the same. Would be helpful if @tommydcjung could help double check!

gaozihou:
I've updated the module, with round_robin_1_to_n it looks much cleaner! I also slightly modify the bsg_round_robin_1_to_n to resolve the port width mismatch warning (let it match bsg_circular_ptr). 

Should I change the file name and replace the previous module?

taylor-bsg:
Looks great.  Can you add a comment in front of line 41 of bsg_round_robin_1_to_n that says

// bsg_decode_with_v could potentially be used to optimize this critical path at the cost of area

Feel free to change the file name and replace the old module.

@tommydcjung can you confirm that none of your code relies upon the bubble being there? =)

bsg_dataflow/bsg_serial_in_parallel_out_full.v
```diff
@@ -2,72 +2,122 @@
...
-  // counter
+  // Full array is valid when all fifos have valid data
+  assign v_o = & fifo_valid_lo;
+  
+  // Push received data into fifos in round-robin way
   //
-  logic [counter_width_lp-1:0] count_lo;
-  logic clear_li;
-  logic up_li;
+  // bsg_decode_with_v could potentially be used to optimize this critical path
```
taylor-bsg: I meant add that comment to the decoder line inside the bsg_round_robin_1_to_n module (around line 41). Thanks again!

taylor-bsg: @gaozihou

PR #71, opened 2019-06-07
Channel Tunnel for Wormhole

gaozihou:


bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,595 @@
...
+  // Default value matches child module bsg_channel_tunnel
+  // Channel tunnel return credits to sender every (2^lg_credit_decimation_p) packets
+  // This is independent of how many flits each wormhole packet has
+  ,parameter lg_credit_decimation_p = `BSG_MIN($clog2(remote_credits_p+1),4)
+  
+  // Local parameters
+  
+  // Initial value for wormhole flit counters
+  // Since wormhole length is (num_flits-1), counter should count from 0
+  ,localparam counter_min_value_lp = 0
```
taylor-bsg: move these _lp's out of the module interface, since they are not required by the interface

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,595 @@
...
+  // Local parameters
+  
+  // Initial value for wormhole flit counters
+  // Since wormhole length is (num_flits-1), counter should count from 0
+  ,localparam counter_min_value_lp = 0 
+  
+  ,localparam mux_num_in_lp = num_in_p+1
+  ,localparam tag_width_lp = $clog2(mux_num_in_lp)
+  ,localparam raw_width_lp = width_p-tag_width_lp
+  ,localparam len_offset_lp = width_p-reserved_width_p-x_cord_width_p-y_cord_width_p-len_width_p
```
taylor-bsg: a. should not be in module interface
b. should not be computing this value like this in this module. what if the layout of the struct changes?  You should use the macro to define the struct, then cast the packet to the struct, and then access the fields using the struct access notation.

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,595 @@
...
+  
+  logic [num_in_p-1:0] inside_valid_lo, inside_yumi_li;
+  logic [num_in_p-1:0][raw_width_lp-1:0] inside_data_lo;
+
+  
+  bsg_channel_tunnel
+ #(.width_p(raw_width_lp)
+  ,.num_in_p(num_in_p)
+  ,.remote_credits_p(remote_credits_p)
+  ,.use_pseudo_large_fifo_p(1)
```
taylor-bsg: should have a comment justifying why it is okay to use a pseudo-large fifo that only has one port on the SRAM.

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,595 @@
...
+  ,parameter num_in_p = "inv"
+  
+  // Max number of wormhole packets buffer can store
+  // This is independent of how many flits each wormhole packet has
+  ,parameter remote_credits_p = "inv"
+  
+  // Max possible "wormhole packet payload length" setting
+  // An n-flits wormhole packet has 1 header flit and (n-1) payload flits
+  // This parameter determines the size of payload-flits buffer
+  ,parameter max_payload_flits_p = "inv"
```
taylor-bsg: does this really need to be passed in, or can it be computed with a macro?

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,595 @@
...
+// Typical usage: When there are n wormhole routers (n wormhole networks)
+// and need to merge them together to the IO channel (n >= 2)
+// All wormhole packet parameters should match corresponding wormhole network
+//
+// Packet interleaving algorithm:
+//
+// When multiple source channels are valid at the same time, channel tunnel will interleave
+// whole-packets in round-robin way. "whole-packet" means that flits of each wormhole 
+// packet are always sent out back-to-back. After sending out current packet, channel tunnel 
+// will decide whether to switch to another source channel.
```
taylor-bsg: What if the channel is full before we are able to finish sending the whole packet? Do you compute ahead of time if the entire packet will fit in the destination buffer, or do you start interleaving the splits of the two packets?

taylor-bsg: how is fairness addressed? Is it fair in terms of the # of packets transmitted on each side, or by number of flits on each side? Do longer packets inherently get more bandwidth under contention? (Ok if they do, but useful to mention this in a comment.

bsg_dataflow/bsg_channel_tunnel_wormhole.v
```diff
@@ -0,0 +1,595 @@
...
+  // outgoing multiplexed data
+  ,output [width_p-1:0] multi_data_o
+  ,output               multi_v_o
+  ,input                multi_yumi_i
+
+  // demultiplexed data
+  ,input  [num_in_p-1:0][bsg_ready_and_link_sif_width_lp-1:0] link_i
+  ,output [num_in_p-1:0][bsg_ready_and_link_sif_width_lp-1:0] link_o
+  );
+
```
taylor-bsg: No need for more than one newline of white space per gap

bsg_dataflow/bsg_channel_tunnel_wormhole.v
gaozihou: Thanks prof, In my opinion we need to pass it in.

Intuitively, this param is equal to (1<<len_width_p). The concern is that we cannot assume all kinds of packets in wormhole network will go through channel tunnel to IO channels. len_width_p is a **property of network**, the max possible payload flits is **settings of channel tunnel**. Using (1<<len_width_p) may result in unnecessarily large pseudo_large_fifo size :)

gaozihou:
I've redefined wormhole_packet_s, now x_cord, y_cord and len are all on fixed places of low bits.

taylor-bsg:
This is a very clever module -- will revisit after tapeout!

gaozihou:
Sure prof, Thanks for your suggestions!

taylor-bsg:
We could keep it open just so we don't forget?

dpetrisko:
Maybe change the title?

PR #72, opened 2019-06-07
Wormhole Packet Definition

gaozihou:
This definition is a little bit tricky:

1. wormhole_packet_width: Given a payload data packet, what is the minimum bits needed for the full wormhole packet (including both header and payload). 
2. wormhole_packet_s: Given a pre-calculated full packet width (which is multiple of wormhole channel width), define the wormhole packet struct.

### DIFF
--- a/bsg_noc/bsg_noc_links.vh
+++ b/bsg_noc/bsg_noc_links.vh
@@ -31,6 +31,9 @@
       logic [in_data_width-1:0] data;                                     \
   } in_struct_name
 
+`define bsg_ready_and_link_sif_s_cast(link_i, link_o, link_i_cast, link_o_cast) \
+  assign link_i_cast = link_i;                                                  \
+  assign link_o = link_o_cast // no semicolon
 
 /******************* bsg wormhole packet definition ******************/
 
--- a/bsg_noc/bsg_noc_links.vh
+++ b/bsg_noc/bsg_noc_links.vh
@@ -31,4 +31,18 @@
       logic [in_data_width-1:0] data;                                     \
   } in_struct_name
 
+
+// bsg_noc_wormhole
+`define bsg_wormhole_packet_width(reserved_width_p, x_cord_width_p, y_cord_width_p, len_width_p, data_width_p) \
+  (reserved_width_p+x_cord_width_p+y_cord_width_p+len_width_p+data_width_p)
+  
+`define declare_bsg_wormhole_packet_s(width_p, reserved_width_p, x_cord_width_p, y_cord_width_p, len_width_p, in_struct_name) \
+  typedef struct packed {                                               \
+    logic [reserved_width_p-1:0] reserved;                              \
+    logic [x_cord_width_p-1:0] x_cord;                                  \
+    logic [y_cord_width_p-1:0] y_cord;                                  \
+    logic [len_width_p-1:0] len;                                        \
+    logic [width_p-reserved_width_p-x_cord_width_p-y_cord_width_p-len_width_p-1:0] data; \
+  } in_struct_name
+
  `endif // BSG_NOC_LINKS_VH
--- a/bsg_noc/bsg_noc_links.vh
+++ b/bsg_noc/bsg_noc_links.vh
@@ -33,16 +33,25 @@
 
 
 // bsg_noc_wormhole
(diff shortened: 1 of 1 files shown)


bsg_noc/bsg_noc_links.vh
```diff
@@ -31,4 +31,18 @@
...
   } in_struct_name
 
+
+// bsg_noc_wormhole
+`define bsg_wormhole_packet_width(reserved_width_p, x_cord_width_p, y_cord_width_p, len_width_p, data_width_p) \
+  (reserved_width_p+x_cord_width_p+y_cord_width_p+len_width_p+data_width_p)
+  
+`define declare_bsg_wormhole_packet_s(width_p, reserved_width_p, x_cord_width_p, y_cord_width_p, len_width_p, in_struct_name) \
+  typedef struct packed {                                               \
+    logic [reserved_width_p-1:0] reserved;                              \
```
taylor-bsg: recommend you put x_cord, y_cord, and len at known locations in the packet -- ideally in the low bits, since they don't change with data payload size, but if not there, then in the highest bits. to match, I suggest putting X in the lowest bits, then Y, then len.

gaozihou: Thanks prof! Putting on low bits is what Tommy is doing on coherence network (our wormhole router supports both header_on_low_bits and header_on_high_bits mode). The tricky part is that bsg_channel_tunnel appends "tag" on high bits, if we do so, we'll need to "reassemble" the reserved bits (typically only 2 bits, though) inside channel_tunnel_wormhole.

It will be even harder if we put reserved on high bits and x_cord, y_cord and len on low bits, since we cannot easily assemble a "header+payload" array, then go through PISO. I would rather go with the first option.

gaozihou: Another idea is to slightly modify bsg_channel_tunnel, add a parameter "tag on low bits", then all problems are solved!

bsg_noc/bsg_noc_links.vh
```diff
@@ -31,4 +31,27 @@
       logic [in_data_width-1:0] data;                                     \
   } in_struct_name
 
+
+// bsg_noc_wormhole
+`define bsg_wormhole_packet_width(reserved_width, x_cord_width, y_cord_width, len_width, data_width) \
+  (reserved_width+x_cord_width+y_cord_width+len_width+data_width)
+ 
+`define declare_bsg_wormhole_packet_s(width, reserved_width, x_cord_width, y_cord_width, len_width, in_struct_name) \
```
taylor-bsg: maybe a comment as to why they need to be different (or do they?)

dpetrisko: From a user perspective, I would rather pass in my payload width than calculate the packet width and pass it in. Is there a reason to do this the other way?

dpetrisko: (This has the extra benefit of having the width and struct macro signatures match)

gaozihou: Thanks prof and Dan! Honestly they do not have to be different, I can change the data structure to pass in the payload width.

Indeed, this data structure is not intended for users... It is used by wormhole router and channel tunnel,. So it might make sense to change the current data structure name to "wormhole_header_flit_s" (which has fixed width), add a new "wormhole_packet_s" for end users.

dpetrisko: Sounds good! Would it make sense to use a packed union, where the flit could be either the header flit or a payload flit?

gaozihou: I guess we do not need a data structure for payload flit... :)

dpetrisko: Yeah, but couldn’t an adapter use the payload flit struct?

gaozihou: I'm afraid not, the "header flit" also contains part of payload, it is not "header only" :) Typically we only have ~10 "header bits" in a header flit, all the rest is part of payload.

dpetrisko: Do the payload flits have any information except data?

gaozihou: payload flits are data-only

dpetrisko: Ah, I see. I thought there was some metadata associated with the payload as well. Makes sense then!

PR #75, opened 2019-06-10
normal number should not include zero

tommydcjung:


tommydcjung:
that's right. the result of fclass.s should be one-hot.

PR #85, opened 2019-06-13
[bsg_tag_trace_replay] adding an output enable array

stdavids:
Used to enable or disable tag masters for multi-master systems.


stdavids:
I don’t know how concerned we need to be about backwards compatibility
because I don’t think anything uses this outside of EE477. I do like the
idea of the generate if though, but what do you want to do with the en_r_o
port if num_masters_p is 0?

taylor-bsg:
In that case I guess we wire it to ‘z?

stdavids:
Updated to reflect feedback.

PR #86, opened 2019-06-13
Fixing bsg_clkgate_optional to use latest bsg_dlatch

dpetrisko:


dpetrisko:
This fixes the Verilator build. VCS apparently doesn't mind if you use incorrect port names in modules which aren't used.  Verilator prebuilds all used files, so it'll error out even if bsg_dlatch is not explicitly instantiated

PR #89, opened 2019-06-16
Dynamic Ratio PISO

gaozihou:
Add these modules for BP chip.


bsg_dataflow/bsg_parallel_in_serial_out_dynamic.v
```diff
@@ -0,0 +1,112 @@
...
+  
+  // Input side
+  ,input                               v_i
+  ,input  [lg_max_els_lp-1:0]          len_i
+  ,input  [max_els_p-1:0][width_p-1:0] data_i
+  ,output                              ready_o
+  
+  // Output side
+  ,output                              v_o
+  ,output                              len_v_o
```
tommydcjung: how is len_v_o used?

gaozihou: This is an interesting question! I decided to add this port just in case user will need it (have a symmetric port **len_ready_o** on SIPO), but indeed I am not using it in the BP adapters...

Technically, each PISO / SIPO has an internal length counter. len_v_o (and len_ready_o) means counter_value==0, meaning that it is in the init state, not in the middle of conversion process. The user may use it in the way of:

1. Drive the user control logic (get counter==0 signal for free).
2. Compare with user's local state, display error message when they mismatch.

I am not using it in BP adapters, because PISO / SIPO is directly to noc_link port.

bsg_dataflow/bsg_parallel_in_serial_out_dynamic.v
```diff
@@ -0,0 +1,112 @@
+/**
```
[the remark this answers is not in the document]
taylor-bsg: Nice draft! I suggest: 

1) separating this into two pull requests.
2) for bsg_parallel_in_serial_out_dynamic, you are paying a high cost to provide a helpful interface on both sides (see the BaseJump STL paper if you need a refresher on this), basically adding a twofer. In these cases, it is better to push the FIFO outside of the module since this will result in FIFOs that you may not need. This module would be a lot more natural if you make it valid->yumi on the input and ready/valid on the output. You should be able to eliminate the FIFOs, with the rule that len_i must be asserted for as long as data_i is asserted.
3) for bsg_parallel_in_serial_out_dynamic, it looks like that value you are actually putting into len_i is not the length, but the last_index_i.  understandably, the math works out more cleaning since you are trying to march through payload+1 elements. I think to handle this, maybe you can have a parameter last_index_not_length and have a generate statement that generates the code to do the comparison of the next value of the counter versus the current value of the counter. You will need to adjust max_els accordingly.

dpetrisko:
Can we bump priority on this? BlackParrot is blocked on moveingto master by this

gaozihou:
@dpetrisko I've updated PISOD. Going to put SIPOD in another pull request.

bsg_dataflow/bsg_parallel_in_serial_out_dynamic.v
```diff
@@ -0,0 +1,86 @@
...
+  ,output [width_p-1:0]                data_o
+  ,input                               yumi_i // late
+  );
+
+  logic [lg_max_els_lp-1:0] count_r, count_lo;
+  logic clear_li, up_li;
+  logic count_r_is_zero, count_r_is_last;
+  
+  // fix evaluate to Z problem in simulation
+  assign count_lo = count_r;
```
dpetrisko: What was this problem?

bsg_dataflow/bsg_parallel_in_serial_out_dynamic.v
```diff
@@ -0,0 +1,86 @@
...
+  assign count_r_is_zero = (count_lo == lg_max_els_lp'(0));
```
dpetrisko: This doesn't seem to be used?

bsg_dataflow/bsg_parallel_in_serial_out_dynamic.v
gaozihou: The problem was that with debug_pp in VCS, count_r (output of counter_clear_up) evaluates to Z. Need count_lo to make it evaluate to actual value.

gaozihou: I also see this effect in Manycore tile, the address evaluates to Z but the wire connected to it seems correct. I guess this is related to debug_pp option?

dpetrisko: Oh yes, I've seen that before too.  If I recall, it doesn't cause functionality problems, it just looks weird? The question is whether there is a better fix to get around this tool bug

gaozihou: Using debug instead of debug_pp does help, but it runs much much slower...

taylor-bsg:
Obsolete? @gaozihou

gaozihou:
Yeah seems that this is also obsolete

PR #92, opened 2019-06-18
Pseudo large fifo: only verbose when verbose_p==1

gaozihou:
Many group members asked me why channel_tunnel prints out "fifo is overflowing" during simulation. Should not print this out when verbose_p==0!


taylor-bsg:
A better non-synthesized check is whether this fifo is receiving data faster than the max expected run length. This gets to the heart of what corner this module is cutting and detecting whether this cut corner is reasonable in the context of the design it is being used in.

PR #94, opened 2019-06-19
Hotfix BSG Unconcentrate Static for Synthesis

stdavids:
'z will get synthesized into tri-state (not disconnect as it is intended here). Therefore, I surrounded the 'z assigns in translate off/on pragmas so that DC simply leaves them disconnected.

### DIFF
--- a/bsg_misc/bsg_defines.v
+++ b/bsg_misc/bsg_defines.v
@@ -19,9 +19,9 @@
 `endif
 
 `ifdef SYNTHESIS
-`define BSG_DISCONNECT 1'b0
+`define BSG_DISCONNECTED 1'b0
 `else
-`define BSG_DISCONNECT 1'bz
+`define BSG_DISCONNECTED 1'bz
 `endif
 
 // using C-style shifts instead of a[i] allows the parameter of BSG_GET_BIT to be a parameter subrange                                                                                                                                                                               
--- a/bsg_misc/bsg_defines.v
+++ b/bsg_misc/bsg_defines.v
@@ -19,9 +19,9 @@
 `endif
 
 `ifdef SYNTHESIS
-`define BSG_DISCONNECTED 1'b0
+`define BSG_DISCONNECTED_IN_SIM 1'b0
 `else
-`define BSG_DISCONNECTED 1'bz
+`define BSG_DISCONNECTED_IN_SIM 1'bz
 `endif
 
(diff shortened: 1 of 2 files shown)


stdavids:
Update, also added the SYNTHESIS definition to analysis in our cad flow.

dpetrisko:
What do you mean by added SYNTHESIS definition? My understanding is that that is implicitly set by Synopsys

stdavids:
Is that true? When you read RTL via the "analyze" command, there is a -define {list} flag where you can set definitions. I added SYNTHESIS to that list. I wasn't aware that DC does this by default.

dpetrisko:
Depends on the tool, but seems like most do: https://groups.google.com/forum/#!topic/chisel-users/db1diDbg7T4

dpetrisko:
Here's a better reference where it recommends using SYNTHESIS over // translate_off

https://www.veripool.org/papers/Preproc_Good_Evil_SNUGBos10_paper.pdf

taylor-bsg:
Let's have it match the UNDEFINED macro for symmetry:

`ifdef SYNTHESIS
`define BSG_DISCONNECTED_IN_SIM(val) (val)
`else
`define BSG_DISCONNECTED_IN_SIM(val) ('z)
`endif

Yeah my reason for using ifdef SYNTHESIS comes from that SNUG article bouncing around in my brain for a fewyears.

taylor-bsg:
Let's make it a macro with parameters like in my comment above

stdavids:
@Dan Yeah that article... I've read it a few times, can't say I was super compelled by it. Macros are fine, the translate tags have just always been there (from my perspective) and I'm not a huge fan of mixing and matching. The // pragmas are used for more than just translate so we can't really get rid of all of them.

I abhor the idea of relying on DC to do anything automatically, especially something so detail oriented as auto defining a macro so I think I'll keep the redefinition in the flow haha.

stdavids:
@taylor-bsg updated.

dpetrisko:
> I've read it a few times, can't say I was super compelled by it.

Unfortunately, Wilson Synder was convinced by it, as Verilator only supports the ifndef `SYNTHESIS, but not // translate_off.  

> The // pragmas are used for more than just translate so we can't really get rid of all of them.

Yeah, definitely. The ones that actually do synthesis-y things, we need.

> I abhor the idea of relying on DC to do anything automatically, especially something so detail oriented as auto defining a macro so I think I'll keep the redefinition in the flow haha.

Totally fair! I can't imagine a scenario where a redundant define will cause issues

PR #95, opened 2019-06-19
Add wormhole network loopback test. Fix wormhole parameter default value.

gaozihou:


taylor-bsg:
Accepting without review

PR #98, opened 2019-06-20
Changing if-errors to assert-else-errors in mesh_router

dpetrisko:
Use asserts instead of ifs.  Asserts can be disabled by tools, but if () finishes cannot.  Standard use for Verilator is to disable asserts until reset sequence is over, then re-enable them.  Otherwise, the asserts will fire and terminate during reset sequence (which can extend before or after when reset_i itself is low)

### DIFF
--- a/bsg_noc/bsg_mesh_router.v
+++ b/bsg_noc/bsg_mesh_router.v
@@ -89,18 +89,18 @@ if( XY_order_p == 1 ) begin:XY_dor
     //synopsys translate_off
     always@(negedge clk_i) begin
        //Y dim to X dim
-       assert( ~(v_i[N]& ~x_eq[N]) | (v_i[S] & ~x_eq[S]) ) else begin
+       assert( ~((v_i[N]& ~x_eq[N]) | (v_i[S] & ~x_eq[S])) ) else begin
                $error("%m:Y dim to X dim routing. XY_order_p = %b", XY_order_p);
                $finish();
        end  
        // the same X dim routings. For X/Y routing, we don't care Y dimesions.
-       assert( ~(v_i[W] & x_lt[W]) | (v_i[E] & x_gt [E] ) ) else begin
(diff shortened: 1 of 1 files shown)


proftaylor:
Pretty sure the operator precedence is wrong on this change.

Use of assert is okay.

Not sure what you mean about ventilator and reset though — how does it know
what reset is? This code has been in here forever so I am not sure that it
is “the problem.”

Usually you throw in a (reset !=== 0) || into your assert to guard against
asserts during reset.


M

dpetrisko:
Usecase: reset is set by bsg_tag. It takes several cycles to get set to 1, before then it’s indeterminate. The assert fires during this time, even though the system is not yet valid.

taylor-bsg:
I see — the reset check should fix this more properly, but the bug in the
inversion of the assertion condition also needs to be fixed.

dpetrisko:
Fixed. Here’s an example of the Verilator and VCS sequences to avoid this

Verilator:
```
  Verilated::assertOn(false);
  reset = 1;
  for (int i = 0; i < 20; i++) {
    sc_start(CLK_TIME, SC_NS);
  }
  reset = 0;
  Verilated::assertOn(true);

  while (!Verilated::gotFinish()) {
    sc_start(CLK_TIME, SC_NS);
  }
```

VCS:
```
initial 
  begin
    $assertoff();
    @(negedge reset)
    $asserton();
  end
```

taylor-bsg:
Very cool! How do you control this for multiple clock domains though?

dpetrisko:
Hm, interesting question.  AFAIK, there’s only a global on/off switch, so in this solution you would need to trigger on the _last_ reset going low.

Spitballing:

```
`define bsg_assert_after_reset(reset_mp, condition_mp) \
  @nededge(reset_mp); \
  assert (condition_mp);
```

Maybe wrapped in a do-while?

PR #99, opened 2019-06-21
bsg_link downstream harden memory

gaozihou:


taylor-bsg:
Thanks for the pull request, Paul. This fifo code is less tested than the other FIFO code.  What is the geometry of the FIFO that you need?

gaozihou:
Thanks prof. May I ask what "the other fifo" refers to?
The geometry is 17-width * 64-depth.

taylor-bsg:
The other FIFO is that is very well tested is the 1r1w_small. The pseudo fifo is reasonable well tested as well. But 1r1w_large I don't think is silicon proven. Part of the issue is that there are very large overheads (>= 3 flip flops and 2 twofers and control logic) associated with the 1r1w_large so it only barely breaks even at 64els x 32bits.

Unless we have another reason, I suggest that we use 1r1w_small; maybe rename the signals to be "sync" instead of large. 

We can file an issue to create a FIFO with a 1r1w hardened RAM to cover these cases in the future...

M

gaozihou:
Thanks for the kind suggestions prof!
I am wondering that fifo_1r1w_small and async_fifo both use same type memory (bsg_mem_1r1w):
https://github.com/bespoke-silicon-group/basejump_stl/blob/e03010b0253cd42a0ac2a24f696cae1a64cbb9ae/bsg_dataflow/bsg_fifo_1r1w_small.v#L70
https://github.com/bespoke-silicon-group/basejump_stl/blob/e03010b0253cd42a0ac2a24f696cae1a64cbb9ae/bsg_async/bsg_async_fifo.v#L48
I think using 1r1w_small_fifo is identical to using memory inside async_fifo directly...
Or, maybe we set async_fifo to be 17x32, 1r1w_small_fifo also be 17x32? Will this help on the back-end (easier to synthesize / place and route)?

taylor-bsg:
This change is pending adding the code for the two modules in this issue:

https://github.com/bespoke-silicon-group/basejump_stl/issues/100

Then we can decide whether to use an 8-element ASYNC fifo plus a 56 element synchronous fifo, or a single 64 element ASYNC FIFO.

PR #101, opened 2019-06-21
request to add a hidden folder .regression into basejump_stl

tanglingshu:
adding .regression into basejump_stl for ppa regression purpose


tanglingshu:
outdated, will make a new PR soon

taylor-bsg:
Seems like you deleted a lot of files in this PR that should not be deleted...?

tanglingshu:
> Seems like you deleted a lot of files in this PR that should not be deleted...?

That is true, mainly because I was testing iterative multiplier and pulled in extra files from our international contributor. I just wanted to get the .regression folder in basejump_stl in there without touching all the other directories. I probably shouldn't have deleted those files but I didn't want you to pull in those extra files either. 

I'll make a PR soon without all these removals and a better structured .regression directory. The current structure can be found at https://github.com/tanglingshu/basejump_stl/tree/master/.regression
will add more valid configs

PR #110, opened 2019-07-23
bsg_encode_one_hot; adding delay to resolve race condition to check o…

tommydcjung:
…ne-hot assertion

### DIFF
--- a/bsg_misc/bsg_encode_one_hot.v
+++ b/bsg_misc/bsg_encode_one_hot.v
@@ -80,7 +80,7 @@ module bsg_encode_one_hot #(parameter width_p=8, parameter lo_to_hi_p=1)
   `else
     assign addr_o = (((i-1) & i) == '0)
       ? addr_lo
-      : {width_p{1'bx}};
+      : {`BSG_SAFE_CLOG2(width_p){1'bx}};
   `endif
 
 
--- a/bsg_misc/bsg_encode_one_hot.v
+++ b/bsg_misc/bsg_encode_one_hot.v
@@ -75,7 +75,7 @@ module bsg_encode_one_hot #(parameter width_p=8, parameter lo_to_hi_p=1)
 
 
    // synopsys translate_off
-   always
+   always @ (i)
      begin
 //        $display("bsg_encode_one_hot %b -> add_o(%b), v_o(%b)",i,addr_o, v_o);
         #1;
--- a/bsg_misc/bsg_encode_one_hot.v
+++ b/bsg_misc/bsg_encode_one_hot.v
@@ -75,11 +75,12 @@ module bsg_encode_one_hot #(parameter width_p=8, parameter lo_to_hi_p=1)
 
 
    // synopsys translate_off
-   always @(addr_o)
+   logic [width_p-1:0] i2;
+   always
      begin
 //        $display("bsg_encode_one_hot %b -> add_o(%b), v_o(%b)",i,addr_o, v_o);
-
-        assert ($countones(i) <= 1) else $error("bsg_encode_one_hot: invalid input %b",i);
+        i2 = #1 i;
+        assert ($countones(i) <= 1) else $error("bsg_encode_one_hot: invalid input %b",i2);
      end
    // synopsys translate_on
 
--- a/bsg_misc/bsg_encode_one_hot.v
+++ b/bsg_misc/bsg_encode_one_hot.v
@@ -75,12 +75,11 @@ module bsg_encode_one_hot #(parameter width_p=8, parameter lo_to_hi_p=1)
 
 
    // synopsys translate_off
-   logic [width_p-1:0] i2;
    always
      begin
 //        $display("bsg_encode_one_hot %b -> add_o(%b), v_o(%b)",i,addr_o, v_o);
-        i2 = #1 i;
-        assert ($countones(i) <= 1) else $error("bsg_encode_one_hot: invalid input %b",i2);
+        #1;
+        assert ($countones(i) <= 1) else $error("bsg_encode_one_hot: invalid input %b",i);
      end
    // synopsys translate_on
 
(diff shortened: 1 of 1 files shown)


dpetrisko:
This will break Verilator.  A few alternatives

1) use $asserton and $assertoff to disable assertions before reset is asserted.
2) use if (i !== 'X) to gate the assertion when it's not relevant.

tommydcjung:
add `ifndef verilator

taylor-bsg:
Back to Dan's comment, is the issue that the input value has X's, or that the input value is toggling and some of the bits are simultaneously set because they are in an intermediate state? For the former, 'X would work, for the latter, Tommy's solution is necessary and would need to be composed with the `ifndef VERILATOR

M

dpetrisko:
The problem with ifndef Verilator is that this is a helpful assertion for Verilator as well.  We shouldn't get rid of it

dpetrisko:
> Back to Dan's comment, is the issue that the input value has X's, or that the input value is toggling and some of the bits are simultaneously set because they are in an intermediate state?

Whether or not either of these are problems is a property of the encompassing design, not bsg_encode_one_hot, which is why using $asserton/$assertoff is a better solution

taylor-bsg:
But asserton and assertoff require non-local code; I think it would be better to address Dan's concern about localization by having an input parameter  disable_onehotness_check_p=0. 

But neither of these address the issue that intermediate toggling may trigger the warning. Normally we check the property on the negedge of the clock, but there is not clock here.  I don't really see how we can keep this check for Verilator. Tommy's #1 is essentially saying, check the value of this signal after the combinational stuff has settled.

dpetrisko:
> But neither of these address the issue that intermediate toggling may trigger the warning

Yeah, if you're concerned about intermediate toggling, I agree disable_onehotness_check_p=0 is the way to go.

>I don't really see how we can keep this check for Verilator.

If we're doing the ifndef Verilator, just put it around the nonsynth delay statement.  Then the user has the option to disable the check through the parameter, or do the proper asserton assertoff sequence.

dpetrisko:
delay statements are just fragile in general.  How long should it be high for?  What if we reset in the middle of the test to test power gating?  What if the timescale is different between different instantiations of the module? 

Only the user of the module knows this info and can orchestrate through non-local code.

tommydcjung:
why does it break the verilator? because of '#1'? I think the real problem is Verilator not supporting the standard. maybe Dan can relay the urgency of this feature to the verilator community.

dpetrisko:
Verilator only supports synthesizable SystemVerilog + system tasks, such as assertions and print statements.

drichmond:
Is there a reason change this would be VASTLY slowing down simulation?

dpetrisko:
> Is there a reason change this would be VASTLY slowing down simulation?

Yes, I think so.  Having #1 in an always block evaluates every sim tick.  Usually #1 are only used in initial blocks for this reason.

taylor-bsg:
Tommy, we do have an interest in supporting Verilator since VCS is not open
source, and is extremely expensive. But certainly there is a tension there;
we want our Verilog to take advantage of modern features.


M

taylor-bsg:
Oh are we seeing extreme slowdowns?

dpetrisko:
Another reason to not use delay.  While simulating actual chips, there is often a delay between simulation starting and reset being asserted (pipelined reset).  This delay implicitly assumes that reset is initially high and will trigger if the reset pipeline is longer than 1 timeunit.

taylor-bsg:
On Tue, Jul 23, 2019 at 4:32 PM Dan Petrisko <notifications@github.com>
wrote:

> But neither of these address the issue that intermediate toggling may
> trigger the warning
>
> Yeah, if you're concerned about intermediate toggling, I agree
> disable_onehotness_check_p=0 is the way to go.
>
> I don't really see how we can keep this check for Verilator.
>
> If we're doing the ifndef Verilator, just put it around the nonsynth delay
> statement. Then the user has the option to disable the check through the
> parameter, or do the proper asserton assertoff sequence.
>

Well we want don’t want the user to have to change the source to run
Verilator.

taylor-bsg:
Driving this to its logical conclusion...

a. Agreed that we don’t generally like # as a construct. This is perhaps
one of two cases that I have seen that make some sense. (The other being
generating a clock. :-] )
b. Agreed that have an “error” is a bit strong since there are all kinds of
cases where we may not want to terminate the simulation if the invariant is
not true (reset, combinational signaling, etc). We could make it a warning,
but that is also not great.

My final takeaway is that we are trying to apply a sequential logic mindset
to test combinational logic, which doesn’t make sense. For combinational
logic, the mechanism we have is X propagation. This will sort out the times
when we do and do not care. So in this case if (i & (i-1)) then we would
use BSG_UNDEFINED_IN_SIM to override the output value.

tommydcjung:
The slowdown comes from not including the sensitivity list so the always block is triggered every time tick. I added back the sensitivity list, and I think it reduces the simulation time. 

The current implementation works fine for vcs-2016, but it does not work for later versions. Maybe the scheduling algorithm changed so that the assertion is triggered when the combinational logic has not fully settled yet. Maybe there is some new flag in vcs-2019 to go back to the old logic. maybe I could get access to the vcs-2019 user guide.

For verilators, I'm not sure what the clean solution is besides disabling the assertion. Since this module is purely combinational and there is no clock signal to use the negedge to trigger the assertion.

taylor-bsg:
Hi Tommy,

See my previous message. The solution is to remove the #1 and use X
propagation to signal the problem.

tommydcjung:
so are you suggesting something like

```
assign addr_o = ($countones(i) <= 1)
  ? addr_lo
  : {width_p{1'bx}};
```

taylor-bsg:
Use the expression I provided earlier instead of countones, and wrap it in
a ifndef SYNTHESIS

tommydcjung:
okay I think we can merge this in.

PR #112, opened 2019-07-24
Minor fixes for verilator and adding reset_val to dff

dpetrisko:


dpetrisko:
These changes got incorporated in the hb_fpu PR

PR #114, opened 2019-07-25
nonsynth mem using associative array

tommydcjung:


taylor-bsg:
Looks great, put a comment in the header that this is for simulating
arbitrarily large memories.

dpetrisko:
Can we put an ifndef VERILATOR elaboration assertion that errors if you try to instantiate? (Similar to bsg_dlatch). Verilator doesn’t support associative arrays

tommydcjung:
it will fail to compile regardless, so I think it's fine without it?

dpetrisko:
Yeah, but it’s a more informative error. Makes it clear to the user that we know it’s unsupported

dpetrisko:
Not a showstopper though, just a thought

PR #119, opened 2019-07-26
tagfl regression test; fixed the bug with dma counter not initializin…

tommydcjung:
…g properly; tagfl_op reset


tommydcjung:
not sure what happen to this branch...

PR #122, opened 2019-07-31
Changing wormhole adapters to use SIPO/PISO

dpetrisko:
and new wormhole header format

### DIFF
--- a/bsg_noc/bsg_wormhole_router.vh
+++ b/bsg_noc/bsg_wormhole_router.vh
@@ -4,19 +4,19 @@
 //         struct, then it will match bsg_wormhole_router
 //
 
-`define declare_bsg_wormhole_router_header_s(in_cord_width,in_len_width,in_struct_name) \
+`define declare_bsg_wormhole_router_header_s(cord_width_mp,len_width_mp,struct_name_mp) \
 typedef struct packed {                 \
-  logic [in_len_width-1:0]    len;      \
-  logic [in_cord_width-1:0 ]  cord;     \
-} in_struct_name
+  logic [len_width_mp-1:0]    len;      \
+  logic [cord_width_mp-1:0 ]  cord;     \
+} struct_name_mp
 
-`define declare_bsg_wormhole_router_packet_s(in_cord_width,in_len_width,in_payload_width,in_struct_name) \
+`define declare_bsg_wormhole_router_packet_s(cord_width_mp,len_width_mp,payload_width_mp,struct_name_mp) \
 typedef struct packed {                 \
-  logic [in_payload_width-1:0] payload; \
-  logic [in_len_width-1:0]     len;     \
-  logic [in_cord_width-1:0]    cord;    \
-} in_struct_name
+  logic [payload_width_mp-1:0] payload; \
+  logic [len_width_mp-1:0]     len;     \
+  logic [cord_width_mp-1:0]    cord;    \
+} struct_name_mp
 
-`define bsg_wormhole_router_packet_width(in_cord_width,in_len_width,in_payload_width) \
-  (in_cord_width + in_len_width + in_payload_width)
+`define bsg_wormhole_router_packet_width(cord_width_mp,len_width_mp,payload_width_mp) \
+  (cord_width_mp + len_width_mp + payload_width_mp)
 
--- a/bsg_noc/bsg_wormhole_router.vh
+++ b/bsg_noc/bsg_wormhole_router.vh
@@ -9,3 +9,14 @@ typedef struct packed {                 \
   logic [in_len_width-1:0]    len;      \
   logic [in_cord_width-1:0 ]  cord;     \
 } in_struct_name
+
+`define declare_bsg_wormhole_router_packet_s(in_cord_width,in_len_width,in_payload_width,in_struct_name) \
+typedef struct packed {                 \
+  logic [in_payload_width-1:0] payload; \
+  logic [in_len_width-1:0]     len;     \
+  logic [in_cord_width-1:0]    cord;    \
+} in_struct_name
+
+`define bsg_wormhole_router_packet_width(in_cord_width,in_len_width,in_payload_width) \
+  (in_cord_width + in_len_width + in_payload_width)
+
--- a/bsg_noc/bsg_wormhole_router_adapter_in.v
+++ b/bsg_noc/bsg_wormhole_router_adapter_in.v
@@ -1,134 +1,68 @@
 /**
  *  bsg_wormhole_router_adapter_in.v
  *
- *  packet = {payload, length, y_cord, x_cord}
+ *  packet = {payload, length, cord}
  */
 
 `include "bsg_noc_links.vh"
+`include "bsg_wormhole_router.vh"
 
 module bsg_wormhole_router_adapter_in
-  #(parameter max_num_flit_p="inv"
-    , parameter max_payload_width_p="inv"
-    , parameter x_cord_width_p="inv"      
-    , parameter y_cord_width_p="inv"
+  #(parameter max_payload_width_p = "inv"
+    , parameter len_width_p       = "inv"
+    , parameter cord_width_p      = "inv"
+    , parameter link_width_p      = "inv"
 
-    , localparam len_width_lp=`BSG_SAFE_CLOG2(max_num_flit_p)
-    , localparam max_packet_width_lp=(x_cord_width_p+y_cord_width_p+len_width_lp+max_payload_width_p)
-    , localparam width_lp=
-      (max_packet_width_lp/max_num_flit_p) + ((max_packet_width_lp%max_num_flit_p) == 0 ? 0 : 1)
- 
-    , localparam padding_width_lp= 
-      ((max_packet_width_lp%width_lp) == 0 ? 0 : (width_lp-(max_packet_width_lp%width_lp)))
-    , localparam padded_packet_width_lp=
-      max_packet_width_lp+padding_width_lp
-    , localparam len_offset_lp=(x_cord_width_p+y_cord_width_p)
+    , localparam bsg_ready_and_link_sif_width_lp =
+        `bsg_ready_and_link_sif_width(link_width_p)
+    , localparam bsg_wormhole_packet_width_lp = 
+        `bsg_wormhole_router_packet_width(cord_width_p, len_width_p, max_payload_width_p)
+    )
+   (input                                          clk_i
+    , input                                        reset_i
 
-    , localparam bsg_ready_and_link_sif_width_lp=`bsg_ready_and_link_sif_width(width_lp)
-  )
-  (
-    input clk_i
-    , input reset_i
-
-    , input [max_packet_width_lp-1:0] data_i
-    , input v_i
-    , output logic ready_o
+    , input [bsg_wormhole_packet_width_lp-1:0]     packet_i
+    , input                                        v_i
+    , output                                       ready_o
 
     , output [bsg_ready_and_link_sif_width_lp-1:0] link_o
     // Used for ready_i signal, the rest should be stubbed, since this an input adapter
-    , input [bsg_ready_and_link_sif_width_lp-1:0] link_i 
+    , input [bsg_ready_and_link_sif_width_lp-1:0]  link_i 
   );
 
   // Casting ports
-  `declare_bsg_ready_and_link_sif_s(width_lp,bsg_ready_and_link_sif_s);
+  `declare_bsg_ready_and_link_sif_s(link_width_p, bsg_ready_and_link_sif_s);
   bsg_ready_and_link_sif_s link_cast_i, link_cast_o;
 
-  assign link_cast_i = link_i;
-  assign link_o = link_cast_o;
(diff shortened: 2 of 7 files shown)


taylor-bsg:
&& what regressions were run / should be run

dpetrisko:
> Maybe some background on the change? 
This is a generalized version of the wormhole_adapter_in/_out.  It reuses the SIPO/PISO instead of duplicating code, and uses a struct instead of relying on bit positions for len/cord.

>Was this version taped out? 
No, the old version was taped out

>Is this used only in BlackParrot or also in the manycore? 
I grepped manycore and it seems like it's BlackParrot only.  In fact, it's only used in the coherence network -- the I/O network uses the SIPO/PISO in a less general fashion than this

bsg_noc/bsg_wormhole_router.vh
```diff
@@ -9,3 +9,14 @@ typedef struct packed {                 \
   logic [in_len_width-1:0]    len;      \
   logic [in_cord_width-1:0 ]  cord;     \
 } in_struct_name
+
+`define declare_bsg_wormhole_router_packet_s(in_cord_width,in_len_width,in_payload_width,in_struct_name) \
```
tommydcjung: macro params are supposed to have "_mp" as suffix.

dpetrisko: yeah, you're right.  I was trying to be consistent with the rest of file.  I'll change both

dpetrisko:
> && what regressions were run / should be run

I'm using them in a branch of BlackParrot that I'm working on.  I've found a few small issues -- we should hold off merging until it works there

dpetrisko:
Just added the dynamic versions last night :)

taylor-bsg:
Code looks nice

dpetrisko:
For the delay, I think negedge(reset) is better, since it will only trigger once?

dpetrisko:
Comments addressed, except for dff_reset reset_val_p, which should be its own issue.  Can we merge this?

dpetrisko:
Actually, I'm trying to fix up the CCE-MEM adapters in BP and those require the src_cord to be sent along with the packet.  Does it make sense to have declare_bsg_wormhole_router_packet_s have a src_cord and dst_cord instead of just cord (meaning dst_cord)? @taylor-bsg

dpetrisko:
Another option is to do something like:
```
`define declare_bsg_wormhole_router_header_s(cord_width_mp,len_width_mp,struct_name_mp) \
typedef struct packed {                 \
  logic [len_width_mp-1:0]    len;      \
  logic [cord_width_mp-1:0 ]  cord;     \
} struct_name_mp

`define declare_bsg_wormhole_router_header_with_src_s(cord_width_mp,len_width_mp,struct_name_mp) \
typedef struct packed {                 \
  logic [len_width_mp-1:0]    len;      \
  logic [cord_width_mp-1:0 ]  dst_cord;     \
  logic [cord_width_mp-1:0]  src_cord;      \
} struct_name_mp

`define declare_bsg_wormhole_router_packet_s(header_name_mp,payload_width_mp,struct_name_mp) \
typedef struct packed {                 \
  logic [payload_width_mp-1:0] payload; \
  ``header_name_mp`` header;
} struct_name_mp
```

This would force clients to declare the header before declaring the packet, though

gaozihou:
> Another option is to do something like:
> 
> ```
> `define declare_bsg_wormhole_router_header_s(cord_width_mp,len_width_mp,struct_name_mp) \
> typedef struct packed {                 \
>   logic [len_width_mp-1:0]    len;      \
>   logic [cord_width_mp-1:0 ]  cord;     \
> } struct_name_mp
> 
> `define declare_bsg_wormhole_router_header_with_src_s(cord_width_mp,len_width_mp,struct_name_mp) \
> typedef struct packed {                 \
>   logic [len_width_mp-1:0]    len;      \
>   logic [cord_width_mp-1:0 ]  dst_cord;     \
>   logic [cord_width_mp-1:0]  src_cord;      \
> } struct_name_mp
> 
> `define declare_bsg_wormhole_router_packet_s(header_name_mp,payload_width_mp,struct_name_mp) \
> typedef struct packed {                 \
>   logic [payload_width_mp-1:0] payload; \
>   ``header_name_mp`` header;
> } struct_name_mp
> ```
> 
> This would force clients to declare the header before declaring the packet, though

gaozihou:
> > Another option is to do something like:
> > ```
> > `define declare_bsg_wormhole_router_header_s(cord_width_mp,len_width_mp,struct_name_mp) \
> > typedef struct packed {                 \
> >   logic [len_width_mp-1:0]    len;      \
> >   logic [cord_width_mp-1:0 ]  cord;     \
> > } struct_name_mp
> > 
> > `define declare_bsg_wormhole_router_header_with_src_s(cord_width_mp,len_width_mp,struct_name_mp) \
> > typedef struct packed {                 \
> >   logic [len_width_mp-1:0]    len;      \
> >   logic [cord_width_mp-1:0 ]  dst_cord;     \
> >   logic [cord_width_mp-1:0]  src_cord;      \
> > } struct_name_mp
> > 
> > `define declare_bsg_wormhole_router_packet_s(header_name_mp,payload_width_mp,struct_name_mp) \
> > typedef struct packed {                 \
> >   logic [payload_width_mp-1:0] payload; \
> >   ``header_name_mp`` header;
> > } struct_name_mp
> > ```
> > 
> > 
> > This would force clients to declare the header before declaring the packet, though
dest_cord should always be on low bits of header, need to move src_cord to high bits.

taylor-bsg:
Need for a source cord is protocol dependent. Many protocols do not need
it. Prefer not to have it part of the header, should be payload.

dpetrisko:
Makes sense! I guess thinking of it like a traditional network stack helps. So the “payload” in this case will be user defined. Would it be a good idea to provide:

‘define_wormhole_payload_x_s
‘define_wormhole_payload_y_s

for common protocols? Or just leave it to clients to define their own?

dpetrisko:
Added two more options:
For adapter_out: buffered_p puts a fifo on the output of the link.  This will save adding the fifos manually to each one of our networks, when we expect input buffering

For adapter_out and adapter_in: stub_p stubs the output links to 0.  This is the default behavior, but sometimes we don't want to drive the link outputs. For instance, for a bidirectional link where we want both an adapter_in and adapter_out, we need to not stub either, so they don't multiply drive.  Since this isn't an actual tri-state, I didn't use `BSG_DISCONNECTED_IN_SIM and actually let them dangle

dpetrisko:
Added input and output buffering options to the wormhole adapters.  Also added a bidirectional adapter.  There's a slight overhead to reusing the adapter_in/adapter_out, because each adapter stubs the unused fields of the output link.  However, the other ways to get around this issue without multiple drivers are splitting up the output links into 3 ports (unideal) or using a tristate (Verilator incompatible and also could be QoR implications).

taylor-bsg:
This input/output buffering thing is indicative of misuse or
misunderstanding of ready/valid somewhere in the code. If a module has a
ready/valid input that is fed directly into another ready/valid module,
then by definition, it does not need a FIFO.  Both modules are supposed to
send their values at the beginning of the cycle, so there is already a full
cycle of wire delay that can be hidden.  Now if you are trying to bridge
additional distance that is understandable, but that repeater should go one
level up since when we do move bounds we would not want to include that
FIFO.

Does that make sense?


M

dpetrisko:
Not sure I understand (Possible it’s a vocab problem). It’s not about changing the handshake.  From what I’ve been told, the LCE-CCE networks all require input buffering to prevent deadlock. Right now how we handle this is we have a twofifo on the outputs from the network to the LCE/CCE. All this parameter is doing is moving that fifo inside the adapter instead of between the adapter and LCE/CCE. I don’t believe there is any physical difference between the two other than code partitioning

taylor-bsg:
Look inside parallel in serial out dynamic — as the ready/valid interface
suggests, you already have input buffering...

Input buffering is not really a deadlock thing; it’s more they in order to
obey valid/ready you need either buffering on input or on output. You don’t
want both. We prefer input over output for the same reason we prefer input
registering of modules: it makes the timing paths in the module invariant
of the neighbor modules.

There are limited scenarios where you actually are counting buffers but
these are only in cases where you are otherwise violating the invariant
that you are required to sink input messages and are trying to fix it by
having enough space to pull in anything you may receive. Even then, this
only really is practical where the number of such senders is limited.

If that fifo was required then there should be a big comment saying why it
is required...

M

dpetrisko:
Okay, makes sense! 

>There are limited scenarios where you actually are counting buffers but
these are only in cases where you are otherwise violating the invariant
that you are required to sink input messages and are trying to fix it by
having enough space to pull in anything you may receive.

This is what I thought we were doing, but it sounds like it was originally put in for handshaking concerns (since the routers are ready-valid&).  With the adapters being helpful, I think we can get rid of most of the buffers in the coherence networks.  I'll get rid of the parameter

taylor-bsg:
I think the buffer counting scenarios are mostly  associated with the CCE
(since it ignores incoming packets and even then not in all cases). Maybe
you can ask Mark to annotate in the comments the minimum number of packets
or flits of buffering he needs in each case.  Generally there are three
cases:

L - need to hold at least one message from each LCE

M - need to hold one message per outstanding miss to memory system

1 - only one packet is ever outstanding at a time; or this network only
receives packets at most every other cycle

2 - typical case — network can have one packet per cycle; two elements are
required for full rate ready/valid handshaking.


Annotating the buffer use will allow us to allow code maintainers to
maintain correctness but also to downsize to what we need.

M

dpetrisko:
Sounds good. We have an outstanding issue on BlackParrot to justify the buffers, I’ll add annotation to it. 

As far as this code goes, though, it seems to me that an input_buffer_p=(elements) with default 0 is useful for these cases. I suppose there’s a placement bounds argument, but for clarity you could just as easily annotate the output adapter as you could a separate FIFO. And providing output buffering could be useful from a library perspective, even if we personally don’t use it (clients often latch outputs instead of inputs). 

Your call though, I’m don’t feel passionately either way.

bsg_noc/bsg_wormhole_router_adapter_out.v
```diff
@@ -1,154 +1,102 @@
...
-      count_r <= '0;
+  `declare_bsg_wormhole_router_header_s(cord_width_p, len_width_p, bsg_wormhole_header_s);
+  bsg_wormhole_header_s header_li;
+  assign header_li = link_cast_i.data;
+
+  `declare_bsg_wormhole_router_packet_s(cord_width_p,len_width_p,max_payload_width_p,bsg_wormhole_packet_s);
+  bsg_wormhole_packet_s packet_lo;
+  logic ready_li, v_lo;
+
+  localparam max_num_flits_lp = 2**len_width_p;
```
taylor-bsg: This is going to instantiate a FIFO with width that is exponential relative to the field in the packet. Instead you should put the largest expected packet size in flits, based on the protocol, not based on the field; and add an assert to make sure that the len that comes in is not larger than that size.

bsg_noc/bsg_wormhole_router_adapter_out.v
```diff
@@ -1,154 +1,102 @@
...
+  bsg_serial_in_parallel_out_dynamic
+   #(.width_p(link_width_p)
+     ,.max_els_p(max_num_flits_lp)
+     )
+   sipo
+    (.clk_i(clk_i)
+     ,.reset_i(reset_i)
+
+     ,.data_i(link_cast_i.data)
+     ,.len_i(header_li.len)
```
taylor-bsg: Add check that header_li.len is not larger than the maximum expected flit width of a packet

bsg_noc/bsg_wormhole_router_adapter_in.v
```diff
@@ -1,134 +1,97 @@
...
-
-  always_ff @ (posedge clk_i) begin
-    if (reset_i) begin
-      state_r <= WAIT;
-      count_r <= '0;
+  `declare_bsg_wormhole_router_packet_s(cord_width_p, len_width_p, max_payload_width_p, bsg_wormhole_packet_s);
+  bsg_wormhole_packet_s packet_li;
+  logic ready_lo, v_li;
+
+  localparam max_num_flits_lp = 2**len_width_p;
```
taylor-bsg: Fix -- should be actual max flits based on the protocol being spoken on this network. Assertion should be used to check that incoming flit does not exceed this.

taylor-bsg:
@dpetrisko lets remove the extra input/output buffer flag.  you want "extra FIFOs" like these at the top level so that you can see how wire delay and/or deadlock is being managed in fewer places rather than hunting around the modules.   FYI, from a physical design point of few. if you have a module A that contains B, it is not always easy to put B in one place and the rest of A somewhere else.

taylor-bsg:
See also the other comments I made about the length bug in the dynamic SIPOs/PISOs

dpetrisko:
@proftaylor Comments addressed

dpetrisko:
If there are no further comments, can we merge this? I'd like to rebase #126 on top of it

taylor-bsg:
Merged.

PR #123, opened 2019-08-01
bsg_delay_queue

tommydcjung:


drichmond:
Why not do this with a circular buffer? Shift registers are vastly slower to simulate because there is N toggles instead of 2

tommydcjung:
Yeah I could try make one using circular buffer.

taylor-bsg:
Yes, just use bsg_fifo_1r1w and a bsg counter to handle the initialization
case.

M

PR #124, opened 2019-08-02
Refactor 20190731

tommydcjung:
- refactor mostly related to bsg_cache_dma
- fixing tagfl and adding regression test


bsg_cache/bsg_cache_dma.v
```diff
@@ -15,7 +15,9 @@ module bsg_cache_dma
 
     ,localparam lg_block_size_in_words_lp=`BSG_SAFE_CLOG2(block_size_in_words_p)
     ,localparam lg_sets_lp=`BSG_SAFE_CLOG2(sets_p)
+    ,localparam counter_width_lp=`BSG_SAFE_CLOG2(block_size_in_words_p+1)
```
taylor-bsg: If it's not used by an argument, don't put the localparam in the arguments

bsg_cache/bsg_cache_dma.v
```diff
@@ -55,19 +57,40 @@ module bsg_cache_dma
...
-  logic [lg_block_size_in_words_lp:0] counter_n; 
+
+  // dma counter
+  //
+  logic counter_en;
+  logic counter_set;
+  logic [counter_width_lp-1:0] counter_val;
+  logic [counter_width_lp-1:0] counter_r;
+
+  bsg_counter_set_en #(
```
taylor-bsg: bsg_counter_clear_up is sufficient, no need to add a new counter. 
You can add a clear_and_up_exclusive_p like in https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_misc/bsg_counter_set_down.v

bsg_cache/bsg_cache_dma.v
```diff
@@ -131,113 +151,107 @@ module bsg_cache_dma
...
+          dma_state_n = IDLE;
+        end
+        else if (dma_get_fill_data_i) begin
+          counter_set = 1'b1;
+          counter_val = '0;
+          dma_state_n = GET_FILL_DATA;
+        end
+        else if (dma_send_evict_data_i) begin
+          counter_set = 1'b1;
+          counter_val = (counter_width_lp)'(1);
```
taylor-bsg: instead of setting to one, you could increment?

bsg_cache/bsg_cache_dma.v
```diff
@@ -131,113 +151,107 @@ module bsg_cache_dma
...
-      SEND_EVICT_ADDR: begin
-        dma_state_n = dma_pkt_yumi_i
-          ? IDLE
-          : SEND_EVICT_ADDR;
-        dma_pkt_v_o = 1'b1;
-        dma_pkt.write_not_read = 1'b1;
-        done_o = dma_pkt_yumi_i;
+        done_o = 1'b0;
+
+        if (dma_send_fill_addr_i) begin
```
taylor-bsg: the interface of the parallel dma_XXX_i control signals is not great for this code, since it leaves the question of whether they can simultaneously be asserted. perhaps it would be better to have an enum? would it work to define the enum as a one hot enum?

bsg_cache/bsg_cache_dma.v
```diff
@@ -131,113 +151,107 @@ module bsg_cache_dma
...
-    data_mem_v_n = data_mem_v_r;
-    data_buf_n = data_buf_r;
     dma_state_n = IDLE;
-    out_fifo_data_li = data_way_selected;
+    out_fifo_v_li = 1'b0;
+    counter_en = 1'b0;
+    counter_set = 1'b0;
+    counter_val = '0;
 
     case (dma_state_r)
```
taylor-bsg: Comments are good. it would be good to say what is supposed to be accomplished in each state.

bsg_cache/bsg_cache_dma.v
```diff
@@ -121,8 +141,8 @@ module bsg_cache_dma
 
   assign dma_pkt_o = dma_pkt;
   assign dma_pkt.addr = {
```
taylor-bsg: This line of code violates the "don't split struct writes across assign and always_comb block statements" limitation of some tools. Easy to fix here by moving into always_comb block.

tommydcjung:
all the comments are addressed. can we have this merged now?

drichmond:
i'm willing to merge this PR, but @stdavids, what's the standard for merging in BaseJump? Merge, Rebase, or Squash?

stdavids:
Not totally sure, I am not really a maintainer of basejump_stl. My opinion is that it should be squashed -- given that this is all done by a single author and achieves a single goal (refactoring).

tommydcjung:
I think squash makes sense. would be nice to have this merged so I can move onto next thing.

taylor-bsg:
Squash

PR #131, opened 2019-08-17
[bsg_cache] variable associativity/data_width_p. cache line locking.

tommydcjung:
- bsg_cache regression passed
- bsg_cache axe test passed (ways_p=2,4,8)
- bsg_cache dmc test passed (ways_p=2,4,8)
- manycore regression passed


tommydcjung:
https://github.com/bespoke-silicon-group/basejump_stl/issues/74

taylor-bsg:
@vb000 can you take the first pass on a review? Thanks!

taylor-bsg:
Adding @max-ruttenberg -- from your class project, you guys would be quite familiar with the code!

bsg_cache/bsg_cache.v
```diff
@@ -188,15 +200,17 @@ module bsg_cache
...
+  logic [lg_sets_lp-1:0] tag_mem_addr_li;
+  bsg_cache_tag_info_s [ways_p-1:0] tag_mem_data_li;
+  bsg_cache_tag_info_s [ways_p-1:0] tag_mem_w_mask_li;
+  bsg_cache_tag_info_s [ways_p-1:0] tag_mem_data_lo;
   
   bsg_mem_1rw_sync_mask_write_bit #(
-    .width_p((tag_width_lp+1)*2)
+    .width_p(tag_info_width_lp*ways_p)
     ,.els_p(sets_p)
     ,.latch_last_read_p(1)
```
vb000: A bsg_mem question... logic for this seems to be implemented in bsg_mem_1rw_sync_mask_write_bit_synth -- lowest level module in synthesized hierarchy. Isn't it better to move it to higher level to avoid duplicating the logic in hardened modules?

tommydcjung: what "logic" are you referring to?

vb000: Logic related to latch_last_read_p...

tommydcjung: We had a discussion about this long ago... there are several issues when we are modeling hardened SRAM. once we read some location, and write to the same location, we don't want the written value to magically appear at the output. A more conservative measure was taken to make the output to go to 'X, when nothing is being read. However, hardened SRAMs for most technology do retain the last read value at the output so we added an option "latch_last_read_p" so that we don't have to duplicate this logic wherever we use the bsg_mem_1rw_sync. The hardened modules wouldn't need this logic with bsg_dff_en_bypass for the reason mentioned before.

vb000: Do hardened SRAMs retain the old read value even after a new value is written to the same address?

tommydcjung: It depends... if the technology supports it

vb000: Wouldn't this be architectural limitation on what kind of SRAMs can be used with bsg_cache?

dpetrisko: No, because one can always use some surrounding logic to emulate that behavior. 

Basically, the hardened methodology makes it so if that parameter is set, the RTL designer is guaranteed that the module will latch last read. Whether that’s provided by a dff_bypass or the hardened RAM itself isn’t a concern for the designer.

tommydcjung: that's why we have this parameter to add the necessary logic, in case where SRAM does not support it, if the user explicitly wants it..

vb000: Sounds good!

bsg_cache/bsg_cache.v
```diff
@@ -45,13 +51,18 @@ module bsg_cache
     ,output logic v_we_o
```
vb000: What's the purpose of this signal?

tommydcjung: that signal tells the outside world that the instruction is moving from tag-lookup stage to tag-verify stage. It had an use for some metadata outside the cache that need to move together with the corresponding instruction, but it is no longer in use.

vb000: Okay. Wanted to know the use case because, there'd a combo loop if yumi_i is generated with v_we_o in an external module...

tommydcjung: That would be a misuse of the signal.... yumi_i is paired with v_o. Added a comment to clarify.

tommydcjung:
@vb000 @mrutt92 @taylor-bsg are we done with the code reviewing? I have a next PR ready for cache line locking. Can we merge this in so I can put up the next one?

vb000:
Hi @Tommy! I'll finish it by the end of the day.

bsg_cache/bsg_cache.v
```diff
@@ -714,49 +795,65 @@ module bsg_cache
...
-  assign stat_mem_v_li = miss_v
-    ? miss_stat_mem_v_lo
-    : ((st_op_v_r | ld_op_v_r) & v_o & yumi_i);
+  bsg_lru_pseudo_tree_decode #(
+    .ways_p(ways_p)
+  ) plru_decode (
+    .way_id_i(tag_hit_way_id)
+    ,.data_o(plru_decode_data_lo)
+    ,.mask_o(plru_decode_mask_lo)
+  );
```
vb000: This plru decoder seems to be exercised only when there's a hit. So, this can ideally be shared to also decode `chosen_way_n` in `bsg_cache_miss.v`. For that, `chosen_way_n` should be generated in this module, which is possible. Minor detail, but might be worth a thought as decoders grow exponentially.

bsg_cache/bsg_cache.v
```diff
@@ -546,24 +575,35 @@ module bsg_cache
...
+  logic [ways_p-1:0][data_width_p-1:0] sbuf_data_mem_data;
 
-  assign sbuf_data_mem_w_mask = sbuf_set_lo
-    ? {sbuf_mask_lo, {(data_mask_width_lp){1'b0}}}
-    : {{(data_mask_width_lp){1'b0}}, sbuf_mask_lo};
+
+  for (genvar i = 0 ; i < ways_p; i++) begin
+    assign sbuf_data_mem_data[i] = sbuf_entry_lo.data;
+    assign sbuf_data_mem_w_mask[i] = {data_mask_width_lp{sbuf_way_decode[i]}} & sbuf_entry_lo.mask;
+  end
```
vb000: The same muxing and demuxing seems to be repeated in `bsg_cache_dma`. This could be done at a single place (`data_mem` interface). One benefit of that is, `bsg_cache_dma` module would be independent of number of ways.

bsg_cache/bsg_cache.v
```diff
@@ -500,17 +537,14 @@ module bsg_cache
...
   logic sbuf_v_li;
+  bsg_cache_sbuf_entry_s sbuf_entry_li;
 
-  logic [data_width_p-1:0] sbuf_data_lo;
-  logic [addr_width_p-1:0] sbuf_addr_lo;
-  logic sbuf_set_lo;
-  logic [data_mask_width_lp-1:0] sbuf_mask_lo;
   logic sbuf_v_lo;
   logic sbuf_yumi_li;
+  bsg_cache_sbuf_entry_s sbuf_entry_lo;
```
vb000: nice refactor!

bsg_cache/bsg_cache_miss.v
```diff
@@ -75,166 +127,241 @@ module bsg_cache_miss
...
+  );
+
+  bsg_dff_en_bypass #(
+    .width_p(stat_info_width_lp)
+  ) stat_info_buffer (
+    .clk_i(clk_i)
+    ,.en_i(stat_mem_read_r)
+    ,.data_i(stat_info_i)
+    ,.data_o(stat_info_r)
+  );
```
vb000: With `latch_last_read_p` set to 1, do we still need this register? In fact, I think, because of this register `chosen_way_n` would be computed one cycle after we move to `SEND_FILL_ADDR` state. But, since we would never have a DMA that gets done in single cycle there won't be a problem.

bsg_cache/bsg_cache_miss.v
```diff
@@ -260,30 +392,13 @@ module bsg_cache_miss
   always_ff @ (posedge clk_i) begin
     if (reset_i) begin
       miss_state_r <= START;
-      chosen_set_r <= 1'b0;
-      stat_flopped_r <= 1'b0;
+      chosen_way_r <= 1'b0;
+      // added to be a little more X pessimism conservative
```
vb000: Why for X pessimism? I think we should always reset all our flops as a matter of principle =)

bsg_cache/bsg_cache.v
```diff
@@ -98,13 +113,14 @@ module bsg_cache
     ,.tag_read_op_o(tag_read_op)
```
vb000: A general architectural question: is it good idea decode `cache_pkt_i` directly? That means bsg_cache would be encroaching the processor timing path (not an issue in manycore's vcache case). 

Isn't it better to pipeline the input first before decoding? That way, in a use-case where bsg_cache is used an IP block, RTL designer of the processor, without the knowledge of this design, need not worry about losing some slack to the cache.

vb000:
@tommydcjung Awesome design with clean code, tough to find a fault! Did some nitpicking in the comments though =). Nothing critical, look into those time permitting...

bsg_cache/bsg_cache.v
tommydcjung: The cache needs to know what's the incoming instruction is (load, store, tagst, etc), since data_mem and tag_mem are read as the instruction enters the TL stage.

bsg_cache/bsg_cache_miss.v
tommydcjung: the problem was that some flops were not connected to reset previous;y for the interest of saving a bit of area. But now we know that we need to reset all the flops related to ctrl logic.

bsg_cache/bsg_cache_miss.v
tommydcjung: right, we probably don't need those. we just need to make sure that when something is written to SRAM, it does not perturb the output value.

bsg_cache/bsg_cache.v
tommydcjung: okay I could include that in the next PR

bsg_cache/bsg_cache.v
tommydcjung: okay I could include that in the next PR

bsg_cache/bsg_cache.v
tommydcjung: and presumably there is a flop outside the cache holding the incoming cache_pkt.

vb000: I'm not saying we should pipeline `cache_pkt_i` as is, but the pipeline could be designed in a way that that we start by decode something like `cache_pkt_r`.  I think, bsg_cache being a plug n play IP in a processor, should be expecting flopped inputs. And about the input flop, vanilla core's exe stage, for instance, doesn't flop data_mem reqs! Anyway, this comment doesn't expect to change anything in this pr -- just trying to toss an idea.

tommydcjung:
> @tommydcjung How are we verifying PLRU replacement policy?

I could write a regression test to check that manually. another way is to implement a random replacement policy (e.g. bsg_cache_miss_random.v), measure the miss rate on some manycore programs, and then see if PLRU improves the miss rate at all.

mrutt92:
> > @tommydcjung How are we verifying PLRU replacement policy?
> 
> I could write a regression test to check that manually. another way is to implement a random replacement policy (e.g. bsg_cache_miss_random.v), measure the miss rate on some manycore programs, and then see if PLRU improves the miss rate at all.

The second option sounds interesting but does it really confirm that you've implemented the policy you meant to implement? It seems to me like it just shows you've implemented something better than random.

dpetrisko:
> It seems to me like it just shows you've implemented something better than random.

Which is surprisingly hard :) Like Tommy suggests, actual programs may show improvements, but many small tests will/may not

tommydcjung:
@mrutt92 that's why we still need the manual regression test, using TAGLA, TAGLV, etc, to check that the expected way is replaced.

mrutt92:
> > It seems to me like it just shows you've implemented something better than random.
> 
> Which is surprisingly hard :) Like Tommy suggests, actual programs may show improvements, but many small tests will/may not

Sure! But let's imagine for a second that Tommy's accidentally implemented something BETTER than PLRU. If he just tests if he's done better than random and assumes his cache policy is a "correct" PLRU he may never know that he's implemented something awesome by accident.

@tommydcjung cool, sounds good to me.

vb000:
Hi @tommyjung, What manual effort are referring to? Wouldn't it be possible to add a routine in trace generator script to generate cache instruction stream to precisely verify plru?

dpetrisko:
@vb000 the problem is a classic one in verification. In order to verify that the PLRU logic is working, what do you want to check? If you want to see that the algorithm is being followed, you can:

1) check for asymptotic behavior. This is like max suggested, with benchmarks and somewhat saying “if it works, don’t fix it”
2) write targeted tests, with expected inputs and outputs hand verified, like Tommy has (It’s automated to some degree, but it’s still roughly “logic-ed out”
3) reimplement the algorithm in a higher level language, then prove equivalence. This is hard and arguably less valuable. 

2 is a lot of manual effort, especially if the algorithm / interface changes (unlikely)

vb000:
@dan Yeah, I'm suggesting to 3. Yes, it's not a trivial effort to do that in the Python script. But, basejump_stl's goal is to create IPs that many larger projects can use. So, this would give us more confidence in quoting plru to the external world. Also, this effort would pay off when we make more changes. I think we should have automated testing, at least in our long term plans.

dpetrisko:
@vb000 oh, okay! Yes, I’m a huge proponent of automated testing. Just trying to illuminate some issues that come up with ‘verification’. I’d love to see some higher level tests in BaseJump

bsg_misc/bsg_expand.v
```diff
@@ -0,0 +1,24 @@
+/**
+ *  bsg_expand.v
+ *
+ *  This module expands each bit in the input vector by the factor of
+ *  expand_p.
+ *
```
dpetrisko: Maybe add an example?

in_width_p=2, expand_p=4
00 -> 0000_0000
01 -> 0000_1111
10 -> 1111_0000
11 -> 1111_1111

bsg_cache/bsg_cache.v
```diff
@@ -602,63 +597,75 @@ module bsg_cache
     ? snoop_word_lo
     : bypass_data_masked;
 
-  for (genvar i = 0; i < (data_width_p>>3); i++) begin
+  for (genvar i = 0; i < data_mask_width_lp; i++) begin
     assign ld_data_masked[8*i+:8] = {8{mask_v_r[i]}} & snoop_or_ld_data[8*i+:8];
   end
```
taylor-bsg: Is this expand as well?

bsg_misc/bsg_expand.v
```diff
@@ -0,0 +1,24 @@
...
+ *  This module expands each bit in the input vector by the factor of
+ *  expand_p.
+ *  
+ *  @author tommy
+ *
+ */
+
+
+module bsg_expand #(parameter in_width_p="inv", expand_p="inv")
```
taylor-bsg: expand bitmask? Do we see any other forseeable use?

tommydcjung: if we have bsg_decode selecting some way, and need to create a wider bitmask to SRAM, bsg_expand could be used.

tommydcjung: are you saying that the module name should be bsg_expand_bitmask?

taylor-bsg:
yep

bsg_cache/bsg_cache.v
tommydcjung: yep

drichmond:
Squash and Merge?

Issue #125, opened 2019-08-06
deprecate?


taylor-bsg:
https://github.com/bespoke-silicon-group/basejump_stl/blob/a1471e4cdf59fecaf08034eaf072f11a6609a25d/bsg_noc/bsg_wormhole_router.v#L16

Is this router in use anywhere? Are we ready to deprecate it?

dpetrisko:
Not used in BP

dpetrisko:
Doesn't appear to be in manycore, other than a testbench filelist

dpetrisko:
When we deprecate this, we should probably rename bsg_wormhole_router_generalized -> bsg_wormhole_router

dpetrisko:
#137

PR #132, opened 2019-08-21
Slightly modifying wormhole parameters and adding bsg_noc repeater

dpetrisko:
- Modifying concentrators to use cord_width instead of cord markers
- Adding optional parameter to wormhole generalized to allow passing of larger cord_markers arrays. This is necessary, because Verilator doesn't allow slicing unpacked arrays to trim to the correct size.  And localparam int markers[dims_p:0] = (dims_p == 1 ) ? mesh_1d  : mesh_2d isn't supported by VCS or Verilator
- Adding bsg_noc_repeater, used for connecting distant links


dpetrisko:
Also deprecating traditional wormhole router and replacing with generalized router for  #125

dpetrisko:
This needs some work, disregard...

PR #134, opened 2019-08-22
Added XILINX attributes to eliminate FPU timing loops/warnings

drichmond:
These added attributes solve warnings in the Xilinx tool chain related to: 
* Unused registers (Trimmed)
* Resource Sharing

The latter prevents a timing loop from being inferred in the toolchain, or at
least prevents the messages from occurring.

This is preliminary. We don't have a failure case (yet), but we do have the warnings.

### DIFF
--- a/bsg_fpu/bsg_fpu_add_sub.v
+++ b/bsg_fpu/bsg_fpu_add_sub.v
@@ -96,10 +96,18 @@ module bsg_fpu_add_sub
 
   assign larger_exp = (exp_a_less ? exp_b : exp_a) + 1'b1;
 
-  assign exp_diff = exp_a_less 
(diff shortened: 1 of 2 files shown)


drichmond:
Will do. Just to make confirm - in the comments?

tommydcjung:
Looks okay to me. maybe there can be similar issue in bsg_abs?

drichmond:
Added tool version to comment

drichmond:
@tommydcjung I wondered that. But BSG is written for signed numbers, and the exponent abs is for unsigned numbers. They seem to be implemented slightly differently and I dont see any warnings regarding bsg_abs (and it is used)

Issue #145, opened 2019-09-05
pureness in logical expressions


taylor-bsg:
https://github.com/bespoke-silicon-group/basejump_stl/blob/89080b2a11a238df48ce4b55d6e55c6d223f66c2/bsg_mem/bsg_mem_2r1w_sync.v#L81

dpetrisko:
Fixed in #147

Issue #144, opened 2019-09-05
Assertion violates 'Be "pure" in your logical expressions.' from coding guidelines


taylor-bsg:
https://github.com/bespoke-silicon-group/basejump_stl/blob/89080b2a11a238df48ce4b55d6e55c6d223f66c2/hard/gf_14/bsg_mem/bsg_mem_2r1w_sync.v#L117

dpetrisko:
Fixed in #147

PR #147, opened 2019-09-05
Converting bsg_mem_xr1w.v assertions to pure logical expressions

dpetrisko:
#144 #145 #146 

I was unable to parse this part of the style guide

>Purely logical operators for purely non-bitvector inputs. Here you have a bunch of word values that you that you are trying to combine together, C-style into logical boolean expressions. The expression will have a bunch of numerical values that you are testing for equality, less than, greater than, or zero/non-zeroness. You can not have any of the inputs to the expression be bitvector values.  In this case, it is permissible to use logical boolean operators (i.e. &&, ||, if, !), but you should not use any bitwise operators (i.e., &, |, ^), especially the bitwise ~ operator. This is consistent with C usage.

Is using w_v_i and r_v_i in mixed logical expressions permitted? w_v_i == 1'b1 seems verbose


taylor-bsg:
Plain w_v_i is okay in a pure logical expression

dpetrisko:
Great, merge away then

PR #153, opened 2019-09-16
Cocotb Demo (Pseudo PR)

stdavids:
Hey All,

This is a not proper PR but a demo that I'm looking for some feedback on. I've been looking for an excuse to try out Cocotb (which is a python testing infrastructure) so I took a few hours this weekend to try it. I liked it enough to hack up some of my own makefile infrastructure, added a couple test benches and write a few READMEs to kinda explain how I was envisioning a potential testing infrastructure inside basejump_stl. If you have some free cycles feel free to check it out and let me know your thoughts.

Cocotb was really easy to get going and for simple designs it is super straight forward. It uses co-routines to kinda manage time where the hardware is simulating and time where the testbench is checking and driving signals, pretty unique. There is a lot more to it, I've really only scratched the surface. I really just wanted to get a working demo out there. But I also think there is some value in the surrounding infrastructure as well which could be useful with out testing backends that we might want to try out such as PyMTL or even our own trace-driven testing infrastructure.

As for my thoughts on the Pros/Cons of cocotb specifically, the biggest downside is that currently the only "useful" simulator that it supports for us is VCS. If it supported verilator then I would be really happy with it. The good news is that the cocotb developers have stated multiple times in talks which can be found on youtube and in github issues (e.g. https://github.com/cocotb/cocotb/issues/227) that they want to support verilator however there are some VPI calls that I think verilator is lacking support for which is blocking them. I've been keeping my eye on this and just over 2 weeks ago one of the VPI main features was accepted and resolved in verilator... I don't know what that means in terms of the full roadmap but it is a good sign! In fact, some people at Antmacro posted a blog (https://antmicro.com/blog/2019/06/verilog-with-cocotb-and-verilator/) where they modified both verilator and cocotb to get them working together. I haven't tried it out yet, but that is a pretty strong proof-of-concept so I am hopeful. The other major sticking point is that I haven't gotten cocotb to dump a vpd/vcd file but I think that is just my ignorance and I'm sure there is a way to get that done.


dpetrisko:
Nice hacking! I looked at cocotb wayyyyy back when, but stopped when I couldn’t get icarus Verilog to quite work. Curious to see how far they’ve progressed!

PyMTL is also an interesting route. I think having a good infrastructure where we “plug in” testing frameworks would be invaluable, since people are likely to prefer their own, regardless

stdavids:
I like PyMTL, I feel more comfortable writing testbenches in PyMTL because
I don’t have a ton of experience with co-routines, only really came up when
I was building video games way back in high-school :). But cocotb has a
decent following these days which makes me believe that with a bit more
time co-routines will start to click a bit more.

Also, cocotb generates result xml files that are junit compatible so it
integrates with CI out of the box. Minor point but notable.

PyMTL emphasizes hardware description a bit more than the verilog-testing
but the features are there. It is totally coupled with verilator though
from my understanding, which isn’t any worse than cocotb is today with just
VCS but I think cocotb has a real shot at supporting both VCS and verilator
soon.

VCS is obviously important as that is used for our ASIC signoff
verification but I love the idea of just having a virtual box with
verilator, python and gtkwave and having an offline dev box for basejump
stl... that is the dream!

The backend “plugin” nature of the makefile is admittedly fairly weak right
now, another day and a second backend to target and it would be much
better. I do fear that having too many backend infrastructures makes
maintenance a nightmare but I’m also finding that there just isn’t really a
perfect solution. I think our trace driven development ideas are actually
really powerful (would work on any system verilog simulator and is
synthesizable) however there is a huge amount of work required to get that
as user friendly as some of these other python infrastructures.

dpetrisko:
Agree on all points. 1 additional benefit of PyMTL is an eager Cornell development team we can collaborate with :)

tommydcjung:
I think bsg_and and bsg_dff is way too simple for an example. It's a good start though. maybe you can create one for the module with some reasonable complexity like bsg_cache and bsg_fpu? :)

dpetrisko:
> I think bsg_and and bsg_dff is way too simple for an example

Agreed, something managing internal state would be good to see (and integrate toggle coverage into)

dpetrisko:
I took a closer look.  Great job on the infrastructure.  A few quick notes:

- I like the 'model' abstraction.  It might be worth it to come up with an overengineered example of UVM-like abstractions (sequencer, driver, scoreboard, checker).  Then we can evaluate which are useful.
- Having the test logic in Python, but the Verilog-ish infrastructure in Makefiles should make it easy to switch to/between PyMTL, since the models can be reused
- Could it make sense to have the BSG_SWEEP_PARAM logic in the python? Or possibly json config?
- For useful testing, we'll need methods for 1) directed testing 2) constrained random testing 3) truly random testing (for library components, it's sometimes useful to root out violated assertions).  I'm not sure what's all available in Python for 2).  Long time ago, I investigated Hypothesis + cocotb.  IIRC, it was super slow, so I'm not sure if it's worth it.  There may be others.
- Coverage is a killer feature for this -- VCS has a method to do it (Verilator does too, but I guess that doesn't work with cocotb yet).  Long term thought, iterate on constrained random testbench until X% coverage is achieved :)
- Formal is another, very orthogonal angle to this testing.  I think Symbiyosys (Verilog) is still SOTA for this.  But if there's a Python-based alternative we could work into this framework, that would also be very cool

stdavids:
Thanks for the feedback Dan!

> I like the 'model' abstraction. It might be worth it to come up with an overengineered example of UVM-like abstractions (sequencer, driver, scoreboard, checker). Then we can evaluate which are useful.

Yeah I like this idea, this might be some future work to create a full fledged bsg-library for use in the verification language. For Cocotb, one of the examples they have is for a simple DFF but it got tricky quickly. We should talk about this more in person and flush out some ideas to present for the larger testing presentation/talk.

> Having the test logic in Python, but the Verilog-ish infrastructure in Makefiles should make it easy to switch to/between PyMTL, since the models can be reused

I have some PyMTL stuff working and there is even a bit more abstraction that makes it pretty smooth to just between the two. Making models reusable between both is ambitious for all designs (some are obviously very easy), I'd need to stew on it a bit longer.

> Could it make sense to have the BSG_SWEEP_PARAM logic in the python? Or possibly json config?

Yeah, makes way more sense good call! Json feels good? I feel like every infrastructure I use now includes json, I hope people are okay with it hahaha.

> For useful testing, we'll need methods for 1) directed testing 2) constrained random testing 3) truly random testing (for library components, it's sometimes useful to root out violated assertions). I'm not sure what's all available in Python for 2). Long time ago, I investigated Hypothesis + cocotb. IIRC, it was super slow, so I'm not sure if it's worth it. There may be others.

Maybe you can explain (2) a bit more? (1) and (3) are very easy to do in almost every verification language. With Cocotb we also force the seed in the makefile (not python code needed) so the random tests can be replicated.

> Coverage is a killer feature for this -- VCS has a method to do it (Verilator does too, but I guess that doesn't work with cocotb yet). Long term thought, iterate on constrained random testbench until X% coverage is achieved :)

Cocotb has some native support for coverage which we can look into.

> Formal is another, very orthogonal angle to this testing. I think Symbiyosys (Verilog) is still SOTA for this. But if there's a Python-based alternative we could work into this framework, that would also be very cool

Agreed, given the difficulty of formally verifying strong assertions I think a formal verification efforts would be adjacent to the more traditional simulation testing.

dpetrisko:
>Maybe you can explain (2) a bit more?

Sure! Basically, constrained random verification is a concept used to explore a design space in a more targeted way.  For a simple example, consider the difference between 
```
x = $random
y = $random
z = $random
and
x = $random
do
y = $random
while (y > x)
do
z = $random
while (x > z > y)
```
(Ignore that there are mathematical tricks you can do to avoid the while loops here)

Basically, this is what constrained random testing does.  You pass in a set of constraints which can be either combinational (like above) or even sequential (do transaction 1, then transaction 3, then wait 2<N<5 cycles...) depending on tool support.  It lets you construct scenarios much more effectively.

Here's an example (this whole article is really good if you're interested):
http://www.testbench.in/CR_09_INLINE_CONSTRAINT.html

This stuff is built into SystemVerilog but not supported by all simulators.  Advanced constraints are a UVM thing.

Hypothesis is a constrained random framework for python that PyMTL is attempting to integrate.  The extra feature is that for the sequential constraints of Hypothesis, it will find a failure and then shrink the scenario.  So if A->B->C->D->E (where that sequence was selected randomly) causes a failure, Hypothesis might shrink the sequence down to B->C, as long as B->C satisfies the constraints and still causes the failure.

cmarqu:
> * constrained random testing 3) truly random testing (for library components, it's sometimes useful to root out violated assertions

Check out https://github.com/mciepluc/cocotb-coverage for this (ping @mciepluc).

themperek:
>that they want to support verilator however there are some VPI calls that I think verilator is lacking support for which is blocking them

Verilotor kind of works already but may need some special time treatment to be efficient https://github.com/cocotb/cocotb/pull/943.

You can also try https://github.com/cocotb/cocotb/pull/943 ;-)

wallento:
Hi @stdavids, Verilator >= 4.020 have all necessary VPI. If you clone #943 you should be able to test it. Looking forward to your feedback!

stdavids:
@wallento, @themperek and @cmarqu -- Thank you so much for the input! Super excited to give verilator support and constrained random testing a go, both seem very promising :)

Xusine1131:
@stdavids Hi Scott. I've learned Cocotbthese days and I think I can help with this RP. Any task assigned to me?

stdavids:
Hey @Xusine1131, I'm open to suggestions. There is no guarantee that this testing infrastructure is going to ultimately be how we decide to move forward so I'm hesitant to build anything too large in case it dies out. I think @tommydcjung's comments about needing more advanced test cases is probably the most critical thing we are missing right now and that would give us a much clearer vision about the future of cocotb+basejump_stl (and also get some external feedback on the infrastructure I've hacked up). Given your work on the iterative booth multiplier, one potential option would be to create/port some test cases for the iterative booth multiplier?

Xusine1131:
@stdavids Thanks for your suggestion! I'll try it out, and don't hesitate to let me know if you need help!

Xusine1131:
@stdavids Hey Scott I have an idea. Can we combine your JSON configuration with @tanglingshu 's regression configuration file together? They do have some fields in common, like design name, file list, and parameters. Maybe we can define a unified version JSON so that both verification and backend processes can use!

dpetrisko:
Had some spare cycles, so I added an overengineered version of two fifo validation.  I spent a lot more time than I thought setting up the utility classes, but I'm pretty pleased with the result.  All of the input streams used in the tests are manually constrained random  

Still interested in checking out this! https://github.com/mciepluc/cocotb-coverage

dpetrisko:
Just putting this here for notetaking.  Looks like the next steps are to flesh out the library a bit:

1) ReadyValidInputDriver/Monitor
2) ValidYumiInputDriver/Monitor
3) ReadyValidOutputDriver/Monitor
4) ValidYumiOutputDriver/Monitor

We could probably play with inheritance, but most likely I'll just do copy-paste for now

dpetrisko:
@stdavids any suggestion for a simple module which is demanding-demanding?

Xusine1131:
@dpetrisko I like this driver and I have advice for design. Because of the similarity of these drivers,  can we combined them together and use an enumeration type to identify which type of driver we use?
```python
from enum import Enum
class BSGDriverHandshakeType(Enum):
    ReadyValid = 0
    ValidYumi = 1

class BSGDriverDirectionType(Enum):
    Input = 0
    Output = 0

class BSGDriver(Driver):
    def __init__(self, hand_shake: BSGDriverHandshakeType, direction: BSGDriverDirectionType):
        # ....
```
This may help code reuse.

stdavids:
@dpetrisko Demanding interfaces in general are kinda tricky to find because we like helpful interfaces. I thought bsg_noc could be changed but it looks like that actually gets implemented at r&v.

Side question: is there a downside to passing -Wno-fatal to every single verilator sim? Removes needing that flag in every interesting config.json file :)

dpetrisko:
> Demanding interfaces in general are kinda tricky to find because we like helpful interfaces. I thought bsg_noc could be changed but it looks like that actually gets implemented at r&v.

Yeah, fair. Maybe I’ll write a demanding GCD or something just to have something to test. Probably will never use the demanding drivers in basejump, but maybe if this gets used in BlackParrot or something

dpetrisko:
> Side question: is there a downside to passing -Wno-fatal to every single verilator sim? Removes needing that flag in every interesting config.json file :)

Good question! I agree that we don’t want no-fatal in every config.json. The question is if we will want to require waivers for specific warnings. I.e. have no-width-mismatch in the json rather than no-fatal

dpetrisko:
So I think I started to add a test for bsg_mesh_router as a more complicated example and I think I found a bug in cocotb.  

{'_type': 'GPI_MODULE', '_discovered': False, '_def_name': '', '_len': None, '_log': <cocotb.log.SimBaseLog object at 0x7f76ae864e90>, '_handle': 29259776, '_name': 'bsg_mesh_router', '_def_file': '', '_path': 'bsg_mesh_router', '_sub_handles': {'my_y_i': ModifiableObject(bsg_mesh_router.my_y_i), 'my_x_i': ModifiableObject(bsg_mesh_router.my_x_i), 'clk_i': ModifiableObject(bsg_mesh_router.clk_i), 'reset_i': ModifiableObject(bsg_mesh_router.reset_i)}, '_fullname': 'bsg_mesh_router(GPI_MODULE)', '_invalid_sub_handles': {}}

It looks like the ports aren't being detected (namely data_i/data_o).  Possibly because bsg_mesh_router has multiple modules in the same file?  cocotb does have support for multi-dimensional arrays, so that shouldn't be the problem. 
 https://github.com/cocotb/cocotb/blob/master/tests/designs/multi_dimension_array/cocotb_array.sv

dpetrisko:
Thinking a little bit more about this, we should figure out if it's a cocotb problem, a Verilator problem or a cocotb+Verilator problem

Xusine1131:
@stdavids (Unrelated topic) Hi Scott! Similar to the testbench based on Python, can we introduce PyMTL to replace the Python script that generates the code we can not use `generate` statement in SystemVerilog to do so?

For instance, we can write a  PyMTL-based version of bsg_round_robin_arb.py and compare which one is more appropriate. Also, the origin file is for Python2, and it's time to embrace Python3?

For me, I'm working on a Wallace Tree generator with PyMTL, and I think the code is more logical and reusable than the raw Python script.

taylor-bsg:
Hi Xusine,

We always like thinking about how PyMTL could make our lives easier since
the Cornell team are our friends.

For the round robin arb, we use python because there is no support for
generating case statements in Verilog. Does PyMTL allow you to generate
case statements?

M

Xusine1131:
@taylor-bsg Unfortunately, I have searched for `case` in PyMTL source code, and the results are meaningless, so I think currently generating `case` is not available for PyMTL. Maybe an issue to their repository?

For my project, I need to generate instances and connect them dynamically and flexibly, and PyMTL works fine.  And I think bsg_mesh_to_ring_stitch.py is a suitable choice implemented with PyMTL?

dpetrisko:
> It looks like the ports aren't being detected (namely data_i/data_o). Possibly because bsg_mesh_router has multiple modules in the same file? cocotb does have support for multi-dimensional arrays, so that shouldn't be the problem.

Turns out MDAs were the problem: https://github.com/verilator/verilator/issues/2105

dpetrisko:
For the anniversary of this PR, I rebased and updated to the latest (mainline) verilator and cocotb, so we can take another look at this!

PR #156, opened 2019-09-16
Requesting .regression to be added into bsg_misc

tanglingshu:
As a part of the ongoing effort of bsg_ppa_regression, I'm requesting to add some existing run config jsons into bsg_misc.


dpetrisko:
Maybe squash/rebase the 'new edits'?

stdavids:
Hey Ling, awesome stuff! A couple bits of feedback:

1. I think there might actually be too much of a good thing here. Do we have an estimate for the average run time of a single run_config or memory footprint? It looks like there are 2 main design spaces that we are exploring -- parameters and clock speed. I think for now let's just do the 4 corners (small fast, small slow, large fast, large slow)? Have you gotten any requirements from anyone to do "mediums"? I don't hate medium parameters and typical speed but that just jumps our configs from 4 to 9 (125% increase) which hurts. I'd say absolutely no more than 9 configs per module as a base line, that already feels really heavy.

2. What is the "design_size" field for?

PR #161, opened 2019-10-04
Flattened always_ff logic tree in bsg_cache

drichmond:
Makes FPGA compilation correctly infer output register for BRAM

This is a pretty mild optimization. I didn't change any functionality, and I didn't change any logic. However, sometimes the Xilinx synthesis tool struggles to infer output BRAM registers when a certain pattern isn't found. So, I fixed it. I think.

I'd appreciate if someone can run manycore regression with this...


tommydcjung:
I ran the bsg_cache and manycore regression. everything that is expected to pass has passed.

PR #162, opened 2019-10-06
fix the bug with dma evict data read disrupted by incoming load

tommydcjung:
The bug is that when the DMA engine evicts a block, it reads a word and wait for the read word to be accepted by the evict FIFO, before it reads the next word. If the evict FIFO is not ready, and the incoming load takes the empty spot in TL stage (reading tag/data_mem), it disturbs the output of the data_mem, before it is accepted by the evict FIFO. This bug went undetected, because in the previous test setup, the evict FIFO was always able to drain without getting stalled. It was discovered when I added delays between receiving the dma_request and starting accepting/sending main memory data in bsg_nonsynth_dma_model.

The fix is to block the incoming requests when the eviction is in progress. I added a signal dma_evict_o to DMA engine which says the eviction is in progress.


taylor-bsg:
More description of the bug and how you fix it is necessary.

PR #164, opened 2019-10-07
bsg_mem_1r1w assertion on negedge

tommydcjung:


tommydcjung:
mostly used in small FIFOs.

dpetrisko:
Sorry, I meant are there other bsg_mems that incorrectly use posedge assertions? A lot of the mems are copy paste so I could see this issue be in multiple modules

tommydcjung:
yup, I found some more and fixed those as well.

PR #177, opened 2019-10-18
efficient implementation of lru_encode

tommydcjung:
Good evening,

Here is the basic idea. the efficient implementation only needs 4 of 2-to-1 muxes at most.
the current implementation causes huge critical path in the cache design. 

lru_encode


tommydcjung:
@farzamgl can you review this?

dpetrisko:
I'll take a closer look tomorrow, but is the LRU algorithm the same after this change or no? Has the state coverage changed?

tommydcjung:
the algorithm is the same. the hardware is actually pretty simple if you look at it.

dpetrisko:
LGTM; maybe put a link to/embed to the LRU table in the header comment

PR #180, opened 2019-10-30
[cache_to_dmc] add dram size selection capability

gaozihou:
Currently we append cache_id to top bits of dram-request-address, then different caches are mapped to different dram banks, so that cmds from multiple caches can be processed in parallel.

However, we may want to support DRAM of different sizes. In this pull request, cache_id bits can be inserted to multiple places, based on new input port dram_size_i. It supports all possible DRAM sizes that can be purchased from Digikey (256Mb, 512Mb, 1Gb, 2Gb and 4Gb). Test is passing.

### DIFF
--- a/bsg_cache/bsg_cache_to_dram_ctrl.v
+++ b/bsg_cache/bsg_cache_to_dram_ctrl.v
@@ -201,11 +201,11 @@ module bsg_cache_to_dram_ctrl
   // tag_r not used when only 1 cache exists
   always_comb
     case (dram_size_i)
-      0: app_addr_o = {tag_r, addr_r[25-$clog2(num_cache_p)-1:0]};
-      1: app_addr_o = {tag_r, addr_r[26-$clog2(num_cache_p)-1:0]};
-      2: app_addr_o = {tag_r, addr_r[27-$clog2(num_cache_p)-1:0]};
-      3: app_addr_o = {tag_r, addr_r[28-$clog2(num_cache_p)-1:0]};
-      4: app_addr_o = {tag_r, addr_r[29-$clog2(num_cache_p)-1:0]};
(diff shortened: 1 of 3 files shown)


bsg_cache/bsg_cache_to_dram_ctrl.v
```diff
@@ -11,18 +11,23 @@ module bsg_cache_to_dram_ctrl
...
-    , parameter dram_ctrl_burst_len_p="inv"
-
+    
     , localparam mask_width_lp=(data_width_p>>3)
     , localparam lg_num_cache_lp=`BSG_SAFE_CLOG2(num_cache_p)
-    , localparam dram_ctrl_addr_width_lp=(addr_width_p+lg_num_cache_lp)
     , localparam dma_pkt_width_lp=`bsg_cache_dma_pkt_width(addr_width_p)
+
+    , parameter dram_ctrl_burst_len_p="inv"
+    , parameter dram_ctrl_addr_width_p=(addr_width_p+lg_num_cache_lp)
```
tommydcjung: if it's derived from other parameters, this should be "_lp"?

gaozihou: Oh this is because dram_ctrl_addr_width_p may not be equal to (addr_width_p+lg_num_cache_lp), so user can pass in their own parameter.

tommydcjung: or are you saying that this is the default, but it can have different value?

gaozihou: Yeah that was my intention

tommydcjung: I see, thanks

PR #183, opened 2019-11-06
Refactor non-blocking cache regression; bug fix; coverage metric

tommydcjung:
Bug: Invalid way in a set that has a pending DMA request should not be selected for replacement. This bug was discovered by memtest9 with e_vcache_non_blocking_dmc_lpddr4 mode.

Action:
The fix is made in replacement policy. https://github.com/bespoke-silicon-group/basejump_stl/commit/8aa8fc8af90d634671e54b7cb55752fee7b60240
The bug was reproduced in test_block in the regression.
https://github.com/bespoke-silicon-group/basejump_stl/blob/7133ed7e9cdb14fe2b7c2801418d5a709db10f14/testing/bsg_cache/regression_non_blocking/test_block.py
the fix is confirmed.


testing/bsg_cache/regression_non_blocking/exclusion.el
```diff
@@ -0,0 +1,81 @@
+//==================================================
```
drichmond: Was this file supposed to be included?

tommydcjung: yes, that's the exclusion list for coverage

testing/bsg_cache/regression_non_blocking/sv.include
```diff
@@ -55,4 +55,11 @@ $BASEJUMP_STL_DIR/bsg_cache/bsg_cache_non_blocking_dma.v
 $BASEJUMP_STL_DIR/bsg_cache/bsg_cache_non_blocking.v
```
drichmond: Can you make this a makefile fragment instead of a VCS filelist? This way, if someone else edits and re-runs your tests, they won't have to run `make clean` between each iteration

testing/bsg_cache/regression_non_blocking/exclusion.el
drichmond: Was it autogenerated?

tommydcjung: yes

drichmond: I'm afraid I don't understand the purpose of the file, so these questions may seem weird.

Why did you need to include an auto-generated file? How is it used? Why can't it be generated before it's used?

testing/bsg_cache/regression_non_blocking/sv.include
tommydcjung: the problem with makefile fragment is that if some file is missing, it doesn't say which one is missing. I prefer having it this way, because I can continue to run tests even after having some changes.

drichmond: I'm afraid I don't understand what you mean - what doesn't say which one is missing? `make`? `vcs`?

testing/bsg_cache/regression_non_blocking/exclusion.el
tommydcjung: when we open up the coverage report, we need to waive some coverage holes that are valid exclusions (e.g. default case that's never hit). Instead of going through the report each time we open up the report, we save those exclusions in a file, which is loaded next time.

testing/bsg_cache/regression_non_blocking/sv.include
tommydcjung: `make`

testing/bsg_cache/regression_non_blocking/exclusion.el
drichmond: Got it, so this isn't easy to regenerate? If it isn't then I'm fine with this

testing/bsg_cache/regression_non_blocking/sv.include
drichmond: If you include the files in the target dependencies though, it will tell you it can't find that file.

tommydcjung: the message from `make` is not very helpful. It will simple say

make: *** No rule to make target `4x4/simv'.  Stop.

I don't think automatically rebuilding is a critical feature.

drichmond: Is it a pattern rule?

tommydcjung: no it's not. I could delete one of the files expected in the filelist to create this message.

drichmond:
I agree with Dan, I like all the testing work!

PR #185, opened 2019-11-06
Add assertion in round robin verilog generator!

leonardxiang:
The round robin module should have assertion for the inputs_p parameter.

### DIFF
--- a/bsg_misc/bsg_round_robin_arb.py
+++ b/bsg_misc/bsg_round_robin_arb.py
@@ -148,6 +148,18 @@ logic hold_on_sr, reset_on_sr;
 
 """
 
+print """
+// synopsys translate_off
+initial begin
+assert (inputs_p <= %d)
+  else begin
+    $error("[%%m] Can not support inputs_p greater than %%d!", %s);
+    $finish();
+  end
+end
+// synopsys translate_on
+""" % (max_reqs, max_reqs)
+
 for reqs_w in range(1, max_reqs+1):
     print """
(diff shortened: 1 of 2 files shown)


vb000:
Lgtm!

In fact, isn't it better to always generate .v file at compile time and not check in the .v file at all?

drichmond:
I have the same question as @vb000 -  perhaps @proftaylor can adjudicate. 

If this verilog file has a generator, then why is it in the repository?

mrutt92:
++ to @vb000's comment.

dpetrisko:
Basically, the common case that the library is designed for is to be used as is by including Verilog files.  If we included the generator, we would require a user to run a python script upon cloning, or as part of their build flow.  When in reality, it's unlikely that a user would require a rr arbiter with more than 10 max inputs (but if they do, they can run the generator).

drichmond:
OK, so the assert here is very useful then.

dpetrisko:
yep! It might make sense to point to the .py in the assertion.  Something like
"Please run bsg_misc/bsg_round_robin_arb.py to generate larger arbiters"

taylor-bsg:
Great discussion, all!

Possibly with bsg_ppa, we will be able to see what a reasonable upper bound
value. Round robin arbiters can be done with logarithmic depth, so although
they are expensive, they are not ridiculously expensive. I filed an issue.

M

taylor-bsg:
manually merged to master

PR #191, opened 2019-11-21
[bsg_cache_non_blocking] bug fix for clearing dirty bits

tommydcjung:
Bug: dirty bits is not cleared when miss handling without any store miss completes.
Fix: added e_stat_set_lru_and_clear_dirty option to stat_mem, and fixed the logic in MHU FSM.


tommydcjung:
That's right. the coverage report will have some false negatives, and the person who checks needs to manually go through each coverage hole and decide whether it can be waived or not. This saves the repeats.

drichmond:
Is there a python test that can expose this?

tommydcjung:
This is not a correctness bug, but a performance bug. There is no direct way to check that, unless running some experiments.

drichmond:
Ah, got it.

drichmond:
Looks good to me then

PR #192, opened 2019-11-21
Adding edge detector

dpetrisko:


tommydcjung:
lgtm

taylor-bsg:
I think maybe it would be better to remove the reset? The semantics of reset seem at best ambiguous and at worst incorrect. If the input on reset is held to 1, do you want it to register an edge when reset is dropped? 

Also rename input and output to i and o?

dpetrisko:
renaming suggestions make sense.

>I think maybe it would be better to remove the reset? The semantics of reset seem at best ambiguous and at worst incorrect. If the input on reset is held to 1, do you want it to register an edge when reset is dropped?

I agree it's ambiguous.  If we drop it though, then may trigger during reset.  I would say the least surprising semantics are 0 during reset regardless of input signal.  If the signal remains high after reset, that is a rising edge at the start of operation.

Maybe the fix would be:

```
   if (falling_not_rising_p == 1)
     begin : falling
        assign detect_o = ~reset_i & ~sig_i &  sig_r;
     end
   else
     begin : rising
        assign detect_o =  ~reset_i &  sig_i & ~sig_r;
     end
```

taylor-bsg:
I think triggering during reset is okay? If the inputs to a combinational
model change during reset, so too will the output.

The property that we are relying on in our somewhat minimalist HW reset
style is that the outputs of BaseJump STL modules stabilize if reset is
asserted sufficiently long. Probably we should document that in the style
guide...

M

dpetrisko:
> I think triggering during reset is okay? If the inputs to a combinational
model change during reset, so too will the output.

Fair! It's a little unusual because it's not a purely combinational model, but I think I'm convinced.

taylor-bsg:
Okay let’s go with that. I will add the blurb to the style guide.

PR #194, opened 2019-11-26
ramulator files

tommydcjung:


bsg_test/bsg_nonsynth_ramulator_hbm_channel.v
```diff
@@ -0,0 +1,55 @@
...
+    , input [data_width_p-1:0] data_i
+    , output logic data_yumi_o
+
+    // read channel
+    , output logic data_v_o
+    , output logic [data_width_p-1:0] data_o
+  );
+
+
+  logic [data_width_p-1:0] mem [mem_els_p-1:0];
```
mrutt92: Is there a reason why we can't reuse something like bsg_nonsynth_mem_1rw_sync_assoc.v here?
I know this a trivial bit of RTL but it surprises me that there isn't anything we can reuse already in basejump.

tommydcjung: bsg_nonsynth_mem_1rw_sync_assoc uses associative array, and we want to zero out the memory at the beginning.

dpetrisko: Why? Couldn’t zeroing out memory hide bugs?

mrutt92: Oh OK. I guess for testing purposes that's a valuable feature.

mrutt92: sorry @dpetrisko didn't see your comment.

mrutt92: On second thought, I have the same question.

bsg_cache/bsg_cache_to_ramulator_hbm.v
```diff
@@ -0,0 +1,217 @@
+/**
+ *    bsg_cache_to_ramulator_hbm.v 
+ *
+ */
+
+module bsg_cache_to_ramulator_hbm
```
mrutt92: This might be a silly question. How does cache know the address of read data from HBM? I know that the current cache is hit-under-miss so I think that means there should be only one outstanding request at a time? In which case the cache is responsible for keeping track of the address for the outstanding request. Is that understanding correct?

bsg_test/bsg_nonsynth_ramulator_hbm_channel.v
tommydcjung: I remember having this problem with uninitialized value that caused spec benchmark to fail, because the way strlen() checks.

bsg_cache/bsg_cache_to_ramulator_hbm.v
tommydcjung: yes, the cache knows the address of read data.

mrutt92: So this is only good for one outstanding request at a time?

bsg_test/bsg_nonsynth_ramulator_hbm_channel.v
mrutt92: oh. does the assoc array return Xs when there's no data for the input key?

mrutt92: or was it that only some of the bytes in the data word were set?

bsg_cache/bsg_cache_to_ramulator_hbm.v
tommydcjung: HBM channel will return read data in the order the requests arrive (for ramulator). If multiple caches are attached to one channel, and read can complete out-of-order, then this module should be handling the id tracking.

bsg_test/bsg_nonsynth_ramulator_hbm_channel.v
tommydcjung: Accessing the key that does not exist will cause an error and terminate the simulation. Also, you cannot apply byte mask when writing into an associative array.

taylor-bsg:
I think the handling of uninitialized data can be set by a parameter...

bsg_cache/bsg_cache_to_ramulator_hbm.v
mrutt92: OK, well if our design space is just one hit-under-miss cache per channel than this will work just fine.

mrutt92:
The only outstanding comment is pretty small issue. Let's get this merged?

tommydcjung:
Cool thanks. we might want to check if the delay simulated by ramulator is accurate for HBM. I'm not sure what's the best way to go about doing that. There is a testbench setup in testing/bsg_test/bsg_nonsynth_ramulator_hbm to experiment.

mrutt92:
I expanded your test bench a little and ran some experiments to get latency numbers.
I ran tests that stride by column (you had that already), by row, by bank, and by bank group.
I inserted about 200ns of wait time between requests. Latencies are here:

| Stride     | Mean Latency (ns) |
|------------|-------------------|
| column     | 31.18             |
| row        | 47.48             |
| bank       | 56.18             |
| bank group | 58.81             |

These seem pretty close to what we would expect.

drichmond:
Do we have the corresponding numbers for the Xilinx IP?

mrutt92:
I'll work on getting them but I think we can expect them to be on the order 30ns higher or more. The AXI memory controller could have a big impact.

tommydcjung:
yeah, AXI controller might change the numbers. what's the reason for having 200ns of wait between each request?

mrutt92:
To eliminate queuing latency in the controller.

mrutt92:
Here's the latency numbers from the AXI memory controller.

| stride pattern | mean latency (ns) |
|----------------|-------------------|
| column         | 49.59             |
| row            | 55.81             |
| bank           | 65.32             |
| bank group     | 65.48             |

mrutt92:
can we merge this?

PR #195, opened 2019-11-29
Add 3r1w mem

dpetrisko:
Untested, but I was playing around with FPU integration in BP and that requires 3r1w memories for the FP regfile.  They need to be 3r1w to support high throughput FMA

### DIFF
--- a/bsg_mem/bsg_mem_3r1w_sync.v
+++ b/bsg_mem/bsg_mem_3r1w_sync.v
@@ -86,13 +86,13 @@ module bsg_mem_3r1w_sync #(parameter width_p=-1
           assert (w_addr_i < els_p)
             else $error("Invalid address %x to %m of size %x\n", w_addr_i, els_p);
 
-          assert (~(r0_addr_i == w_addr_i && w_v_i && r0_v_i && !read_write_same_addr_p))
+          assert (~(r0_addr_i == w_addr_i && r0_v_i && !read_write_same_addr_p))
             else $error("%m: port 0 Attempt to read and write same address");
 
-          assert (~(r1_addr_i == w_addr_i && w_v_i && r1_v_i && !read_write_same_addr_p))
+          assert (~(r1_addr_i == w_addr_i && r1_v_i && !read_write_same_addr_p))
             else $error("%m: port 1 Attempt to read and write same address");
 
-          assert (~(r2_addr_i == w_addr_i && w_v_i && r2_v_i && !read_write_same_addr_p))
+          assert (~(r2_addr_i == w_addr_i && r2_v_i && !read_write_same_addr_p))
             else $error("%m: port 2 Attempt to read and write same address");
        end
 
--- a/bsg_mem/bsg_mem_3r1w_sync.v
+++ b/bsg_mem/bsg_mem_3r1w_sync.v
@@ -1,0 +1,107 @@
+// MBT 7/7/2016
+// DWP 11/27/2019
+//
+// 3 read-port, 1 write-port ram
+//
+// reads are synchronous
+//
+// although we could merge this with normal bsg_mem_1r1w
+// and select with a parameter, we do not do this because
+// it's typically a very big change to the instantiating code
+// to move to/from sync/async, and we want to reflect this.
+//
+
+module bsg_mem_3r1w_sync #(parameter width_p=-1
+                           , parameter els_p=-1
+                           , parameter read_write_same_addr_p=0
+                           , parameter addr_width_lp=`BSG_SAFE_CLOG2(els_p)
+                           , parameter harden_p=0
+                           , parameter enable_clock_gating_p=0
+                           )
+   (input   clk_i
+    , input reset_i
+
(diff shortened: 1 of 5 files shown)


dpetrisko:
This looks like a huge PR because we still haven't merged #137

dpetrisko:
Has been tested thoroughly in the context of BP+FPU

bsg_mem/bsg_mem_3r1w_sync.v
```diff
@@ -0,0 +1,107 @@
...
+
+//synopsys translate_off
+
+   always_ff @(negedge clk_lo)
+     if (w_v_i)
+       begin
+          assert (w_addr_i < els_p)
+            else $error("Invalid address %x to %m of size %x\n", w_addr_i, els_p);
+
+          assert (~(r0_addr_i == w_addr_i && w_v_i && r0_v_i && !read_write_same_addr_p))
```
tommydcjung: since there is w_v_i in the if, we don't need another w_v_i in assert.

hard/gf_14/bsg_mem/bsg_mem_3r1w_sync.v
```diff
@@ -0,0 +1,118 @@
...
+
+  // no hardened version found
+   begin: notmacro
+     bsg_mem_3r1w_sync_synth #(.width_p(width_p), .els_p(els_p), .read_write_same_addr_p(read_write_same_addr_p), .harden_p(harden_p))
+      synth
+       (.*);
+   end // block: notmacro
+
+  //synopsys translate_off
+  always_ff @(posedge clk_i)
```
tommydcjung: maybe on negedge?

dpetrisko:
Fixed review comments.  Don't merge yet, though, I need to retest

bsg_mem/bsg_mem_3r1w_sync.v
```diff
@@ -0,0 +1,107 @@
...
+//
+
+module bsg_mem_3r1w_sync #(parameter width_p=-1
+                           , parameter els_p=-1
+                           , parameter read_write_same_addr_p=0
+                           , parameter addr_width_lp=`BSG_SAFE_CLOG2(els_p)
+                           , parameter harden_p=0
+                           , parameter enable_clock_gating_p=0
+                           )
+   (input   clk_i
```
tommydcjung: Maybe not within the scope of this PR, but it does seem that hard-macros usually have separate ports for read clock and write clock. It would be more generic to provide "w_clk_i" and "r0_clk_i".

dpetrisko: Oh, interesting.  Maybe @stdavids or @proftaylor can comment on the actual usage of different write and read clocks.  One thing I know they used to do is have write and read shifted by phase so that you can write and read same cycle.

The downside of exposing this at an RTL level in the generic memory level is that all of our references to bsg_mem would need to change, when all the existing code ties the clocks together.

One thing we could do instead is have an additional hierarchy with a thin wrapper tying the clocks together:
bsg_mem_3r1w_sync(input clk_i) -> bsg_mem_3r1w_sync_clocks(input wclk_i, rclk_i) -> bsg_mem_3r1w_sync_synth(input wclk_i, rclk_i) or hard/bsg_mem_3r1w_sync_synth(input wclk_i, rclk_i)

taylor-bsg: Tommy's observation is true. However, I think the goal in this module is to implement the standard notion of a synchronous 3R1W sram; and adding the clocks on the interface would start to make people think that we want them to use multiple clocks. Possibly the version that exposes the multiple clocks would be called _two_clocks. However generally, the two clock rams have a bunch of usages that are disallowed and we would spend a lot of time trying to come up with an interface that would specify this (and also RTL that would somehow convert one type to the other, if somebody specified they had to have a particular behavior.)

taylor-bsg:
An aside: what is the benefit of this module versus: 

https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_mem/bsg_mem_multiport.v

M

dpetrisko:
>https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_mem/bsg_mem_multiport.v

Oh, I guess there's not.  I think it's a similar argument for having mRnW modules vs a single multi-port one.  Someone using an async mem generator may want to harden, and they'll want a substitution.

I think there is an argument for writing multiport and multiport_sync and having mRnW modules wrap around them?

taylor-bsg:
Oh I see .. your module is sync, so that is an important difference. I'm
okay just committing this, and then agreeing that we will not check in a
4r1w. =)

M

dpetrisko:
> Oh I see .. your module is sync, so that is an important difference. 

To be fair, there are both sync and non-sync 3r1w in this commit

> I'm okay just committing this, and then agreeing that we will not check in a
> 4r1w. =)

=)

dpetrisko:
Can we merge?

PR #196, opened 2019-11-30
Bsg cache amoswap

tommydcjung:
- regression v2
- atomic swap/or


testing/bsg_cache/regression/sv.include
```diff
@@ -39,7 +39,7 @@ $BASEJUMP_STL_DIR/bsg_dataflow/bsg_fifo_1r1w_small.v
 $BASEJUMP_STL_DIR/bsg_dataflow/bsg_fifo_1r1w_small_unhardened.v
 $BASEJUMP_STL_DIR/bsg_dataflow/bsg_fifo_tracker.v
 $BASEJUMP_STL_DIR/bsg_cache/bsg_cache_pkg.v
-$BASEJUMP_STL_DIR/bsg_cache/bsg_cache_pkt_decode.v
+$BASEJUMP_STL_DIR/bsg_cache/bsg_cache_decode.v
```
drichmond: Why rename? Is this a second feature in the PR?

drichmond:
Can you add some comments so that we can review this? Just describe some of the changes you made and why (like renaming)

taylor-bsg:
Nice — do we have a plan for line and toggle coverage testing?

tommydcjung:
I think this is more accurate name for the module, since it's not really decoding the pkt.
I will add some comments and add more description in design doc.
I will set up a coverage testbench like the one we have for non-blocking cache.

mrutt92:
It seems like this can be merged? I assume it's been tested as much as can be expected?

PR #201, opened 2019-12-03
Small ramulator changes

mrutt92:
* adds a request trace in the hbm ramulator module
* expands testbench trace to test DRAM latency


testing/bsg_test/bsg_nonsynth_ramulator_hbm/testbench.v
```diff
@@ -101,7 +104,7 @@ module testbench();
       ,.rom_addr_o(rom_addr[i])
       ,.rom_data_i(rom_data[i])
 
-      ,.done_o()
+      ,.done_o(ch_done[i])
```
tommydcjung: This only means that trace_replay is done sending out all the requests, but does not necessary mean that all the requests are completed.

tommydcjung: but I think for the purpose of this testbench, it's not a big deal.

PR #204, opened 2019-12-09
fix the bug with reading mem when w_i=1, but write_mask_i[i]=0

tommydcjung:
if latch_last_read_p=1, last read value is latch at the output. When there is byte-masked write, it reads the bytes that are not masked, therefore destroying the last read value.


dpetrisko:
LGTM.

Do you think it would be more or less readable like: 

```
wire r_li = ~w_i;
wire w_li = w_i & write_mask_i[i];
wire v_li = r_li | w_li;

,.v_i    (v_li)
,.w_i    (w_li)
```

PR #212, opened 2020-01-09
fix #211

### LINKED ISSUE #211: Should be converted into include file with IFNDEF BSG_FPU, since it only contains defines.
https://github.com/bespoke-silicon-group/basejump_stl/blob/f52aaaf3ec588541c2ab61296136d438ee9ffb99/bsg_fpu/bsg_fpu_pkg.v#L2

See https://symbiflow.github.io/sv-tests/#Surelog|basejump|BaseJumpSTL_bsg_fpu_add_sub

Xusine1131:
Question: Why don't we do the same thing to bsg_defines.v? 

I think for dealing with these head files, it's not clear when we should use the `include` macro in systemverilog or just compile the head file with source file together? Maybe we need a unified rule.

### DIFF
--- a/bsg_fpu/bsg_fpu_pkg.v
+++ b/bsg_fpu/bsg_fpu_pkg.v
@@ -5,6 +5,10 @@
  *
  */
 
+ `ifndef BSG_FUP
+
+ `define BSG_FUP
+
 package bsg_fpu_pkg;
 
   // Special numbers
@@ -16,3 +20,5 @@ package bsg_fpu_pkg;
 
 
 endpackage
+
+`endif
--- a/bsg_fpu/bsg_fpu_pkg.vh
+++ b/bsg_fpu/bsg_fpu_pkg.vh
@@ -5,20 +5,12 @@
  *
  */
 
- `ifndef BSG_FUP
-
- `define BSG_FUP
-
-package bsg_fpu_pkg;
-
+`ifndef BSG_FPU
+`define BSG_FPU
   // Special numbers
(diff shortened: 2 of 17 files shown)


taylor-bsg:
This adds the ifdef, but the next steps are to rename the file to .vh, remove the package in this file, and then add the .vh file include to the correct file, and remove the import statement.

bsg_fpu/bsg_fpu_pkg.v
```diff
@@ -5,6 +5,10 @@
  *
  */
 
+ `ifndef BSG_FUP
```
taylor-bsg: typo

bsg_fpu/bsg_fpu_pkg.vh
```diff
@@ -5,14 +5,12 @@
  *
  */
 
-package bsg_fpu_pkg;
-
+`ifndef BSG_FPU
```
tommydcjung: Please rename this file to "bsg_fpu_defines.vh", and use the macro "BSG_FPU_DEFINES_VH" for header guard.

tommydcjung:
could you please rebase this on top of master so I can test it?

tommydcjung:
Actually I could just rebase this in local to test. The tests have passed.

tommydcjung:
https://github.com/bespoke-silicon-group/bsg_manycore/pull/211

Xusine1131:
@tommydcjung Thanks for your reviewing!

PR #214, opened 2020-01-12
Update some script to Python3.

Xusine1131:


Xusine1131:
I wonder whether I should update the script in the hard directory? In my opinion, these scripts will be executed on a server where python3 is not supported, So I don't touch them.

dpetrisko:
Perhaps it's better to convert and use `from __future__ import print_function` so that the scripts are compatible with 2 and 3?

dpetrisko:
As you observed, many users will be running ASIC tools on CentOS, which is supporting Python2 (until 2024, I believe).

Xusine1131:
@dpetrisko Done! Thanks for your advice!

PR #217, opened 2020-01-13
Dramsim3

mrutt92:
* Adds DRAMSim3 as a memory system backend
* Adds BSG_STRINGIFY macro (useful to make macro constants on the command line into strings)
* Provides convenience macros for instantiating HBM memories


bsg_test/bsg_nonsynth_dramsim3_map.v
```diff
@@ -0,0 +1,56 @@
+/////////////////////////////////
+// bsg_nonsynth_dramsim3_map.v //
+/////////////////////////////////
+module bsg_nonsynth_dramsim3_map
```
tommydcjung: please refer to "For reusable modules, omit the outer loop" section in coding guideline.
https://docs.google.com/document/d/1xA5XUzBtz_D6aSyIBQUwFk_kSUdckrfxa2uzGjMgmCU/edit#heading=h.6t9bi53q6bfw

Basically, num_channels_p can be taken out, so that bsg_nonsynth_dramsim3_map is more of a replicable module that can be instantiated using generate-for in the parent module.

bsg_test/bsg_nonsynth_dramsim3.v
```diff
@@ -0,0 +1,221 @@
...
+ *
+ */
+
+module bsg_nonsynth_dramsim3
+  #(parameter channel_addr_width_p="inv"
+    , parameter data_width_p="inv"
+    , parameter num_channels_p="inv"
+    , parameter num_columns_p="inv"
+    , parameter address_mapping_p="inv"
+    , parameter size_p=0
```
tommydcjung: please add comment that the unit is in "bits". (e.g. size of memory in a channel in "bits")

bsg_test/bsg_nonsynth_dramsim3_pkg.v
```diff
@@ -0,0 +1,37 @@
+package bsg_nonsynth_dramsim3_pkg;
+`include "bsg_defines.v"
+
+  localparam int max_cfgs = 128;
+  localparam int lg_max_cfgs = `BSG_SAFE_CLOG2(max_cfgs);
+
+  typedef enum bit [lg_max_cfgs-1:0] {
```
tommydcjung: if there are only 3 enums, and if we don't expect that there will be anymore in future, then we can just set it to "typedef enum bit [1:0]"?

bsg_test/bsg_nonsynth_dramsim3.v
```diff
@@ -0,0 +1,221 @@
...
+  end
+
+  // channels
+  logic [num_channels_p-1:0] read_v_li;
+  logic [num_channels_p-1:0][channel_addr_width_p-1:0] read_addr_li;
+  logic [num_channels_p-1:0] write_v_li;
+
+  for (genvar i = 0; i < num_channels_p; i++) begin
+
+    bsg_nonsynth_ramulator_hbm_channel
```
tommydcjung: If this interface is pretty generic enough so that it can be reused, maybe consider refactoring it to something else? (e.g. bsg_nonsynth_memory_channel) What's missing with this interface, if we were to more closely model the features of the memory?

bsg_test/bsg_nonsynth_dramsim3_pkg.v
```diff
@@ -0,0 +1,37 @@
+package bsg_nonsynth_dramsim3_pkg;
```
tommydcjung: I think this file should just be named "bsg_dramsim3_pkg", since this is just a package. _nonsynth_ really applies "module" objects.

bsg_test/bsg_nonsynth_dramsim3.v
dpetrisko: Maybe width_p rather than size_p? More in line with other basejump modules

dpetrisko: to clarify, width is usually bits and size is usually discrete elements

taylor-bsg: usually width_p and els_p

bsg_test/bsg_nonsynth_dramsim3_pkg.v
taylor-bsg: I would gravitate towards using the symbolic name, but is there a reason to go the other way?

bsg_test/bsg_nonsynth_dramsim3.v
taylor-bsg: Are you saying that we could have a slightly more generic interface, and then it can be parameter that specifes ramulator, dramsim3, etc?

taylor-bsg:
BSG_STRINGIFY -- i like it!

bsg_test/bsg_nonsynth_dramsim3_pkg.v
dpetrisko: Maybe out of scope, but similar to @tommydcjung comment on bsg_nonsynth_test_memory_channel, what about making this package bsg_test_memory_pkg, and having the configs contain both ramulator and dramsim3 and RTL...?  Might help with filename profileration

bsg_test/bsg_nonsynth_dramsim3_pkg.v
taylor-bsg: That's an interesting point you raise. The high-level idea is to put alert people that they are contaminating their synthesizeable code with non-synthesizable modules.  But are you basically saying that you can include this package without affecting the synthesizeability of the code, it is only if they instantiate one of the modules?

bsg_test/bsg_nonsynth_dramsim3.v
tommydcjung: I wouldn't have parameters to pick between ramulator and dramsim3. I would keep two separate so that we can be more agile when we have to make updates. Having a parameter and if-else statements to pick which behavior is usually considered anti-pattern in object-oriented design.

What I'm saying is that we have created this simplified interface for the "memory", presumably without the loss of precision in timing. There is some plumbing code behind it that is shared. My question is Is this generic enough to be shared between HBM1 and HBM2? Is that enough to fully model all the features of the "memory" we need for now? Or is there something fundamentally different in HBM1 and HBM2?

bsg_test/bsg_nonsynth_dramsim3_pkg.v
mrutt92: My thought was that any standard supported by DRAMSim3 could be added as an enum here. I reasoned that this was not synthesizable code anyway so it didn't matter if the type was a little bigger than it really needed to be. Also I was really only using it as a parameter, which made it feel even less important exactly how big it was.

dpetrisko: I agree with @mrutt92.  This makes it so there's no chance of needing to expand it

bsg_test/bsg_nonsynth_dramsim3.v
mrutt92: Given that it's a thin wrapper around a RAM, I do not think this module needs any modification between HBM1 and HBM2. I agree that it might need a more generic sounding name to avoid confusion. I also don't really think it needs modification for other DRAM families.

bsg_test/bsg_nonsynth_dramsim3.v
mrutt92: DRAMSim3 models specific chips which have a fixed capacity. I chose to use size_p because in general I know the size of the chip I want to model and would need to do some (albeit trivial) mental math to get the number of elements given the width. I figured I would save some future brain cells and use the two parameters that I don't need to think too hard about and let the tools do the mental math to derive els_p. But I'm open to push back if that's lame. @tommydcjung I'll add a comment.

taylor-bsg:
Is size in bits, bytes, data_width_p, ... ?

mrutt92:
It is in bits. As @tommydcjung pointed out, it warrants a comment.

taylor-bsg:
Then use size_in_bits_p

mrutt92:
Sounds good prof.

taylor-bsg:
Thanks!

M

bsg_test/bsg_nonsynth_dramsim3_pkg.v
mrutt92: Where did we land on this? I think the only reason you should include this in your file list would be with the intention of using non-synthesizeable code. But I understand where @tommydcjung is coming from.

bsg_test/bsg_nonsynth_dramsim3.v
mrutt92: I think this can be resolved because size_p is now size_in_bits_p.

bsg_test/bsg_nonsynth_dramsim3.v
mrutt92: I've renamed the module, which I think is sufficient for its purpose for using it in both ramulator and dramsim3.

bsg_test/bsg_nonsynth_dramsim3_pkg.v
tommydcjung: I suggest "bsg_dramsim3_pkg" for the package.

bsg_test/bsg_nonsynth_dramsim3_map.v
mrutt92: done.

mrutt92:
Do folks have more feedback? I think everything outstanding has been addressed.

bsg_test/bsg_nonsynth_dramsim3_map.v
```diff
@@ -0,0 +1,51 @@
...
+         (lg_num_channels_lp)'(channel_select_p),
+         {byte_offset_width_lp{1'b0}}
+         };
+  end
+
+  else if (address_mapping_p == e_ro_ra_bg_ba_ch_co) begin
+    assign mem_addr_o
+      = {
+         ch_addr_i[channel_addr_width_p-1:lg_num_channels_lp+lg_num_columns_lp+byte_offset_width_lp],
+         (lg_num_channels_lp)'(channel_select_p),
```
tommydcjung: I suggest having channel_select_p be a port instead. making it a parameter makes it create different hardwares in elaboration time.

mrutt92: I'm curious why that's such a bad thing? Does it have to do with the debugging?

tommydcjung: I'm okay with as it is now, since this is a very small module. Just helpful to think about this kind of issues in future.

PR #218, opened 2020-01-13
Fixing bug in repeater node

dpetrisko:
Refactor brought in an obvious bug.  Accidentally was loopback instead of A<->B

### DIFF
--- a/bsg_noc/bsg_noc_repeater_node.v
+++ b/bsg_noc/bsg_noc_repeater_node.v
@@ -50,9 +50,9 @@ module bsg_noc_repeater_node
      ,.data_i  ( links_A_cast_i.data )
      ,.ready_o ( links_A_cast_o.ready_and_rev )
 
-     ,.v_o     ( links_A_cast_o.v    )
-     ,.data_o  ( links_A_cast_o.data )
-     ,.yumi_i  ( links_A_cast_i.ready_and_rev & links_A_cast_o.v )
+     ,.v_o     ( links_B_cast_o.v    )
+     ,.data_o  ( links_B_cast_o.data )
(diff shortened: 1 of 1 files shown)


bsg_noc/bsg_noc_repeater_node.v
```diff
@@ -50,9 +50,9 @@ module bsg_noc_repeater_node
      ,.data_i  ( links_A_cast_i.data )
      ,.ready_o ( links_A_cast_o.ready_and_rev )
 
-     ,.v_o     ( links_A_cast_o.v    )
-     ,.data_o  ( links_A_cast_o.data )
-     ,.yumi_i  ( links_A_cast_i.ready_and_rev & links_A_cast_o.v )
+     ,.v_o     ( links_B_cast_o.v    )
```
tommydcjung: For the cast signals, please replace with suffix with "_lo" or "_li", since those are not port signals.

dpetrisko: these are port signals though?

tommydcjung: Generally, the difference is that if the signal has "_i" or "_o", it means that it's declared at the top of the module, and exposed to the parent module, whereas "links_B_cast_o" is declared within the scope of this module.

dpetrisko: Refer to page 6 of https://docs.google.com/document/d/1xA5XUzBtz_D6aSyIBQUwFk_kSUdckrfxa2uzGjMgmCU/edit#heading=h.yizm2ivn0gzx

> 
> foo_cast_i:
> Input [..] foo_i, which was a parameterized structure that was passed in with the type erased, and has been cast to that parameterized structure.

PR #220, opened 2020-01-17
Change bsg_wormhole_router to rv& via output_control

dpetrisko:
Needs testing

### DIFF
--- a/bsg_noc/bsg_wormhole_router.v
+++ b/bsg_noc/bsg_wormhole_router.v
@@ -225,9 +225,10 @@ module bsg_wormhole_router
               begin
                 if ((reset_i === 1'b0) && link_o_cast[i].v)
                   begin
-                     // shouldn't really be an error condition; more of indicator that something is not tested enough!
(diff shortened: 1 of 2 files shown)


dpetrisko:
Thanks Paul!

hema0730:
This perfectly eliminated feedthrough paths in bp_tile_node.

taylor-bsg:
Does this comment need to be updated? This should be a perfectly normal condition now?

https://github.com/bespoke-silicon-group/basejump_stl/blob/0e3313335586c399c839249364fc209a6b45cb74/bsg_noc/bsg_wormhole_router.v#L228

M

dpetrisko:
>Does this comment need to be updated? This should be a perfectly normal condition now?

Maybe we should change the assertion into a display?

gaozihou:
> > Does this comment need to be updated? This should be a perfectly normal condition now?
> 
> Maybe we should change the assertion into a display?

Assertion updated!

Issue #223, opened 2020-01-29
DRAMSim3 submodule breaks recursive clone


dpetrisko:
Shouldn't have this as a submodule until it's public.

taylor-bsg:
Agree. We should have a makefile that checks it out instead.

PR #224, opened 2020-01-29
Remove DRAMSim3 submodule

dpetrisko:
#223


mrutt92:
We shouldn't just remove all tracking of DRAMSim3.
See #225 - I think we should go with that one.

dpetrisko:
Fine by me

PR #226, opened 2020-02-04
Dramsim3 timing enable2

mrutt92:
* adds a switch for disabling timing in dramsim3 module - used when skipping IO in simulation


tommydcjung:
I would say hold off on this change, since there is more pressing issue about the interface.

PR #228, opened 2020-02-06
adding latch_last_read_p for ee477/525 cse567 module2 lab of wi20

tanglingshu:
Part of the basejump_stl update for module2. For bsg_tag_ram, the module requires the last read data to be latched. Since mem_1rw_sync is harden swapped, the hardened version of mem_1rw_sync needs to have the parameter as well for the design to be synthesizable.


tanglingshu:
Added more fixes in the next pull request (#229)

PR #232, opened 2020-02-08
Bsg Wormhole Streamer Modules

dpetrisko:
These haven't been completely tested, but I'm making the PR for preliminary feedback.

The idea is to prevent wormhole serialization/deserialization penalties in a flexible, extendible way.  This is done by streaming in #header flits, allowing external logic to determine what to do, and then streaming in #data flits, allowing external logic to stream the data unserialized.

A few notes:
- This still requires a wormhole encoder module for specific protocol logic
- Flit width is decoupled from header width and data width; however, both header and data need to be a multiple of flit width at the moment.  For headers, this is a difficult problem to solve.  For data, we could introduce padding
- There may be bubbles introduced in this first implementation, although I believe they can be totally eliminated with some care


taylor-bsg:
Looks interesting!

gaozihou:
Hi Dan, I think in general the module interface looks great, it is straightforward and easy to use. For step 1 we assume data_width is multiple of wormhole_flit_width, later on we could extend the usage and allow arbitrary hdr and data width. (Current module interface allows us to do this)

**Line-to-line comments:**
I think streamer_out looks nice, just not completed yet
For streamer_in, I have two thoughts:
1. Currently state machine is controlling both before and after piso/sipo. If written correctly this will cause 2 cycles of bubble for each wormhole packet transmitted (in first cycle of "is_hdr" or "is_data" it does not have any wormhole flit sent out). Also it requires the parallel interface be "ready-then-valid", which (I think) is not desirable. I guess we could let state machine control wormhole-link side only (just like what we have in streamer_out), in this way it is bubble-free and parallel interface can naturally become ready_valid_and. (counter parameter set_and_down_exclusive_p=0 will be needed, because we extract wh_hdr after PISO/SIPO)
2. The state-machine should jump from is_hdr to is_data based on count_r_o (this is the main issue that make it not passing the test). Also, instead of counting from len+1 to 0, we could count from len to 0 to get rid of one-cycle bubble.

bsg_noc/bsg_wormhole_stream_in.v
```diff
@@ -102,27 +100,83 @@ module bsg_wormhole_stream_in
...
+     ,.set_i(set_counter)
+     ,.val_i(data_len_li)
+     ,.down_i(data_yumi_li)
+     ,.count_r_o(data_flit_cnt)
      );
-  wire flit_done = (flit_cnt == '0);
+
+  wire e_hdr_to_e_data;
+
+  // Single hdr flit
```
dpetrisko: Why is the condition data_len_li in one and data_flit_done in the other?

gaozihou: When there are multiple hdr flits, data_len_li only meaningful in first hdr flit, so for multi-hdr-flit case we use the registered num_data_flit value (data_counter value) to see whether it is zero-data-flit case.

dpetrisko: Got it!

gaozihou: Thanks Dan! Comments added to clarify this piece of code.

bsg_noc/bsg_wormhole_stream_out.v
```diff
@@ -57,7 +55,7 @@ module bsg_wormhole_stream_out
      );
 
   logic data_v_li, data_ready_lo;
-  assign data_v_li = is_data & link_accept;
+  assign data_v_li = is_data & link_v_i;
```
dpetrisko: Don't these need to be link accept? The incoming link is rv& so even if data_ready_lo is low, data_v_li could be high

bsg_noc/bsg_wormhole_stream_out.v
```diff
@@ -98,6 +96,92 @@ module bsg_wormhole_stream_out
...
+
+  // Set counter value when new packet hdr arrives
+  // and all hdr flits are sent
+  // (set data_flit_counter in same cycle)
+  wire set_counter    = is_hdr & hdr_flit_done & link_v_i;
+
+  bsg_counter_set_down
+   #(.width_p(len_width_p)
+     ,.init_val_p(0)
+     // allow set down same cycle to aviod bubble
```
dpetrisko: nit, avoid

bsg_noc/bsg_wormhole_stream_out.v
```diff
@@ -98,6 +96,92 @@ module bsg_wormhole_stream_out
...
+     ,.count_r_o(data_flit_cnt)
+     );
+
+  wire e_hdr_to_e_data;
+
+  // Single hdr flit
+  if (hdr_len_lp == 1)
+    // When wormhole link accept flit 
+    // and data flit non-zero
+    // and hdr_v_lo (avoid possible X-pessimism in simulation)
```
dpetrisko: Should this comment be link_v_i instead of hdr_v_lo?

bsg_noc/bsg_wormhole_stream_out.v
```diff
@@ -98,6 +96,92 @@ module bsg_wormhole_stream_out
...
+  wire data_flit_last = (data_flit_cnt == (len_width_p)'(1));
+  // All hdr flits are sent
+  wire hdr_flit_done  = (hdr_flit_cnt  == '0);
+  // All data flits are sent
+  wire data_flit_done = (data_flit_cnt == '0);
+
+  // Set counter value when new packet hdr arrives
+  // and all hdr flits are sent
+  // (set data_flit_counter in same cycle)
+  wire set_counter    = is_hdr & hdr_flit_done & link_v_i;
```
dpetrisko: I guess it doesn't matter because we'll just set repeatedly until ready, but is there a reason to use link_v_i instead of link_accept here?

bsg_noc/bsg_wormhole_stream_out.v
gaozihou: https://github.com/bespoke-silicon-group/basejump_stl/blob/c10593e4349a5ca1dfa1ac2b1c0d43d787ba3278/bsg_dataflow/bsg_parallel_in_serial_out.v#L35

This line of comment is out-dated :) It is actually rv&

bsg_noc/bsg_wormhole_stream_out.v
gaozihou: Thanks Dan!

bsg_noc/bsg_wormhole_stream_out.v
gaozihou: Yeah... Forgot to modify comment. 
Actually I am working on a module named "bsg_wormhole_stream_control", which includes the state-machine and counters, so that stream_in and stream_out can instantiate that module to prevent copy-n-paste large portion of code.

bsg_noc/bsg_wormhole_stream_out.v
gaozihou: You are right, link_v_i and link_accept would both function correctly (link_v_i updates counter immediately, link_accept updates counter in last cycle). 

I chose link_v_i for no specific reason... Maybe we should use link_accept to make the code more readable?

dpetrisko: Works for me!

bsg_noc/bsg_wormhole_stream_out.v
dpetrisko: Cool!

bsg_noc/bsg_wormhole_stream_out.v
dpetrisko: Oh, okay!

dpetrisko:
Looks great to me! I'll try to write some better tests

dpetrisko:
Added a few tests.  Achieving 90% line and 80% toggle coverage, probably can get more by sweeping parameters (the tests only use one set of parameters).

dpetrisko:
Hey @gaozihou, did you just change parameters in the trace gen manually? We should add the different tests to the testing/ Makefile and then I think we’re good to go on this

dpetrisko:
Any objections to merging this?

bsg_noc/bsg_wormhole_stream_control.v
```diff
@@ -0,0 +1,111 @@
+
```
taylor-bsg: High level descriptions of semantics and design overview for each of the three modules in comments is in order  =) for such a large bunch of code; or if there is a google doc, link to it.

dpetrisko: Haha guilty. I'll write up the many whiteboard drawings Paul and I went through :)

dpetrisko:
Updated with documentation and high level descriptions of the modules.  Good to go?

dpetrisko:
Bump

dpetrisko:
@taylor-bsg Is there anything else that needs to be done for these modules?

dpetrisko:
@taylor-bsg Can we merge this?  Needed for BlackParrot: https://github.com/black-parrot/black-parrot/pull/618

tommydcjung (approved):
I don't see any major issues here. lgtm

bsg_noc/bsg_wormhole_stream_control.v
```diff
@@ -0,0 +1,124 @@
...
+   , input                   reset_i
+
+   , input [len_width_p-1:0] len_i
+   , input                   link_accept_i
+   
+   , output                  is_hdr_o
+   , output                  is_data_o
+   );
+
+  enum logic [1:0] {e_hdr, e_data} state_n, state_r;
```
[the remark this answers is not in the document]
dpetrisko: fixed!

bsg_noc/bsg_wormhole_stream_in.v
```diff
@@ -0,0 +1,171 @@
...
+//   | data   | data  | data  | data  | protocol info | len   cord |
+//   ---------------------------------------------------------------
+//
+//  Header and data can be sent at the same time, but only 1 message at a time 
+//    is supported. 
+//  - Legal: H     H
+//             D-D   D-D
+//  - Legal: H   H
+//           D-D D-D
+//  - Illegal: H H
```
taylor-bsg, line 26: Clarify what is illegal as a comment e.g., (Header sent too early)

bsg_noc/bsg_wormhole_stream_in.v
```diff
@@ -0,0 +1,171 @@
...
+//             D-D-D-D
+//
+module bsg_wormhole_stream_in
+ #(// The wormhole router protocol information
+   parameter flit_width_p      = "inv"
+   // Default to 0 for cord and cid, so that this can be used either
+   //   for concentrator or router
+   , parameter cord_width_p    = 0
+   , parameter len_width_p     = "inv"
+   , parameter cid_width_p     = 0
```
taylor-bsg, line 36: Define cid

bsg_noc/bsg_wormhole_stream_in.v
```diff
@@ -0,0 +1,171 @@
...
+   // Users may set this directly rather than relying on the protocol header derived default
+   , parameter hdr_width_p = cord_width_p + len_width_p + cid_width_p + pr_hdr_width_p
+   )
+  (input                         clk_i
+   , input                       reset_i
+
+   // The wormhole and protocol header information
+   , input [hdr_width_p-1:0]     hdr_i
+   , input                       hdr_v_i
+   , output                      hdr_ready_o
```
taylor-bsg, line 52: hdr_ready_and_o?

bsg_noc/bsg_wormhole_stream_in.v
```diff
@@ -0,0 +1,171 @@
...
+
+   // The protocol data information
+   , input [pr_data_width_p-1:0] data_i
+   , input                       data_v_i
+   , output                      data_ready_o
```
taylor-bsg, line 57: data_ready_and_o?

bsg_noc/bsg_wormhole_stream_in.v
```diff
@@ -0,0 +1,171 @@
...
+
+   // The input to a wormhole network
+   , output [flit_width_p-1:0]   link_data_o
+   , output                      link_v_o
+   , input                       link_ready_i
```
taylor-bsg, line 62: link_ready_and_i?

bsg_noc/bsg_wormhole_stream_in.v
```diff
@@ -0,0 +1,171 @@
...
+  wire is_hdr, is_data;
+
+  localparam [len_width_p-1:0] hdr_len_lp = `BSG_CDIV(hdr_width_p, flit_width_p);
+
+  wire link_accept = link_ready_i & link_v_o;
+
+  // Header is input all at once and streamed out 1 flit at a time
+  logic [flit_width_p-1:0] hdr_lo;
+  logic hdr_ready_lo, hdr_v_lo, hdr_yumi_li;
+  bsg_parallel_in_serial_out
```
taylor-bsg, line 74: seems like maybe it is worth special casing if hdr_len_lp == 1? If I understand correctly, both the output and input are ready/valid, so the piso could be replaced with pass through wires? if so, I might make is_hdr be is_hdr_r, coming from is_hdr_r_o from the bsg_wormhole_stream_control?

dpetrisko: Special case of hdr_len_lp == 1 degenerates into a two fifo.  Just passing through could result in an unexpected timing path to a router?

bsg_noc/bsg_wormhole_stream_in.v
```diff
@@ -0,0 +1,171 @@
...
+    end
+  else
+    // Protocol data is less than a single flit-sized. We accept a small
+    //   protocol data, aggregate it, and then send it out on the wormhole network
+    begin : narrow
+      localparam [len_width_p-1:0] data_len_lp = `BSG_CDIV(flit_width_p, pr_data_width_p);
+      bsg_serial_in_parallel_out_full
+       #(.width_p(pr_data_width_p)
+         ,.els_p(data_len_lp)
+         ,.use_minimal_buffering_p(1)
```
taylor-bsg, line 126: comment as to why this setting is appropriate

bsg_noc/bsg_wormhole_stream_in.v
```diff
@@ -0,0 +1,171 @@
+
+`include "bsg_defines.v"
+
+//
+// Converts a higher level protocol into a wormhole router without deserializing
+//   the data. This module can be used for converting various DMA formats to 
+//   wormhole flits efficently and with minimal buffering. It can also be used to
```
taylor-bsg, line 7: I think more exposition on each expected use cases would be useful. For example, there are a bunch of limitations on the relation of data width and flit width, so I think that probably implies that there are certain relationships between the input and output wormhole networks that must be obeyed?  When you go between two wormhole networks, do you need to manually translate the len field of the header between the two networks? etc.

bsg_noc/bsg_wormhole_stream_control.v
taylor-bsg: Is there a link to a google doc somewhere here with said whiteboard drawings?

bsg_noc/bsg_wormhole_stream_in.v
taylor-bsg: after you update the interface with ready_and versus ready_then, then I can answer that question better.. =)

taylor-bsg: Which paths are you thinking of? Presumably the ready/valid signals are okay, since if they weren't, the interface is incorrect for the module. For the data signal, it doesn't seem like the logic depth is particularly deep; but let me ask this question; what is the intended use of this module? Is the input connected to an input FIFO coming from a router? Is the output going to the local input of a router, or is it intended to go a long distance across the chip?

bsg_noc/bsg_wormhole_stream_in.v
taylor-bsg: but in this system, you can have multiple output data flits associated with a header, right? So you will have a bubble between data flits?

bsg_noc/bsg_wormhole_stream_in.v
dpetrisko: Added use cases and some description of the width limitation.  The width limitations are entirely driven by the SIPO/PISO-based implementation.  Because this module completely decouples the networks on both sides, a v2 could easily support non-flit-multiple header/data widths

bsg_noc/bsg_wormhole_stream_in.v
dpetrisko: this module will most likely be on the input P-port of a router, locally connected to some driver (bsg_cache dma, etc.).  So for stream_in it will be connected to the input fifo of the router. I think you're right that that's probably fine.

for bsg_wormhole_stream_out, it will be on the output P-port of a router, so no input FIFO.  However, this also probably fine as long as the client is latching inputs as is standard in BaseJump

bsg_noc/bsg_wormhole_stream_in.v
dpetrisko: The bubble happens between consecutive bursts, not consecutive words.

  header  -  data   -  header  -   data
(bubble) A A A A A (bubble) B B B B B

bsg_noc/bsg_wormhole_stream_control.v
dpetrisko: No google doc, and I think the whiteboard drawings are lost to the ravages of time

bsg_noc/bsg_wormhole_stream_in.v
dpetrisko: updated

gaozihou: I think passthrough may cause one cycle bubble between hdr and data flit. One example is bsg_cache_dma.v: https://github.com/bespoke-silicon-group/basejump_stl/blob/bsg_wormhole_streamer/bsg_cache/bsg_cache_dma.v, the "single hdr flit" (dma_pkt) is passthrough, and data flits has two_fifo on output. I had to add an extra two_fifo on dma_pkt port outside bsg_cache_dma module to eliminate the bubble...

The bubble appears when hdr flit and data flit cannot arrive at stream_in module in same cycle (which is not rare, like when we use a state-machine to generate the traffic, we send out hdr in state0, and send out data in state1, then go back to state0...). Preserving the extra buffer for hdr_flit will make this module easier to use (not likely to introduce bubble)

bsg_noc/bsg_wormhole_stream_in.v
gaozihou: I think "use_minimal_buffering_p==1" may introduce bubble, because there might be multiple data flits coming out continuously. Output data of SIPOF are not truly-parallel, they are "wider" data flits that run into wormhole link.

For example:
`header - data - (bbl) - data - header - data - (bbl) - data`
`(bbl) - AAAAA - (bbl) - AAAAA - (bbl) - BBBBB - (bbl) - BBBBB`
In this case there will be extra bubble between two data flits.

dpetrisko: Oh you're right

dpetrisko: removed minimal buffering

bsg_noc/bsg_wormhole_stream_in.v
dpetrisko: Not sure I understand.  Yes, there's a bubble between the header and data, but that's no less latency than having the header in a fifo and come out a cycle later. Well, I guess it would tie up the router port for an extra cycle.  But it would free up the endpoint to continue computing.

I think it's safe with all combinations of passthrough and not passthrough.  Because we have the stream control module arbitrate based on is_hdr and is_data, data coming a cycle early should not cause issues.  I could be wrong though.

That's an interesting point I hadn't thought of though, daisy chaining these modules could unacceptably increase critical path on passthrough.

Could you explain the dma_pkt problem a little more in depth?

gaozihou: Absolutely.

Diagram above shows the cache_dma design. Imagine that the outside world is trying to merge dma_hdr and dma_data traffic (like going into a SIPO, inside cache_dma_to_cce adapter)

Cycle 0 - send out dma_hdr - outside world receive dma_hdr
Cycle 1 - send out data0 - outside world receive nothing (bubble cycle)
Cycle 2 - send out data1 - outside world receive data0
...
Cycle 8 - send out data7 - outside world receive data6
Cycle 9 - cannot send - outside world receive data7
Cycle 10 - send out dma_hdr - outside world receive dma_hdr
...

If we add a two_fifo on dma_hdr, the bubble can be eliminated:

Cycle 0 - send out dma_hdr - outside world receive nothing
Cycle 1 - send out data0 - outside world receive dma_hdr
Cycle 2 - send out data1 - outside world receive data0
...
Cycle 8 - send out data7 - outside world receive data6
Cycle 9 - send out dma_hdr - outside world receive data7
Cycle 10 - send out data0- outside world receive dma_hdr
...

dpetrisko: I guess my point is it's still 10 cycles right? As an unrelated aside, seems like we should add a one/two fifo to the dma hdr?

gaozihou: After adding two_fifo it is 9 cycles (see the updated cycle counts)

dpetrisko: Ah, I see!

dpetrisko: I'm fine with changing it back, @taylor-bsg ?

taylor-bsg:
I have some extended comments that are in a desktop browser window, but I
think the high level takeaway is that our implementation has issues because
the interface to piso assumes ready_and on input and yumi on output, which
requires built in FIFO buffering; but this module is double ready_and and
doesn’t need it. The result is a bunch of unnecessary FIFO plumbing and
related complexity that should not be in the module. A pass thru PISO would
simplify this problem.

dpetrisko:
>The result is a bunch of unnecessary FIFO plumbing and
related complexity that should not be in the module. A pass thru PISO would
simplify this problem.

Do you mean a PISO with no internal buffering at all, so an interface like:
```
clk_i
,reset_i

,data_i
,v_i
,yumi_o

,data_o
,v_o
,yumi_i
```

where a N:1 PISO would send out 1 yumi_o for every N yumi_i's?

gaozihou:
> Do you mean a PISO with no internal buffering at all, so an interface like:
> 
> ```
> clk_i
> ,reset_i
> 
> ,data_i
> ,v_i
> ,yumi_o
> 
> ,data_o
> ,v_o
> ,yumi_i
> ```
> 
> where a N:1 PISO would send out 1 yumi_o for every N yumi_i's?

I think it is a PISO that has ready_and_i on one side and ready_and_o on the other side, without any internal buffering. So do SIPOF.

taylor-bsg:
Dan: yes, except it is ready_and on both sides
Paul: yes, except I don’t know what SIPOF is

taylor-bsg:
FYI I think bsg_round_robin_n_to_1 is essentially Dan’s variant.

dpetrisko:
>Dan: yes, except it is ready_and on both sides

Not sure I understand the semantics here.  If we have no buffering, what does (ready_and_o & v_i) on piso input mean? I can see two possibilities:

1) ready_and_o means that we're ready to accept a new request

Then we need an internal bsg_dff_en_bypass to store the wide data, because once the (ready_and_o & valid_i) are high, the data is acked and gone.

2) ready_and_o means that we're done transmitting.  Then it's more natural as a yumi_o, because the valid goes high when there's new data, and yumi goes high when the data is transmitted.

Which do you mean?

taylor-bsg:
On Wed, Sep 30, 2020 at 9:59 PM Dan Petrisko <notifications@github.com>
wrote:

> Dan: yes, except it is ready_and on both sides
>
> Not sure I understand the semantics here. If we have no buffering, what
> does (ready_and_o & v_i) on piso input mean? I can see two possibilities:
>
>    1. ready_and_o means that we're ready to accept a new request
>
> Then we need an internal bsg_dff_en_bypass to store the wide data, because
> once the (ready_and_o & valid_i) are high, the data is acked and gone.
>
>    1. ready_and_o means that we're done transmitting. Then it's more
>    natural as a yumi_o, because the valid goes high when there's new data, and
>    yumi goes high when the data is transmitted.
>
> Which do you mean?
>

No internal data storage.

From the perspective of the piso:

- v_i means there is upstream data available
- ready_and_o, as in all cases, means that if there was a request (v_i), we
accept it.
- ready_and_i means the downstream module is willing to accept a single
word.
- v_o means that a single word is available for the downstream module to
absorb

practically speaking:
    v_o = v_i
   ready_and_o = ready_and_i & last_word_r
last_word_r is updated based on v_o & ready_and_i

A yumi_o signal is dependent on the incoming valid_i signal of that
interface; while a ready_and_o is not...

dpetrisko:
Updated with passthrough sipo/pisos.  Depends on #345 #346

bsg_dataflow/bsg_serial_in_parallel_out_passthrough.v
```diff
@@ -0,0 +1,79 @@
...
+   ,.count_r_o(valid_r)
+  );
+
+  // If send hi_to_lo, reverse the output data array
+  logic [els_p-1:0][width_p-1:0] data_lo;
+
+  for (genvar i = 0; i < els_p; i++)
+    begin: rof
+      wire my_turn = v_i & (valid_r[i] | ((i == 0) & sending));
+      bsg_dff_en_bypass #(.width_p(width_p)) dff
```
taylor-bsg, line 54: I think only the last element needs to be bypassed?

dpetrisko: Oh great point!

gaozihou: The last element may not need to be registered (if we prefer minimum buffering)

dpetrisko: even better :)

dpetrisko: Fixed in the SIPO PR

bsg_dataflow/bsg_serial_in_parallel_out_passthrough.v
```diff
@@ -0,0 +1,86 @@
...
+       bcoh
+        (.clk_i(clk_i)
+         ,.reset_i(reset_i)
+         ,.clear_i(sending)
+         ,.up_i(receiving & ~count_r[els_p-1])
+         ,.count_r_o(count_r)
+         );
+    end
+
+  // If send hi_to_lo, reverse the output data array
```
taylor-bsg, line 55: misplaced comment

dpetrisko:
Updated again with updated passthrough sipo/pisos. Depends on #345 #346

dpetrisko:
Okay, I think this is finally good to go.  I updated the handshakes to be ready_and throughout to match the passthrough piso/sipos.  All tests pass for wormhole_stream, wormhole_stream_in and wormhole_stream_out.

bsg_noc/bsg_wormhole_stream_in.v
```diff
@@ -0,0 +1,186 @@
...
+       #(.width_p(pr_data_width_p)
+         ,.els_p(data_len_lp)
+         )
+       data_sipo
+        (.clk_i(clk_i)
+         ,.reset_i(reset_i)
+
+         ,.data_i(data_i)
+         ,.v_i(data_v_i)
+         ,.ready_and_o(data_ready_and_o)
```
muwyse, line 149: port is currently named ready_o not ready_and_o

bsg_noc/bsg_wormhole_stream_in.v
```diff
@@ -0,0 +1,186 @@
...
+
+         ,.data_o(data_lo)
+         ,.v_o(data_v_lo)
+         ,.ready_and_i(data_ready_and_li)
```
muwyse, line 153: similarly, port currently yumi_i not ready_and_i

dpetrisko: fixed

dpetrisko:
@taylor-bsg can we get a final approval on this?

muwyse:
@taylor-bsg any timeline on this merge? This is currently blocking BP PR 626.

dpetrisko:
Regarding the wormhole_router arbitration issue we discussed on Tuesday, I think we can merge this before fixing it.  We have an assertion that the data/v_i must not change during the PISO transaction, so if this condition does arise, we'll immediately know what the problem is.

Should make a separate issue to add the locking fixed arbiter as a parameter to wormhole_router though

dpetrisko:
@taylor-bsg 

It seems there actually isn't a problem with the wormhole streamers + wormhole routers as currently implemented.
The arbitration issue only happens when the RR switches in the middle of header transmission, not data

- For router -> streamer connection, the header can be 1 or multiple flits.  The header is aggregated by a passthrough SIPO and doesn't have a problem

- For streamer -> router connection, the header can be multiple flits.  This is a passthrough PISO, so the header needs to be held constant.  However, the input to the PISO is connected to the core logic, not the router, so there is no switching

There would be a problem if we connected a round-robin arbiter on the core logic side of the streamer, but (at least in BP), we don't have a use-case for this, since we handle core-side arbitration with concentrators.  The assertion should be sufficient

dpetrisko:
@taylor-bsg I opened #232. The problematic modules (bsg_mesh_router, bsg_parallel_in_serial_out_passthrough) are not included in this PR, the identified problem is not introduced in this PR, we have an assertion for unsupported usage, and there is not a simple fix.  We should merge this to unblock several major PPA improvements in BlackParrot

dpetrisko:
@taylor-bsg @gaozihou I added an additional round-robin arbiter mode and a test.  Should be good to go

dpetrisko:
Depends on #359

dpetrisko:
Bugfix in #358

dpetrisko:
Coming up on this PR's anniversary.  see round robin arbiter feature here: #359 , otherwise this should be good to go

PR #237, opened 2020-02-15
Ee525 module3 fix

tanglingshu:
in bsg_parallel_in_serial_out moving typedef enum out of the generate scope


taylor-bsg:
To be clear, this change was implemented because formality would not run.

PR #238, opened 2020-02-16
[bsg_circular_ptr] fix width mismatch

tommydcjung:


dpetrisko:
It has to do with slots_p being 32 bits

PR #241, opened 2020-02-21
dramsim3 bandwidth

tommydcjung:
- bsg_cache_to_test_dram
- bsg_cache dma_data_width_p
- dramsim3_bandwidth1/2 testbench


testing/bsg_test/dramsim3_bandwidth2/testbench.v
```diff
@@ -0,0 +1,330 @@
...
+
+    ,.data_v_o(dramsim3_data_v_lo)
+    ,.data_o(dramsim3_data_lo)
+    ,.read_done_ch_addr_o(dramsim3_read_done_ch_addr_lo)
+
+    ,.write_done_o()
+    ,.write_done_ch_addr_o()
+  ); 
+
+  typedef struct packed {
```
mrutt92: this is kinda nice - should we add this to a package?

mrutt92: this might be nice to have in a package.

bsg_cache/bsg_cache_to_test_dram.v
```diff
@@ -0,0 +1,336 @@
...
+    ,.data_o(rr_dma_pkt_lo)
+    ,.tag_o(rr_tag_lo)
+    ,.yumi_i(rr_yumi_li)
+  );
+
+  logic counter_clear;
+  logic counter_up;
+  logic [lg_num_req_lp-1:0] count_r;
+
+  bsg_counter_clear_up #(
```
mrutt92: what is this counting? outstanding requests? please add a comment saying "this counts <blah>"

mrutt92: specifically count_r

mrutt92: or give it a name that describes what it is counting.

bsg_cache/bsg_cache_to_test_dram.v
```diff
@@ -0,0 +1,336 @@
...
+    ,.dram_clk_i(dram_clk_i)
+    ,.dram_reset_i(dram_reset_i)
+
+    ,.dram_data_v_o(dram_data_v_o)
+    ,.dram_data_o(dram_data_o)
+    ,.dram_data_yumi_i(dram_data_yumi_i)
+  );
+ 
+
+/*
```
mrutt92: any reason why this is being kept?

bsg_cache/bsg_cache_to_test_dram.v
```diff
@@ -0,0 +1,336 @@
...
+        {(dram_channel_addr_width_p-cache_bank_addr_width_p){1'b0}},
+        dma_pkt_n.addr[cache_bank_addr_width_p-1:dram_byte_offset_width_lp],
+        {dram_byte_offset_width_lp{1'b0}}
+      };
+    end
+    else begin
+      assign dram_req_addr = {
+        {(dram_channel_addr_width_p-cache_bank_addr_width_p){1'b0}},
+        dma_pkt_n.addr[cache_bank_addr_width_p-1:dram_byte_offset_width_lp+lg_num_req_lp],
+        count_r,
```
mrutt92: so looking at this... my guess is that count is used when a cache line spans multiple columns? and count_r is like the 'column select?'

mrutt92: ok I see - we take cache requests from the round robin arbiter all at once, split them into however many dram requests are needed, transmit them in-order, then clear the counter and accept requests from the next cache.

bsg_cache/bsg_cache_to_test_dram_rx.v
```diff
@@ -0,0 +1,151 @@
...
+
+  if (num_cache_p == 1) begin
+    assign cache_id = 1'b0;
+  end
+  else begin
+    assign cache_id = ch_addr_lo[dram_channel_addr_width_p-1-:lg_num_cache_lp];
+  end
+
+
+  logic [num_cache_p-1:0] cache_sel;
```
mrutt92: unused?

testing/bsg_test/dramsim3_bandwidth2/testbench.v
tommydcjung: this is just an ad-hoc thing. It's up to the user to however they want to map the address.

bsg_cache/bsg_cache_to_test_dram_rx_reorder.v
```diff
@@ -0,0 +1,148 @@
...
+    ,.data_o(dma_data_o)
+    ,.yumi_i(dma_data_v_o & dma_data_ready_i)
+  );
+  
+  if (num_req_lp == 1) begin: req1
+
+    // if there is only one DRAM request, there is no need to reorder.
+    assign piso_data_li = dram_data_i;
+    assign piso_v_li = dram_v_i;
+    wire unused = |dram_ch_addr_i;
```
mrutt92: why is this here?

mrutt92: oh - is this to avoid needing to write the control logic more than once?

tommydcjung: if there is only one cache, there is no need to use dram_ch_addr_i. we just assign it to unused wire to make the lint warning go away.

bsg_cache/bsg_cache_to_test_dram_rx_reorder.v
```diff
@@ -0,0 +1,148 @@
...
+    always_ff @ (posedge core_clk_i) begin
+      if (core_reset_i) begin
+        data_buffer_v_r <= '0;
+      end
+      else begin
+        if (clear_buffer) begin
+          data_buffer_v_r <= '0;
+        end
+        else if (write_buffer) begin
+          data_buffer_v_r[req_num] <= 1'b1;
```
mrutt92: so the requests come in serial - we sort them into a buffer based on the reqnum (recovering the value of 'count_r' from the bsg_cache_to_test_dram module) - and then pass them into the piso so that they come out serial and in-order?

bsg_cache/bsg_cache_to_test_dram_rx_reorder.v
mrutt92: is this is a common thing to do?

bsg_cache/bsg_cache_to_test_dram_rx_reorder.v
```diff
@@ -0,0 +1,148 @@
...
+
+      piso_v_li = 1'b0;
+      piso_data_li = dram_data_i;
+      counter_clear = 1'b0;
+      counter_up = 1'b0;
+      clear_buffer = 1'b0;
+      write_buffer = 1'b0;
+   
+      if (count_r == num_req_lp-1) begin
+        piso_v_li = data_buffer_v_r[count_r] | ((req_num == count_r) & dram_v_i);
```
mrutt92: why is this an 'or' ?

mrutt92: I guess `data_buffer_v_r[count_r]` on some earlier cycle the data at `count_r` became valid and `((req_num == counr_r) & dram_v_i)` means it's become valid this cycle?

mrutt92: cause the piso might not be ready on the cycle when the last request arrives?

bsg_cache/bsg_cache_to_test_dram_rx_reorder.v
```diff
@@ -0,0 +1,148 @@
...
+    // reordering logic 
+
+    wire [lg_num_req_lp-1:0] req_num = dram_ch_addr_i[dram_data_byte_offset_width_lp+:lg_num_req_lp];
+
+    logic [dram_data_width_p-1:0] data_buffer_r [num_req_lp-1:0];
+    logic [num_req_lp-1:0] data_buffer_v_r;
+    
+    logic counter_clear;
+    logic counter_up;
+    logic [lg_num_req_lp-1:0] count_r;
```
mrutt92: so this is counting the number of requests that have arrived and been fed to the piso?

mrutt92: a comment or a more descriptive name would be helpful.

mrutt92: I guess that's not quite right because we stop counting when the piso isn't ready.

mrutt92: I guess count_r == the number of dram requests that have arrived starting from zero such that requests {0..count_r} have all arrived?

mrutt92: so no missing requests between 0 and count_r.

bsg_cache/bsg_cache.v
```diff
@@ -274,25 +310,17 @@ module bsg_cache
     ,.v_o(tag_hit_found)
   );
 
-  logic ld_st_miss;
```
mrutt92: this is just a style change right?

bsg_cache/bsg_cache_to_test_dram_rx_reorder.v
tommydcjung: yup

bsg_cache/bsg_cache.v
tommydcjung: yup

drichmond:
Is this ready to roll?

tommydcjung:
Ideally, this should go with this:
https://github.com/bespoke-silicon-group/bsg_manycore/pull/234

drichmond:
Can you explain to me how they’re connected? 234 doesn’t have much explanation

Does it add memories on the top and bottom?

bsg_cache/bsg_cache_to_test_dram_rx.v
```diff
@@ -0,0 +1,150 @@
...
+
+  assign data_afifo_deq = ch_addr_v_lo & dram_data_v_lo;
+  assign ch_addr_afifo_deq = ch_addr_v_lo & dram_data_v_lo;
+
+
+  // synopsys translate_off
+  
+  always_ff @ (negedge dram_clk_i) begin
+    if (~dram_reset_i & dram_data_v_i) begin
+      assert(~data_afifo_full) else $fatal("data async_fifo full!");
```
mrutt92: is it possible that this assertion is triggered by the request that fills the fifo as opposed to the one that would overflow it?

tommydcjung: I don't see how that would be possible. async fifo has certain capacity and there is a limit on how many outstanding requests can be made.

mrutt92: That's not what I'm asking. I understand that the async fifos capacity is the same as the total number of outstanding requests there can be. I'm asking if this assertion is triggered by a request that the fifo can actually handle i.e. the last request that would fit.

tommydcjung: I'm not sure what the concern is. If the fifo can actually handle, then async_fifo wouldn't have been full?

PR #243, opened 2020-02-27
Adding bsg_dff_reset_en_bypass

dpetrisko:


tommydcjung:
This looks good. Can you just briefly give an example where this can be used?

dpetrisko:
```
output logic v_o

logic v_lo;

bsg_dff_reset_en_bypass
 #(.width_p(1))
 output_reg
  (.clk_i(clk_i)
    ,.reset_i(reset_i)

    ,.en_i(enable_li)
    ,.data_i(v_lo)
    ,.data_o(v_o)
    );
```

replaces:

```
output logic v_o

logic v_lo, v_r;

bsg_dff_reset_en
 #(.width_p(1))
 output_reg
  (.clk_i(clk_i)
    ,.reset_i(reset_i)

    ,.en_i(enable_li)
    ,.data_i(v_lo)
    ,.data_o(v_r)
    );
assign v_o = v_r | (v_lo & enable_li);
```

taylor-bsg:
Seems like kind of a weird interface. If you bypass on cycle 0, it says it
still has the data on cycle 1?

M

dpetrisko:
If you bypass on cycle 0, it does have the data in cycle 1

taylor-bsg:
Right, so the user needs to ignore the valid signal if it bypassed it on
the last cycle?   Wouldn’t it make more sense to have a latency insensitive
interface?

dpetrisko:
Yeah, fair point.  My example is slightly contrived.  A simpler use would be where a normal bsg_dff_en_bypass output causes X-pessimism during initialization.

My example could accomplish the same thing with a bypassable bsg_one_fifo, which I do plan to add

taylor-bsg:
I think the bypassable one element fifo would be great to add (can this be
done as a parameter to a non-bypassable one element fifo?) This module, I
feel like the semantics are not easy to describe without looking at the
code which defeats the purpose?


M

dpetrisko:
>I think the bypassable one element fifo would be great to add (can this be
done as a parameter to a non-bypassable one element fifo?)

#244 

I even had the same question :)

>This module, I feel like the semantics are not easy to describe without looking at the code which defeats the purpose?

Why is bsg_dff_reset_en_bypass more difficult to understand than
https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_misc/bsg_dff_en_bypass.v
?

dpetrisko:
For reference, here is a case where bsg_dff_en_bypass is much cleaner than fifo would be: https://github.com/black-parrot/black-parrot/commit/22bf1002e6fe5d120e27fbf96e9b067ce70eefa4#diff-10b7af0045ee418df054c0e79d1b1199

PR #244, opened 2020-02-28
Adding module to bypass fifos

dpetrisko:
This is useful for non-stalling pipelines, where you need to know that the interface will be ready by the time it propagates down the pipe.

Perhaps we want to combine this with the normal one-fifo (bypassable_p)?  Similar question for bsg_dff_en_bypass and bsg_dff_reset_en_bypass


dpetrisko:
Updated comment

taylor-bsg:
Seems like a lot of code replication. I wonder if we could just have a module called bsg_fifo_bypasser that attaches to existing FIFOs. The interesting question is how we model this in terms of val/ready protocol or demanding/helpfulness.

dpetrisko:
Agreed on code replication.

My thought was just having a bypass_p and then modifying the handshake and data output appropriately.  Your idea is interesting though.  I guess it would look something like:

```
module fifo_bypass
  #(parameter width_p = "inv")
  (// From outside
   input v_i // late
    input data_i // late
    output ready_o // early

    output fifo_data_o // late
    input fifo_ready_i // early
    output fifo_v_o // late

    input fifo_data_i // early
    input fifo_v_i // early
    output fifo_yumi_o // late

    // To outside
    output data_o // late
    output v_o // late
    input yumi_i // late
    )
```

And then the fifo_bypass would instantiate this module along with the fifo.

dpetrisko:
Okay, this is ready.  Tested with full coverage except for the data pins of the fifo.

dpetrisko:
> Violates “be pure”

Fixed!

dpetrisko:
Good to go?

dpetrisko:
Can someone merge this?

PR #245, opened 2020-03-05
Moves Test DRAM Data into C++ vector

mrutt92:


bsg_test/bsg_nonsynth_test_dram_channel.v
```diff
@@ -35,26 +36,71 @@ module bsg_nonsynth_test_dram_channel
...
+  logic 		   data_v_lo;
+
+   always_ff @(negedge clk_i) begin
+      for (integer byte_id = 0; byte_id < data_width_in_bytes_lp; byte_id++) begin
+	 if (read_v_i)
+	   mem_data_lo[byte_id*8+:8] <= bsg_test_dram_channel_get(memory, read_addr_i+byte_id);
+
+      end
+
+      data_v_lo <= '0;
```
dpetrisko: Maybe data_v_lo <= read_v_i?

tommydcjung:
wasn't the plan to make something like bsg_nonsynth_mem_1r1w_sync_dma, so that it can be reused in both here and infinite mem?

mrutt92:
yes but this has been ready for quite some time and we agreed that it would wait until another PR was done and then this would be rebased on top.

bsg_test/bsg_nonsynth_test_dram_channel.v
mrutt92: lol

bsg_test/bsg_nonsynth_test_dram_channel.v
```diff
@@ -35,26 +36,69 @@ module bsg_nonsynth_test_dram_channel
...
+  logic [data_width_p-1:0] mem_data_li;
+  logic 		   write_valid;
+
+  assign write_valid = ~reset_i & write_v_i & data_v_i;
+
+  assign mem_data_li = data_i;
+
+   always_ff @(posedge clk_i) begin
+      for (integer byte_id = 0; byte_id < data_width_in_bytes_lp; byte_id++) begin
+	 if (write_valid)
```
tommydcjung: shouldn't write_valid be outside the for loop?

mrutt92: does it make any semantic difference?

bsg_test/bsg_nonsynth_test_dram_channel.v
```diff
@@ -35,26 +36,69 @@ module bsg_nonsynth_test_dram_channel
...
+	   bsg_test_dram_channel_set(memory, write_addr_i+byte_id, mem_data_li[byte_id*8+:8]);
```
tommydcjung: instead of directly using write_addr_i, it might make more sense to create a signal that zeros out bottom bits, since we don't know if the user is providing the column aligned address.

mrutt92: currently it's expected that an address be aligned to its data width.

bsg_test/bsg_nonsynth_test_dram_channel.v
```diff
@@ -35,26 +36,69 @@ module bsg_nonsynth_test_dram_channel
...
+
+  assign data_v_o = data_v_lo;
+  assign data_o = mem_data_lo;
+
+
+  /////////////////
+  // write logic //
+  /////////////////
+
+  logic [data_width_p-1:0] mem_data_li;
```
tommydcjung: mem_data_li seems redundant

tommydcjung:
did you run the test in dramsim3_bandwidth2?
or I can try to run them

bsg_test/bsg_nonsynth_test_dram_channel.v
tommydcjung: That is the expectation. would be nice to have a comment or assertion

mrutt92:
I did run them - everything passed on my end and the reported bandwidth was the same.
You can run for yourself if you want to double check.

bsg_test/bsg_nonsynth_test_dram_channel.v
tommydcjung: I mean one calls the if-statement multiple times, and the other one doesn't. we can save some trees? haha

bsg_test/bsg_nonsynth_test_dram_channel.v
mrutt92: but also I think this module supports unaligned reads/writes if the expectation isn't met... so maybe a feature not a bug?

PR #246, opened 2020-03-09
[WIP] Making associative memory verilator compatible

dpetrisko:
Not quite infinite, but 63 bit (signed address) associative array is supported in verilator.

Need to make sure this fails well for X (X is an invalid index in a non-wildcard associative array)

### DIFF
--- a/bsg_mem/bsg_nonsynth_mem_1rw_sync_assoc.v
+++ b/bsg_mem/bsg_nonsynth_mem_1rw_sync_assoc.v
@@ -27,12 +27,17 @@ module bsg_nonsynth_mem_1rw_sync_assoc
 
   // associative array
   //
-  logic [width_p-1:0] mem [*];
+  `ifdef VERILATOR
+    // Verilator 4.024 supports associative array, but not wildcard indexed.
+    logic [width_p-1:0] mem [longint];
+  `else
+    logic [width_p-1:0] mem [*];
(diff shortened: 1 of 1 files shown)


dpetrisko:
It's actually probably a benefit to just use longint instead of [*] even for vcs.  Reason being you gain a few convenience methods (next, last, etc.) and also can iterate over the array

https://verificationacademy.com/forums/systemverilog/printing-associative-array

bsg_mem/bsg_nonsynth_mem_1rw_sync_assoc.v
```diff
@@ -27,12 +27,17 @@ module bsg_nonsynth_mem_1rw_sync_assoc
...
+    logic [width_p-1:0] mem [longint];
+  `else
+    logic [width_p-1:0] mem [*];
+  `endif
 
   // write logic
   //
   always_ff @ (posedge clk_i) begin
-    if (v_i & w_i) begin
+    if (~reset & v_i & w_i) begin
```
tommydcjung: reset -> reset_i

dpetrisko: good catch

taylor-bsg:
where is this at? Read for merge?

dpetrisko:
Good to go

PR #247, opened 2020-03-09
[WIP] Adding SR flip flop

dpetrisko:
This can technically be emulated by bsg_dff_en and some auxiliary logic, but the intent seems clearer with this module.  I see similar snippets in BP quite a bit.

```
logic set_li, reset_li;
logic [x_p-1:0] data_li, foo_r;
bsg_dff_en
#(.width_p(x_p))
 foo
 (.clk_i(clk_i)

  ,.en_i(set_li | reset_li)
  ,.data_i(set_li ? data_li : '0)
  ,.data_o(foo_r)
  ); 
```

```
logic set_li, reset_li;
logic [x_p-1:0] data_li, foo_r;
bsg_srff
#(.width_p(x_p))
 foo
 (.clk_i(clk_i)

  ,.set_i(set_li)
  ,.reset_i(reset_li)
  ,.data_i(data_li)
  ,.data_o(foo_r)
  ); 
```

### DIFF
--- a/bsg_misc/bsg_srff.v
+++ b/bsg_misc/bsg_srff.v
@@ -1,0 +1,28 @@
+
+module bsg_srff
+ #(parameter width_p = "inv"
+
+   , parameter logic reset_dominates_set_p = 0
+   )
+  (input clk_i
+
+   , input set_i
+   , input reset_i
+   , input [width_p-1:0] data_i
+
+   , output [width_p-1:0] data_o
+   );
+
+  logic [width_p-1:0] data_r;
+
+  wire do_set = set_i & ~(reset_dominates_set_p & reset_i);
+  wire do_reset = reset_i & ~(~reset_dominates_set_p & set_i);
+
+  always @(posedge clk_i)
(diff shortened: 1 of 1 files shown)


proftaylor:
I hope we don’t have many examples of this in BP!

;-)

Can you point me to some examples?

Set overriding reset is a huge red flag. 🚩

M

dpetrisko:
https://github.com/black-parrot/black-parrot/blob/master/bp_be/src/v/bp_be_checker/bp_be_scheduler.v#L89

It happens a lot with pending bits.  There's a condition where you start a transaction and one where you complete a transaction.  In order to support simultaneous start and finish, you want set to override reset

taylor-bsg:
I think this should be obsoleted by #258?

dpetrisko:
Yep

taylor-bsg:
I think a backburner action item would be to convert some of those BP use
cases over to the new module. I think it will help with X prop in gate
level since the resets will be more clearly designated..

m

PR #249, opened 2020-03-12
BSG_HASH_BANK_UP / DOWN

dpetrisko:
Making `BSG_SAFE_CLOG2(0) == 1

Example code:
```
    // Slice logx when x is greater than 1, else just do 0
    foo = (x > 1)
               ? bar[y-:`BSG_SAFE_CLOG2(x)]
               : '0;
```
If x is 0 here, then foo = '0.  However, bar[y-:`BSG_SAFE_CLOG2(x)] will still get elaborated and cause a compilation issue, even though it will be optimized away.

I don't think this change could break anything, because the old code will just error directly on `BSG_SAFE_CLOG2(0)

### DIFF
--- a/bsg_misc/bsg_defines.v
+++ b/bsg_misc/bsg_defines.v
@@ -4,8 +4,8 @@
 `define BSG_MAX(x,y) (((x)>(y)) ? (x) : (y))
 `define BSG_MIN(x,y) (((x)<(y)) ? (x) : (y))
 
-// maps 1 --> 1 instead of to 0
-`define BSG_SAFE_CLOG2(x) ( ((x)==1) ? 1 : $clog2((x)))
+// maps 0, 1 --> 1 instead of to 0
+`define BSG_SAFE_CLOG2(x) ( ((x)<2) ? 1 : $clog2((x)))
 `define BSG_IS_POW2(x) ( (1 << $clog2(x)) == (x))
 `define BSG_WIDTH(x) ($clog2(x+1))
 
--- a/bsg_misc/bsg_defines.v
+++ b/bsg_misc/bsg_defines.v
@@ -4,8 +4,8 @@
 `define BSG_MAX(x,y) (((x)>(y)) ? (x) : (y))
 `define BSG_MIN(x,y) (((x)<(y)) ? (x) : (y))
 
-// maps 0, 1 --> 1 instead of to 0
-`define BSG_SAFE_CLOG2(x) ( ((x)<2) ? 1 : $clog2((x)))
+// maps 1 --> 1 instead of to 0
+`define BSG_SAFE_CLOG2(x) ( ((x)==1) ? 1 : $clog2((x)))
 `define BSG_IS_POW2(x) ( (1 << $clog2(x)) == (x))
 `define BSG_WIDTH(x) ($clog2(x+1))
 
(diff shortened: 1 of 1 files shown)


taylor-bsg:
My experience is that generate statements often are necessary to handle most such zero width R-value problems.  

Keep in mind that causing an error is also a valid and potentially useful outcome of the macro. 

So for this change, we would have to decide whether doing a clog2 on 0 is more likely an error or somebody trying to do something useful. Clog2 on 1 is arguably usually a valid operation since wires of width 1 are valid and clog2 is used to indicate a pointer to a bit in a bus. Are clog2 of 0 or less frequently valid.

dpetrisko:
The cases I've seen are basically address hashing.  So you're distributing over some set of destinations, but want to special case 0, because you don't want it to depend on the last bit.  I can't think of a situation where it wouldn't be "suspicious" if you returned log2(0) == 1. Like without causing width mismatches

I'm not super passionate about this change, but it seems like a clearer alternative to a generate statement.  To me, expected behavior of a "safe" macro means not error-ing out :)

stdavids:
I get both arguments. 99% of the time I am using clog2(N) to answer the question "How many bits does it require to represent N different values?", very common particularly with addresses. And that question kinda breaks down at N=1 because the real answer is zero bits as the function suggests however N=1 stile implies that that there is something it just so happens to be constant. At N=0 it is not a constant, it is non-existent and so I think the error condition is correct and important as it is right now.

I see the value for what has been proposed with more advanced address hashing, but perhaps the correct solution is a new macro? My main reason is that I don't think we want a commit that opts a bunch of modules into being "okay" with the zero case all of a sudden.

dpetrisko:
`BSG_IMPENETRABLE_LOG2? =)

`BSG_SAFE_SLICE_WIDTH?

stdavids:
Maybe BSG_NOFAIL_CLOG2() or BSG_NEVERFAIL_CLOG2()? And then keep BSG_SAFE_CLOG2 at least for now?

taylor-bsg:
I would suggest maybe going a little more upstream on this. If the
envisioned use case is selecting bits, then perhaps we can have
bitselection macros that do the right thing rather than having a variant of
BSG_SAFE_CLOG2 that doesn't have a direct intuition for what it is used
for..

m

dpetrisko:
```
    // Slice logx when x is greater than 1, else just do 0
    foo = (x > 1)
               ? bar[y-:`BSG_SAFE_CLOG2(x)]
               : '0;
```

``` 
`define BSG_SAFE_HASH_BANK_DOWN(sig_mp, start_mp, els_mp) \
  (els_mp > 1) ? sig_mp[start_mp-:`BSG_SAFE_CLOG2(`BSG_MAX(els_mp, 1)) : '0
`define BSG_SAFE_HASH_BANK_UP(sig_mp, start_mp, els_mp) \
  (els_mp > 1) ? sig_mp[start_mp+:`BSG_SAFE_CLOG2(`BSG_MAX(els_mp, 1)) : '0

foo = `BSG_SAFE_HASH_BANK_DOWN(bar, y, x)
```

dpetrisko:
Something like this?

proftaylor:
maybe with helpful comment that says what the macro means? That will help
me give feedback on the name of the macro too. =)

dpetrisko:
```
    // Slice logx when x is greater than 1, else just do 0
    foo = (x > 1)
               ? bar[y-:`BSG_SAFE_CLOG2(x)]
               : '0;
```

``` 
// Slices an address from the start bit down, distributing among els elements
// If there are 1 or fewer elements in the slice, return 0
`define BSG_SAFE_HASH_BANK_DOWN(sig_mp, start_mp, els_mp) \
  (els_mp > 1) ? sig_mp[start_mp-:`BSG_SAFE_CLOG2(`BSG_MAX(els_mp, 2)) : '0
// Slices an address from the start bit up, distributing among els elements
// If there are 1 or fewer elements in the slice, return 0
`define BSG_SAFE_HASH_BANK_UP(sig_mp, start_mp, els_mp) \
  (els_mp > 1) ? sig_mp[start_mp+:`BSG_SAFE_CLOG2(`BSG_MAX(els_mp, 2)) : '0

foo = `BSG_SAFE_HASH_BANK_DOWN(bar, y, x)
```

dpetrisko:
How does this look?

PR #251, opened 2020-03-13
Adding BSG_CAST_I and BSG_CAST_O

dpetrisko:
I've been using an analogue of these in BlackParrot.  It has a few advantages:
1) Less verbose than the two line variant
2) Enforces x_cast_i / x_cast_o coding style (which we could depend on for meta-macros)
3) It is more explicit than assignment statements, which are often forgotten.

Could also add a suffix parameter, but then we lose out on some of the code reduction.

### DIFF
--- a/bsg_misc/bsg_defines.v
+++ b/bsg_misc/bsg_defines.v
@@ -28,8 +28,8 @@
 
 `define BSG_STRINGIFY(x) `"x`"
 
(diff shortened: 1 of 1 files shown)


dpetrisko:
I tend to err on the side of enforcing code style through construction (auto-linters, macros, etc.).  But I can see the argument for being explicit.

Updated, I think this is good to go

PR #252, opened 2020-03-15
bsg_id_pool

tommydcjung:


dpetrisko:
Is there a power of 2 restriction? 

Maybe have id_width_p as the parameter rather than els_p?

tommydcjung:
There is no such restriction.

taylor-bsg:
Can break the priority encoder into its constituent one hot parts and get rid of the second decode  with v.

PR #254, opened 2020-03-17
saving bsg_decode_with_v

tommydcjung:


bsg_misc/bsg_id_pool.v
```diff
@@ -73,10 +75,11 @@ module bsg_id_pool
...
 
       for (integer i = 0; i < els_p; i++) begin
-        if (dealloc_decode[i])
-          allocated_r[i] <= 1'b0;
-        else if (alloc_decode[i])
+        // this allows immediately allocating the deallocated id.
+        if (alloc_decode[i])
           allocated_r[i] <= 1'b1;
+        else if (dealloc_decode[i])
+          allocated_r[i] <= 1'b0;
```
tommydcjung: this ordering changed so that immediately allocating the deallocated id, but not the other way around. the reasoning is that it has one cycle less delay. Also, if there is an action that allocates and immediately deallocates, it should be very likely that that action can be done without actually allocating.

taylor-bsg:
I think we should have an assert that fires if both are asserted.

dpetrisko:
Seems like there are situations where you want to support either behavior (allowed simultaneous or not).  Is it possible to parameterize?

taylor-bsg:
Let's run into those situations before adding the feature...

m

PR #256, opened 2020-03-18
Parameterizing bsg_idiv_iterative

dpetrisko:
Parameterizing to N-bit divider


tommydcjung:
I would say it's good as long as it is tested thoroughly for both 32 and 64-bit cases

dpetrisko:
Yeah, I don’t know how the PLI stimulus files were generated for 32 bit. I’ll probably change it to a random testbench for 64 bit, which is a shame since I think the stimulus has more edge case focus

taylor-bsg:
We can presumably keep the 32-bit checks in place and still run them in
addition to random testing.

dpetrisko:
Yep, just meant that 64 bit will not have a stimulus, only random

dpetrisko:
Oh, another feature is early-out.  Users may want a 64/32 bit divider but not want the full logic of both.

tommydcjung:
I wonder if it's possible to detect if the value has converged and exit the CALC phase early. For iterative multiplier, I guess you just need to see if any of the upper bits are one.

Xusine1131:
I think the maximum iteration time can be determined by the input width?
Also, a dynamic schedule is appreciated as well. For instance, we can terminate the iteration as soon as we detect that the remnant is zero.

dpetrisko:
Yep, definitely possible to have the early termination.  I think that should be a separate PR though.  This is just intended to allow for width parameterization

dpetrisko:
Tested 64-bit with 10000 inputs of random stimulus -- think this is good to go

dpetrisko:
Good to merge?

taylor-bsg:
Code changes look generally good.  

Regarding testing:

If this is the interface:

module bsg_idiv_iterative #(parameter width_p=32)
    (input                  clk_i
    ,input                  reset_i

    ,input                  v_i      //there is a request
    ,output                 ready_o  //idiv is idle 

    ,input [width_p-1: 0]   dividend_i
    ,input [width_p-1: 0]   divisor_i
    ,input                  signed_div_i

    ,output                 v_o      //result is valid
    ,output [width_p-1: 0]  quotient_o
    ,output [width_p-1: 0]  remainder_o
    ,input                  yumi_i
    );

Do we have tests for

dividend_i[max-1] X divisor_i[max-1] x signed_div_i

and also all combinations of the exceptional cases {0,-1,-2,1,2} x signed_div_i for the inputs 

and also (v_o == 1) X ~yumi_i

also for randomization why stop at 10K?

dpetrisko:
The edge cases are handled the same for 32-bit and 64-bit.  So the testing
philosophy here for the 64-bit parameterization is to add tests for the
ordinary cases.  10k is what the original testbench had

taylor-bsg:
So are those cases covered in the existing testbench? And maybe bump up the
number of random tests?

dpetrisko:
Yes, the original stim covers max/min/zero.  Random stimulus also passes with 24000000 entries

PR #257, opened 2020-03-18
Makes backdoor memory more common purpose

mrutt92:
Makes backdoor memory more common purpose


bsg_mem/bsg_nonsynth_mem_1r1w_sync_mask_write_byte_dma.v
```diff
@@ -0,0 +1,101 @@
...
+	   data_r[byte_id*8+:8] <= bsg_mem_dma_get(memory, read_byte_addr+byte_id);
+
+      end
+     data_v_r <= r_v_i;     
+   end
+
+  // most client code expects outputs to change at the positive edge
+  always_ff @(posedge clk_i) begin
+      data_o <= data_r;    
+      data_v_o <= data_v_r;
```
tommydcjung: This data_v_o is unnecessary. Like in the original interface, it is assumed that the output is always valid one-cycle after read.

mrutt92: Seems harmless and maybe nice to keep in case a client wants to write latency insensitive code?

bsg_mem/bsg_nonsynth_mem_1r1w_sync_dma.v
```diff
@@ -0,0 +1,42 @@
...
+  );
+
+  bsg_nonsynth_mem_1r1w_sync_mask_write_byte_dma
+    #(.width_p(width_p)
+      ,.els_p(els_p)
+      ,.id_p(id_p)
+      ,.init_mem_p(init_mem_p))
+  mem
+    (.*
+     ,.w_mask_i('1));
```
tommydcjung: found a bug

mrutt92: I don't understand. Please explain.

tommydcjung: for 4-bit example, '1 is just 4'b0001. I think what you want is {4{1'b1}}?

mrutt92: http://www.asic-world.com/systemverilog/literal_values1.html

mrutt92: is this site wrong or do I not understand?

tommydcjung: I see. I guess you are right then.

bsg_mem/bsg_nonsynth_mem_1r1w_sync_mask_write_byte_dma.v
```diff
@@ -0,0 +1,101 @@
...
+  ////////////////
+  logic [addr_width_lp+byte_offset_width_lp-1:0] read_byte_addr;
+  assign read_byte_addr = { r_addr_i, {(byte_offset_width_lp){1'b0}} };
+
+  logic [width_p-1:0] data_r;
+  logic data_v_r;
+  
+   always_ff @(negedge clk_i) begin
+      for (integer byte_id = 0; byte_id < data_width_in_bytes_lp; byte_id++) begin
+	 if (r_v_i)
```
tommydcjung: if (r_v_i) should be out of for-loop.

bsg_mem/bsg_nonsynth_mem_1r1w_sync_mask_write_byte_dma.v
tommydcjung: I still don't think it's necessary.. the delay is fixed to be one cycle.

bsg_mem/bsg_nonsynth_mem_1r1w_sync_mask_write_byte_dma.v
```diff
@@ -0,0 +1,101 @@
+/**
+ *    bsg_nonsynth_test_dram_channel.v
+ *
+ */
+
+module bsg_nonsynth_mem_1r1w_sync_mask_write_byte_dma
```
tommydcjung: I think there should be a comment that when reading and writing same addr, read value will be the new value not the old value. Maybe there should be read_write_same_addr_p, and an assertion to protect the user, but I can also see that this module is not trying to model hardened 1r1w sram exactly.

bsg_mem/bsg_nonsynth_mem_1r1w_sync_mask_write_byte_dma.v
tommydcjung: Here, data_v_o  last valid for only one cycle. In fact, once the data is read, the output does not change until new data is read, so technically data can be valid indefinitely. Going further, data at the output is always valid, once any data is read (with the exception data being written, and the memory can be write-through). Also, the user can read the data and use the data however many cycles later as wish. Therefore, I don't think this is a right place to have data_v_o, and for the same reason why there is no data_v_o in the original interface.
data_v_o here is really saying that r_v_i was high one cycle ago. the same thing can be achieved with bsg_dff_reset flopping r_v_i outside of this module.

bsg_test/bsg_nonsynth_dramsim3.v
```diff
@@ -198,24 +198,21 @@ module bsg_nonsynth_dramsim3
...
-    bsg_nonsynth_test_dram_channel
-      #(.channel_addr_width_p(channel_addr_width_p)
-        ,.data_width_p(data_width_p)
-        ,.init_mem_p(init_mem_p)
-        ,.mem_els_p((size_in_bits_p/num_channels_p)/data_width_p)
-        ,.channel_id_p(i))
+    bsg_nonsynth_mem_1r1w_sync_dma
+      #(.width_p(data_width_p)
+        ,.els_p((size_in_bits_p/num_channels_p)/data_width_p)
+        ,.id_p(i)
```
tommydcjung: Wouldn't this limit from having multiple instantiations of this modules in future? There should be a parameter for base id, having default value of zero.

mrutt92: You're right. I was thinking of something similar except making id's a string identifier instead of an integer.

bsg_mem/bsg_nonsynth_mem_1r1w_sync_mask_write_byte_dma.v
mrutt92: Well that's pretty hard to argue with. Nice explanation!

bsg_test/bsg_nonsynth_dramsim3.v
mrutt92: otoh can you do something like this in sv?
```sv
string p = sformat("some string", some parameters...);
some_module #(.string_p(p)) m0 (...);
```

tommydcjung: I would stick with integer, since verilog does not like passing non-static value as parameter.

bsg_mem/bsg_mem_dma.cpp
```diff
@@ -1,59 +1,63 @@
...
 {
     Memory *memory = reinterpret_cast<Memory*>(handle);
-    pr_dbg("getting 0x%08llx   (%02x)\n", addr, memory->get(addr));
+    pr_dbg("id = %llu: getting 0x%08llx   (%02x)\n", memory->_id, addr, memory->get(addr));
     return memory->get(addr);
 }
 
-extern "C" void bsg_test_dram_channel_set(
+extern "C" void bsg_mem_dma_set(
     void *handle,
```
tommydcjung: Why is it necessary to have this handle reference saved in verilog? Why not just have map<int,vector<byte>> as a static variable, and the module passes in its ID into these DPI calls?

mrutt92: Saving the pointer means you don't need to do a lookup every time you make the DPI calls.

PR #258, opened 2020-03-19
bsg_dff_reset_set_clear

tommydcjung:


dpetrisko:
I like your interface better than #247.  Do stdcell set/reset flops usually have a separate reset? @stdavids?

tommydcjung:
I think what you are trying to describe is DFF with synchronous clear/set. In some pdks (if not all), it's forbidden to assert both set and clear at the same time. So I think it's just better to provide high-level interface, and let the tool figure out how they want to implement the circuit.

dpetrisko:
Yeah, fair. My thought was to tie off the set/clear combination to some predictable state. But at that point, might as well keep the interface the way it is here.

stdavids:
Just to circle back to @dpetrisko, there is not really a usual for standard cells especially with flip-flops, so I hesitate to make sweeping claims about PDKs. However, in an effort to try and be useful, I will say that from my experience it is not very common to have DFFs with both set and reset (aside from scan-chain DFFs), but if such a cell is in your pdk it is fairly common that only one can be asserted like Tommy suggested.

PR #263, opened 2020-03-25
Adding quite_p parameter to suppress display statements.

stdavids:
Adding the ability to turn off displays in trace-replay to speed up simulations (for large trace roms used for initialization this can be significant). I don't consider these display statements "debug" messages so I called the parameter quiet_p. Default is 0 therefore by default the displays are shown (same behavior as before).

### DIFF
--- a/bsg_fsb/bsg_fsb_node_trace_replay.v
+++ b/bsg_fsb/bsg_fsb_node_trace_replay.v
@@ -17,6 +17,7 @@ module bsg_fsb_node_trace_replay
   #(parameter   ring_width_p=80
     , parameter rom_addr_width_p=6
     , parameter counter_width_p=`BSG_MIN(ring_width_p,16)
+    , parameter quiet_p=0
     )
    (input clk_i
     , input reset_i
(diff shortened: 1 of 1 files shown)


dpetrisko:
1) Seems like done, finish, unknown opcode and recv mismatch shouldn't be quietable
2) Should these changes be applied to bsg_test/bsg_trace_replay as well?

taylor-bsg:
The way we have done this before is to have a debug_p level. What we have
not done well before is to define what those levels mean in the header
comments. :-)

I think errors should always be on, and maybe debug_p = 0 has nothing and
debug_p=1, the default for now, prints finish/done/start, and 2 is
everything.

Yeah before we start to have divergence, maybe we should deprecate the fsb
version, but we should check if it had any new features that need to be
merged?

taylor-bsg:
Make sure to have a begin / end for one line case statement items that you
are adding a generate if to. Too much of a semantics corner case.

stdavids:
Ah okay, this is good feedback. I wasn't aware of bsg_test/bsg_trace_replay.v. I think I should close this PR and start a new one on bsg_test/bsg_trace_replay.v. Since bsg_fsb_node_trace_replay.v is being deprecated, I don't think we need these changes here.

I will also update the messaging levels in the new PR.

PR #264, opened 2020-03-25
Adding debug_p parameter to control how verbose the display messages are.

stdavids:
Continued from PR #263 

Moved changes to bsg_test/bsg_trace_replay.v. Addressed the feedback by having multiple debug levels with comments about what level does what. More explicit begin-end to remove scope ambiguity.

### DIFF
--- a/bsg_test/bsg_trace_replay.v
+++ b/bsg_test/bsg_trace_replay.v
@@ -5,10 +5,16 @@
 // trace format (see enum below)
 //
 
+// Debug Display Levels:
+//            0 = suppress all non-errors. only shows mismatch and unknown op
+//  (default) 1 = adds finish and done
+//            2 = all messages. adds send, recv, cycle dec, and cycle init
+
 module bsg_trace_replay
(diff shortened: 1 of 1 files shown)


dpetrisko:
Changes look good to me. Possibly out of scope: we could add BSG_LEVEL_NONE BSG_LEVEL_ERROR, BSG_LEVEL_INFO, BSG_LEVEL_VERBOSE to bsg_defines as standard debug levels

taylor-bsg:
Maybe add as an issue: standard debug levels?  One idea to throw in there
is that we could have a bit vector with common cases (which we would have
to identify.)

Xusine1131:
Maybe we can refer to UVM_INFO and build a similar function printing message with a given debug level?

tommydcjung:
Can the case statement with default case (line 210), be merged with the case statement above it? so that the code is left cleaner than it was found.

taylor-bsg:
Good pointer Shanqing!

PR #265, opened 2020-03-26
Merging case statements for code clarity

stdavids:
Simple code refactor addressing feedback in #264 from @tommydcjung.

### DIFF
--- a/bsg_test/bsg_trace_replay.v
+++ b/bsg_test/bsg_trace_replay.v
@@ -201,19 +201,11 @@ module bsg_trace_replay
                          $display("### bsg_trace_replay CYCLE INIT = %x (%m)",cycle_ctr_n);
                     end
                end
-               default:
-                 begin
-
-                 end
+               default: begin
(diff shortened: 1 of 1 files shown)


bsg_test/bsg_trace_replay.v
```diff
@@ -201,19 +201,11 @@ module bsg_trace_replay
...
-                 end
-             endcase // case (op)
-
-             case (op)
-               eNop, eSend, eReceive, eDone, eFinish, eCycleDec, eCycleInit:
-                 begin
-                 end
-               default: $display("### bsg_trace_replay UNKNOWN op %x (%m)\n", op);
+               default: begin
+                    $display("### bsg_trace_replay UNKNOWN op %x (%m)\n", op);
```
tommydcjung: can this be $error() instead of $display()?

stdavids: I think this is a good suggestion, I'm going to hold off on this PR because there is actually a bug (or what I think is a bug) that prevents this message from ever being displayed.

PR #267, opened 2020-03-30
store buffer bug fix

tommydcjung:
Synopsis of bug:
when there is a load in TL stage, which needs to bypass store buffer to get the latest data, is stalled, because yumi_o=0, the store buffer writes to DMEM, and the load in TL stage misses the chance to bypass the data.

Fix for the bug:
do not allow store buffer to write when this condition occurs.

QA:
test_store_load.py in regression_v2 replicates the bug.


bsg_cache/bsg_cache.v
```diff
@@ -922,7 +922,15 @@ module bsg_cache
   assign sbuf_v_li = (decode_v_r.st_op | decode_v_r.atomic_op) & v_o & yumi_i;
   assign sbuf_entry_li.way_id = miss_v ? chosen_way_lo : tag_hit_way_id;
   assign sbuf_entry_li.addr = addr_v_r;
-  assign sbuf_yumi_li = sbuf_v_lo & ~((decode.ld_op | decode.atomic_op) & v_i & ready_o) & (~dma_data_mem_v_lo);
```
dpetrisko: I'm a little confused on the mapping of logic to condition 4)

Does miss_v correspond to bypass from sbuf somehow?

tommydcjung: During miss_v, we want the store buffer to drain, and at the end of the miss, TL is replayed so there is no hazard there. But this bug is specific to the condition when there is no miss.

dpetrisko: Oh, I think I understand.

> TL read DMEM (and bypass from sbuf), and TV is not stalled (v_we).

Maybe add a comment about draining the store buffer during misses? I was mostly confused since it just seems like an extra term. 

To be it would be more clear to DeMorgan it out and do “| miss”, but that’s a soft suggestion

taylor-bsg:
Feedback: Add comments to address the invariants and to identify these
special cases and make the intent of the code more clear.

dpetrisko:
Funnily enough, we just encountered this bug in BP during the cache parameterization work.  This fix worked there, as well.

PR #268, opened 2020-03-31
separate bsg_crossbar_control_o_by_i

tommydcjung:


dpetrisko:
From a glance, looks pretty good from a cleanliness perspective.  Do we have another use for bsg_crossbar_control_o_by_i?

taylor-bsg:
Was that last commit a bug fix to the original code? Maybe then do that as
a separate commit?

taylor-bsg:
just an update on this, I believe I committed the bug fix to the original, so this pull request could be resubmitted just to factor the module, which is a good idea.

bsg_misc/bsg_crossbar_control_o_by_i.v
```diff
@@ -0,0 +1,121 @@
+/* bank mem banked crossbar */
```
dpetrisko, line 1: Obsolete comments

taylor-bsg:
Actually, we have since implemented bsg_crossbar_control_basic_o_by_i, a new, simpler module, and renamed this module to bsg_mem_banked_crossbar_control_o_by_i, so we can close this.

https://github.com/bespoke-silicon-group/basejump_stl/commit/101b0a37a347ef2f6387f60d677c6ccdc9fcaf2d

PR #269, opened 2020-04-02
Redirecting to public version of dramsim3

dpetrisko:
Minor thing: The remote repo is called "DRAMsim3" not "DRAMSim3".  We should probably update paths to avoid any weird issues in the future.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,5 +3,4 @@
 	url = https://github.com/bespoke-silicon-group/ramulator
 [submodule "DRAMSim3"]
 	path = imports/DRAMSim3
-	url = git@github.com:bespoke-silicon-group/bsg_dramsim3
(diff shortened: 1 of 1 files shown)


mrutt92:
We made some edits - we might want to make a public fork of UMDs version and checkin our changes.

dpetrisko:
That works as well.  Or make a PR back ;)

dpetrisko:
@mrutt92 , I added a public fork to bsg-external, if you want to put your changes up there and bump the submodule

drichmond:
Fixed in #304 ?

dpetrisko:
I mean yeah, seeing as it's the same change :)

drichmond:
Same, but with some of the changes we made added on

PR #270, opened 2020-04-04
[RFC] Adding async noc link

dpetrisko:
This is a module which connects two bsg_noc links which are in different clock domains.  This is a useful wrapper because CDCs are confusing.  People are more likely to use our building blocks with helper classes like this.

Open for discussion:
- module name
- option for fifo buffering on either end
- lg_size_p parameter name

### DIFF
--- a/bsg_dataflow/bsg_fifo_store_forward.v
+++ b/bsg_dataflow/bsg_fifo_store_forward.v
@@ -1,0 +1,160 @@
+/**
+ *  bsg_fifo_store_forward.v
+ *
+ *  Paul Gao        04/2020
+ *
+ *  This module is designed for wormhole-like (multi-flits) traffic
+ *  It first stores (len_i + 1) number of data flits inside data fifo,
+ *  then bursts all flits out in one batch.
+ *
+ */
+
+module bsg_fifo_store_forward
+
+ #(parameter width_p     = "inv"
+  ,parameter len_width_p = "inv"
+  ,parameter max_els_p   = 2**len_width_p
+  )
+
+  (input clk_i
+  ,input reset_i
+
+  // Input side
+  ,input                      v_i
+  //
+  // len_i is logically associated with first-flit of data_i, and that the 
+  // value must be held constant until first-flit is removed from data_i.
+  //
+  // len_i should be equal to (total_num_of_flits - 1)
+  ,input  [len_width_p-1:0]   len_i
+  ,input  [width_p-1:0]       data_i
+  ,output                     ready_o
+
+  // Output side
+  ,output                     v_o
+  ,output [width_p-1:0]       data_o
+  ,input                      yumi_i
+  );
+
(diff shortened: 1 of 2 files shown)


taylor-bsg:
Some ideas:

Probably good to rope in Paul/Tommy, since they has experience in this and
also has written some code that does this and we could see if this is
sufficient for refactoring the existing code

Probably suggest putting the async after bsg_noc.

Is this for wormhole router or for the bsg_noc?  When going from a slow
clock domain to a faster one in a wormhole routed network, we often to add
a store and forward node, because the one a wormhole channel is started, it
blocks all other usage until it is closed.

dpetrisko:
> Is this for wormhole router or for the bsg_noc?

I've used this mostly on the proc side of the wormhole_router.  That's an interesting point for store and forward.   Maybe the interface is to have the clock ratio as a parameter.  Based on the clock ratio, could have a piso or configure the fifo depth as appropriate

> Probably suggest putting the async after bsg_noc.
Agree

taylor-bsg:
Paul has a store and forward implementation that we use for off-chip
routing..

dpetrisko:
Great, we can use that as well.  Seems like it would belong in bsg_async or bsg_noc rather than a bsg_design somewhere.  Unless I'm missing something

taylor-bsg:
Yes I think so. This is the preferred path for basejump_stl: recognize a
need in the wild that corresponds to a general purpose hardware concept,
define a clean interface, and then commit an implementation by refactoring
it out of the wild..

gaozihou:
Thanks prof and Dan! 

I guess I have store-forward implementation on FPGA designs: (1) Lattice FPGA for PCIe transmission (2) Xilinx FPGA for I2C transmission, if I remember correctly I did not implement them in a separate module. I will add one to basejump_stl ASAP.

For last tapeout we don't have store forward thing in off-chip link... We don't have one in bsg_link because bsg_link supports any kind of traffic (not only wormhole format). Should we add one in coming tapeout? (we could add one between channel-tunnel and wormhole-router)

dpetrisko:
Depending on what frequencies we end up at for the upcoming tapeout, it may be useful for core->mem crossing as well

gaozihou:
Added a store forward fifo (which has similar interface as normal bsg_fifo). Should be easy to use inside async_noc_link.

One concern of adding store-forward is latency. For example, BP has some 10-flits wormhole packets, 9-cycles of extra delay is added with store-forward, which may not be ideal for latency-sensitive links (like mem->core)?

gaozihou:
Thanks prof! Seems that in this way ctrl_fifo can be completely removed, which saves area and power. Will update the store-forward-fifo with flow_counter!

taylor-bsg:
Thanks Paul!!

dpetrisko:
This good to go?

dpetrisko:
Closing for staleness

PR #272, opened 2020-04-08
FIFO-to-DPI and DPI-to-FIFO interfaces.

drichmond:
More documentation and testbenches to follow. This currently works in my own (rudimentary) testbench in verilator, but I'm working to make that testbench more thorough.


bsg_test/bsg_nonsynth_dpi_to_fifo.v
```diff
@@ -0,0 +1,219 @@
...
+// width_p is the bit-width of the FIFO interface. Must be a power of
+//   2 and divisible by 8, i.e. a ctype.
+// 
+// debug_p is the intial value to set on debug_o and to control debug
+//   messages. The debug() DPI function can be used to control
+//   messages at runtime, but this allows it to be set in the initial
+//   block, before any runtime functions can be called.
+module bsg_nonsynth_dpi_to_fifo
+  #(
+    parameter name_p = "bsg_nonsynth_dpi_to_fifo"
```
mrutt92, line 17: could mark the type as string?

bsg_test/bsg_nonsynth_dpi_to_fifo.v
```diff
@@ -0,0 +1,219 @@
...
+   // This bit tracks whether initialize has been called. If data is
+   // sent and recieved before init() is called, then this module will
+   // throw an assertion. TODO
+   bit    init_b = 0;
+   
+   // Print module parameters to the console and set the intial debug
+   // value
+   initial begin
+      if (!(width_p inside {8, 16, 32, 64})) begin
+         $fatal($sformatf("BSG ERROR: width_p: %x is not supported. Must be a power of 2 and divisible by 8", width_p));
```
mrutt92, line 40: not sure you need $sformatf here?

bsg_test/bsg_nonsynth_dpi_to_fifo.v
```diff
@@ -0,0 +1,219 @@
...
+   // this will cause $fatal to be called to indicate a protocol
+   // violation because v_o will return to 0 in the next cycle.
+   //
+   // tx() MUST be called after the positive edge of clk_i is
+   // evaluated.
+
+   // We set _v_o so that we can signal a read to the producer on the
+   // NEXT positive edge. _v_o flows to v_o on the negative edge of
+   // clk_i
+   reg _v_o;
```
mrutt92, line 123: I think the style guide dictates these be called `v_r` and `data_r`

drichmond: yeah, I wondered if there was a style guide comment for this. My default style is from RIFFA.

bsg_test/bsg_nonsynth_dpi_to_fifo.v
drichmond: You do, unfortunately, but it still doesn't work. $fatal with a string in verilator doesn't work. I just separated this into $display and $fatal.

bsg_test/bsg_nonsynth_dpi_to_fifo.v
drichmond: Will do

bsg_test/bsg_nonsynth_dpi_to_fifo.v
mrutt92: I've also seen people use `logic` over `reg`

bsg_test/bsg_nonsynth_dpi_to_fifo.v
dpetrisko: The separate statements work by being in an initial block.  This means that the code will need to first compile and then run to see the error.  Personally, I prefer elaboration system tasks for fatals, which means that the code will fail to compile and it gives users the reason for the compilation error.  For example:

https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_misc/bsg_dlatch.v

I think it needs to be sprintf for this.

bsg_test/bsg_nonsynth_dpi_to_fifo.v
drichmond: Switched to `logic`. Happy to switch names if you can point to a place in the style guide.

bsg_test/bsg_nonsynth_dpi_to_fifo.v
```diff
@@ -0,0 +1,220 @@
...
+   // Same as above, but with data_o.
+   reg [width_p-1:0] _data_o;
+   
+   
+   // We track the "last" data_o and last ready_i values to detect
+   // protocol violations. These are captured on the positive edge of
+   // the clock
+   reg [width_p-1:0] last_data_o;
```
mrutt92, line 131: Style check here too. Typically `_o` and `_i` are reserved for the inputs and outputs of the module. I think these should be `_r`.

drichmond: As with all things, I'm happy to change. 

I'd argue that there is a legitimate reason to violate the style guide here because I'm actually  capturing the last data_o value. But I can switch to _r

bsg_test/bsg_nonsynth_dpi_to_fifo.v
drichmond: @dpetrisko I can't get that to compile in Verilator?

bsg_test/bsg_nonsynth_dpi_to_fifo.v
mrutt92: https://docs.google.com/document/d/1xA5XUzBtz_D6aSyIBQUwFk_kSUdckrfxa2uzGjMgmCU/edit#heading=h.rhtqn8jwjs44

bsg_test/bsg_nonsynth_dpi_to_fifo.v
mrutt92: `last_data_o_r`?

mrutt92: or maybe just `data_o_r`?

drichmond: I like that

mrutt92:
Use standard suffixes: https://docs.google.com/document/d/1xA5XUzBtz_D6aSyIBQUwFk_kSUdckrfxa2uzGjMgmCU/edit#heading=h.rhtqn8jwjs44

bsg_test/bsg_nonsynth_dpi_to_fifo.v
drichmond: Fixed

bsg_test/bsg_nonsynth_dpi_to_fifo.v
drichmond: Fixed

bsg_test/bsg_nonsynth_dpi_to_fifo.v
drichmond: Fixed

drichmond:
> Use standard suffixes: https://docs.google.com/document/d/1xA5XUzBtz_D6aSyIBQUwFk_kSUdckrfxa2uzGjMgmCU/edit#heading=h.rhtqn8jwjs44

Fixed

drichmond:
I added a testbench, which runs three tests: 

* Performs 3 tests:
  * 10 Fill/Empty Iterations
  * 100 R/W iterations with nearly-empty FIFO
  * 100 R/W iterations with nearly-full FIFO


Thanks to @dpetrisko for being my sounding board on this. I think we've got the timing issues nailed, even though there's some *strange* verilog in here.

bsg_test/bsg_nonsynth_dpi_to_fifo.v
```diff
@@ -0,0 +1,227 @@
...
+// name_p is the name to print in this modules BSG DBGINFO messages
+// 
+// width_p is the bit-width of the FIFO interface. Must be a power of
+//   2 and divisible by 8, i.e. a ctype.
+// 
+// debug_p is the intial value to set on debug_o and to control debug
+//   messages. The debug() DPI function can be used to control
+//   messages at runtime, but this allows it to be set in the initial
+//   block, before any runtime functions can be called.
+module bsg_nonsynth_dpi_to_fifo
```
taylor-bsg, line 15: might be good to have a description of what the functions supported are, either in the header comments or in the module comments, so that user does not have to read the code to understand what it does

dpetrisko:
I guess the reason to do that would be if direct manipulation were faster than dpi in Verilator? (No idea if that’s true or not)

bsg_test/bsg_nonsynth_dpi_to_fifo.v
drichmond: Yeah, thanks for the suggestion. I realized this yesterday too and never got around to it, so I'll do it today!

drichmond:
Finished the C++ API and added comments everywhere, if anyone wants to provide more feedback

PR #273, opened 2020-04-10
Dramsim3 bandwidth2

tommydcjung:


mrutt92:
Is it possible to split this into two PRs? One with the microbenchmarks and the other with subcaches?

tommydcjung:
No, the testbench is setup for subcaches.

mrutt92:
How are you using these benchmarks to compare against your old code then? Is num_subcache_p=1 identical hardware to what you had before this PR?

mrutt92:
Does it at least have an identical interface? It seems like it should. Can you just set subcache_p=1 in your python scripts?

tommydcjung:
Sorry I don't have bandwidth to do the same work twice.

mrutt92:
... OK - can you please rebase this branch?

bsg_test/bsg_nonsynth_test_dram_channel.v
```diff
@@ -0,0 +1,60 @@
+/**
```
mrutt92, line 1: I think this file was removed in a previous PR.

taylor-bsg:
This looks ready to go except for the conflict?

bsg_cache/bsg_cache_to_test_dram.v
```diff
@@ -11,6 +11,7 @@
 module bsg_cache_to_test_dram 
   import bsg_cache_pkg::*;
   #(parameter num_cache_p="inv"
+    , parameter num_subcache_p=1 // number of caches that are grouped together for the first-level of round robin.
```
taylor-bsg, line 14: add some detailed header comments about what this module is intended to do, what features it has (e.g. subcache), and what policies it currently uses.

PR #274, opened 2020-04-15
RFC/WIP: Verilator Drop-In replacement for bsg_nonsynth_clock_gen

drichmond:
Depends on/extends #272 

One of the frustrating parts of Verilator is that it "breaks" how we normally build testbenches. Our traditional approach is to use the bsg_nonsynth_clock_gen module and specify the clock period as a parameter. In Verilator, this is not possible because the module uses an unsupported delay statement. It's also more challenging (though not impossible) to have multiple clock domains. 

What I've done is create a drop-in replacement that is backed by a C++ API, called bsg_nonsynth_clock_gen_dpi (name subject to change). The user doesn't need to know the difference, they just use a different module name with the same parameters and include some .cpp/.hpp files during compilation. (See testing/bsg_test/bsg_nonsynth_dpi/top.v if you don't believe me)

This drop-in replacement supports multiple clock generators and they can be embedded anywhere in the hierarchy.

The C++ API is callback based; When each clock-generator module is instantiated, it registers itself with the C++ object via an imported DPI function. The bsg_timekeeper class tracks the global time (no different than normal verilator) and uses a priority queue to track when the next clock generator toggles. To advance time, the users calls bsg_timekeeper::next() (This method name may be subject to change)

This does introduce some weirdness, but it has a benefit. The user of the DPI-FIFO interface does NOT know if there's multiple clocks. However, calling into the functions at the wrong point in the clock cycle can cause simulation to fail. Therefore, they DO need to know when they are in a valid "window" to call into the tx/rx functions. Therefore, I've added a DPI "is_window" (possibly subject to change) function to the FIFO-DPI modules that returns 1 if this is a valid time to call rx/tx. This way, the user can call bsg_timekeeper::next() in a while loop and is_window to check if data is available and doesn't need to worry about tracking positive/negative edges (huge benefit, small weirdness, IMO).

I'll be adding comments and tests over the next couple of days. Obviously this module needs a testbench by itself. I need to catch up on other git work that I've been ignoring (sorry @bornaehsani!) while figuring this module out. 

Also TBD: Does this work with threading in Verilator? I haven't even gotten Verilator to compile with threads so I haven't gotten a chance to answer this. 

Also TBD: Can multiple top-levels be instantiated? This would be super cool, since we could simulate multiple pods without the same RTL simulation. Not on my critical path though.

Anyway, one more small step toward HB-Verilator simulation.


bsg_test/bsg_nonsynth_fifo_to_dpi.v
```diff
@@ -0,0 +1,252 @@
...
+    ,parameter bit debug_p = 0
+    ) 
+   (
+    input clk_i
+    , input reset_i
+      
+    , input v_i
+    , input [width_p-1:0] data_i
+    , output reg yumi_o
+
```
drichmond, line 63: Needed to declare this as reg so that newer versions of Verilator wouldn't complain. Would logic be better?

bsg_test/bsg_nonsynth_dpi_to_fifo.v
```diff
@@ -0,0 +1,284 @@
...
+    parameter string name_p = "bsg_nonsynth_dpi_to_fifo"
+    ,parameter width_p = "inv"
+    ,parameter bit debug_p = 0
+    )
+   (
+    input clk_i
+    , input reset_i
+      
+    , output reg v_o
+    , output reg [width_p-1:0] data_o
```
drichmond, line 68: Needed to declare these as reg so that newer versions of Verilator wouldn't complain. Would logic be better?

dpetrisko: Logic is usually preferred in all cases.  It needs to be reg since you're assigning to it in an always_ff block, I believe VCS would have the same issue

dpetrisko:
> Can multiple top-levels be instantiated? This would be super cool, since we could simulate multiple pods without the same RTL simulation. Not on my critical path though.

Yes I've done this in the past

bsg_test/bsg_nonsynth_dpi_to_fifo.v
drichmond: Fixed

drichmond:
This is ready for review. I've added comments everywhere.

bsg_test/bsg_nonsynth_dpi_clock_gen.v
```diff
@@ -0,0 +1,62 @@
...
+// bsg_timekeeper class tracks the global time (no different than
+// normal verilator) and uses a priority queue to track when the next
+// clock generator toggles. To advance time, the users calls
+// bsg_timekeeper::next()
+//
+// This drop-in replacement supports multiple clock generators and
+// can be embedded anywhere in the hierarchy.
+module bsg_nonsynth_dpi_clock_gen
+/*verilator lint_off WIDTH*/
+  #(parameter longint cycle_time_p="inv"
```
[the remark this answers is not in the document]
drichmond: Just verilator being verilator if I recall. It didn't like the string "inv" because it's not technically correct. 

I tried to use longint to give some type safety, but I can switch to logic

testing/bsg_test/bsg_nonsynth_dpi/Makefile
```diff
@@ -1,36 +1,55 @@
 .DEFAULT_GOAL := test_fifo_dpi
 
 BASEJUMP_STL_DIR := $(abspath ../../../)
-VERILATOR_ROOT=/usr/share/verilator
 
+#VERILATOR=/home/drichmond/Research/repositories/git/verilator/bin/verilator
+#VERILATOR_ROOT=/home/drichmond/Research/repositories/git/verilator
```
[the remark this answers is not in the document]
drichmond: Yep

testing/bsg_test/bsg_nonsynth_dpi/Makefile
```diff
@@ -1,36 +1,55 @@
...
+VERILATOR_ROOT=/usr/share/verilator
```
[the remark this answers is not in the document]
drichmond: Yep

PR #277, opened 2020-04-20
set default rom_data_width_p

tommydcjung:


bsg_tag/bsg_tag_trace_replay.v
```diff
@@ -35,7 +35,8 @@
 module bsg_tag_trace_replay
 
    #( parameter rom_addr_width_p    = -1
-    , parameter rom_data_width_p    = -1
+    , parameter rom_data_width_p    =
+      (4+num_masters_p+`BSG_SAFE_CLOG2(num_clients_p)+1+$clog2(max_payload_width_p+1)+max_payload_width_p)
```
dpetrisko, line 39: Is it okay to use num_masters_p, max_payload_width_p before declaration?

stdavids: $clog2(n+1) can be replaced with `BSG_WIDTH(n) which is also defined in bsg_defines.v

Issue #209, opened 2019-12-15
bsg_trace_replay + bsg_trace_rom methodology does not scale well with Verilator


dpetrisko:
This is an issue with a few possible solutions, feedback welcome.

In Verilator, compiling a modestly sized bsg_trace_rom (1000) rows takes a minute.  A larger rom (10000+) takes nearly half an hour.  The same roms are compiled almost instantly in VCS.

A few spitball ideas:
1) Try to patch verilator to compile roms faster, since they should be just a giant array
2) Use readmemh to populate contents of ROM
3) Add a bsg_nonsynth_rom which dynamically queries its entries from disk via DPI, or similar

taylor-bsg:
Have we filed a bug against Verilator?

M

dpetrisko:
Not yet, since this is a performance bug, we'll need some reproducible experiments to back up the claim.  (someone else feel free to file)

PR #278, opened 2020-04-24
DPI ROM

drichmond:
Reading from the BSG Rom is tedious in Verilator -- it requires setting inputs and calling eval(), or putting it on a bus and then waiting multiple clock cycles. 

I took a shortcut. I created a DPI module that acts like a ROM, and can be instantiated in Verilog, but doesn't require any eval() ing.

It's silly, but useful.

### DIFF
--- a/bsg_test/bsg_nonsynth_dpi_rom.hpp
+++ b/bsg_test/bsg_nonsynth_dpi_rom.hpp
@@ -1,0 +1,63 @@
+#ifndef __BSG_NONSYNTH_DPI_ROM_HPP
+#define __BSG_NONSYNTH_DPI_ROM_HPP
+#include <svdpi.h>
+#include <verilated.h>
+#include <bsg_nonsynth_dpi.hpp>
+
+extern "C" {
+        // DPI Export function: Get the value at an index in the
+        // rom. Calls $fatal if an invalid index is accessed.
+        extern svBitVecVal bsg_dpi_rom_get(int);
+}
+
+namespace bsg_nonsynth_dpi{
+
+        // dpi_rom is the C++ wrapper class for the
+        // bsg_nonsynth_dpi_rom verilog module.
+        // 
+        // Template Parameters:
+        // 
+        //   typename T: The C-type that this rom contains. While
+        //     verilog may represent arbitrary bit-widths, C cannot,
+        //     so this rom is restricted to C-types.
+        //
+        //   unsigned int N: Number of Elements in the ROM.
+        //
+        // Superclasses are used to check template parameters at run
+        // time against the corresponding verilog module.
+        template <typename T, unsigned int N>
+        class dpi_rom : public dpi_base, 
+                    public dpi_width<T>, 
+                    public dpi_nels<N>{
+        public:
+
+                // Constructor: Creates a dpi_rom object. 
+                // 
+                // Arguments: 
+                //   char *hierarchy: The string describing
+                //     the instantiation of this the corresponding
+                //     bsg_nonsynth_dpi_rom verilog module in the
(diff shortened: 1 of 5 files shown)


dpetrisko:
Oh that's awesome, that should help with #209

bsg_test/bsg_nonsynth_dpi_rom.hpp
```diff
@@ -0,0 +1,63 @@
...
+}
+
+namespace bsg_nonsynth_dpi{
+
+        // dpi_rom is the C++ wrapper class for the
+        // bsg_nonsynth_dpi_rom verilog module.
+        // 
+        // Template Parameters:
+        // 
+        //   typename T: The C-type that this rom contains. While
```
mrutt92, line 20: You might consider using https://en.cppreference.com/w/cpp/language/static_assert with https://en.cppreference.com/w/cpp/types#Type_traits_.28since_C.2B.2B11.29 to throw a compile time error if T is an invalid type.

mrutt92: Can T be a struct?

drichmond: Hmmmnh. It could be. But I haven't tested that. There's a different DPI type for structs.

drichmond: Do you want me to use static assert?

mrutt92: It was a suggestion.

mrutt92: check this out: https://en.cppreference.com/w/cpp/types/is_class

drichmond: This would force a static assert (which is nice) but I'd still need to call the dpi_width constructor to force a runtime assert when there's a mismatch between verilog and C

So I'm leaning toward meh, though admitting its usefulness.

Also, #friday.

taylor-bsg:
Just so it’s clear what was the issue you ran into, Dustin? Was it
Verilator runtime, like Dan, or was there was a reason you did not want to
just compile it in the Verilog world?

drichmond:
It’s in the verilog world - I wanted a way to access it in the c world.

In bladerunner we use a ROM to capture the parameters of the currently configured machine. This serves as a double check against misconfiguration, and a way to avoid using macros when building the C runtime (also, error prone). In bladerunner the ROM is on the AXI bus and configuration can be read from the FPGA.

Because we’ve built the runtime this (good) way, we need a similar method to read configuration parameters from the hardware, using the new DPI interface I’m building.

The issue is that the BSG rom we currently use in bladerunner is a case statement in an always block (or similar). To get data out there are two use cases: add it to the manycore (or similar bus, like axi) and spend multiple cycles reading it, or, add it to the DPI interface and call eval() to propagate inputs to outputs each time you read the ROM (and something stranger in VCS since propagating inputs requires delay statements). 

Since the API I’m building in Verilator is a direct interface to the manycore, the ROM would need to be directly attached to the network as an endpoint, or snoop the interface for manycore packets. I didn’t like either solution because I didn’t see a long-term use for a ROM endpoint, and snooping/removing packets seems Hacky. 

I didn’t like the second option because calling eval() at an inopportune time can lead to strange behavior. Better to call eval once per clock edge. 

This module provides a DPI interface for reading configuration parameters that have been put in a ROM, in verilog. It should work in Verilator and VCS, and doesn’t require multiple clock cycles, or eval, to read.

It also took about half a day to code.

PR #279, opened 2020-04-24
DPI FIFO Updates

drichmond:
These are updates for the DPI FIFO Interface that make it easier to use. Necessary as I've learned and used more of the DPI. 

* Standardized naming of exported DPI functions with bsg_dpi* prefix
* Renamed FIFO DPI modules to bsg_nonsynth_dpi_to_fifo and dpi_from_fifo
* Simplified constructor interface
* Added an ERRNO API that allows the DPI to signal different types of recoverable/non-recoverable errors to software (e.g. !valid, !ready, !window)

Merge after #274


bsg_test/bsg_nonsynth_dpi.hpp
```diff
@@ -1,13 +1,40 @@
...
 // The bsg_nonsynth_dpi namespace contains classes that are useful for
 // wrapping DPI interfaces.
 namespace bsg_nonsynth_dpi{
 
+        // Convert a svBitVec val to an integral C-type.
+        //
+        // NOTE: This might be hyper verilator specific. I'm relying
+        // on svBitVec being contiguously allocated to transfer data.
+        template<typename T>
+        inline void svToIntegral(const svBitVecVal *s, T &d){
```
dpetrisko, line 25:
```
    --pins-bv <bits>            Specify types for top level ports
    --pins-sc-uint              Specify types for top level ports
    --pins-sc-biguint           Specify types for top level ports
    --pins-uint8                Specify types for top level ports
```

So if your data width is greater than 64 bits (I believe) it automatically gives you a bitvec.  However, less will be integral types for performance.  These verilator flags will modify that default behavior.

The tradeoff of making all ports bitvecval are performance (though I've never measured the impact

drichmond: I'm hesitant to rely on something that might be verilator specific, but I agree that there might be performance value here

Sound reasonable?

dpetrisko: I guess my point is more that you need to use the bitvecval flag when compiling or this will fail for small fifos

drichmond: Aha. This is an interesting point. It will fail for say, 7-bit fifos?

dpetrisko: There’s some Verilator progression, I believe it uses the smallest integral type until it can’t fit anymore (it’s been a while since I’ve dealt with this). But the easiest thing would be to force the behavior with a flag

drichmond: OK, to resolve this

Dan and I were mutually miscomprehending each other. Dan thought this method was for interacting directly with the generated C++ code (e.g. copying out the value of a port) whereas I intend this function to be used to copy a "returned" value from the DPI interface into a c-type.

@dpetrisko Do you agree? Can we resolve this?

PR #283, opened 2020-04-30
Change instance name to include the string DONT_TOUCH

stdavids:
In our cad flow, we currently search specifically for instances called BSG_BAL41MUX and set a don't touch attribute on those instances. I would like to generalize this such that any instance with *DONT_TOUCH* in the name gets the dont_touch attribute (similar to how we do things with *NO_CLOCK_GATE*.

### DIFF
--- a/hard/gf_14/README.md
+++ b/hard/gf_14/README.md
@@ -1,0 +1,17 @@
+# GF14 Hard Directory
+
+This directory contains hardened (process specific) variants of basejump_stl
+RTL modules. These modules should be swapped with the RTL modules of the same
+name during the chip implementation flow.
+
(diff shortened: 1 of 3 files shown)


stdavids:
Adding dont_touch to clock generator adt, cdt and fdt instance names as well.

I don't think we should change things pre-emptively but rather create PRs, currently these are the only things that we search for in the 14 flow and apply a dont_touch attribute.

dpetrisko:
Also, could we add a quick note on this in the basejump_stl/hard/README as to what magic names exist and what they're expected to do?

stdavids:
Good suggestion, added!

PR #285, opened 2020-05-01
add two_fifo for dma_pkt input

tommydcjung:


taylor-bsg:
Checkin comment should always answer the question: why?

tommydcjung:
In terms of physical design, valid-yumi interface on dma_pkt ends up having big round trip. So it's nice to have these FIFOs.

taylor-bsg:
Maybe put that comment in the code?

gaozihou:
I've verified the correctness of this modification.

Another reason for adding this fifo is to avoid one-cycle bubble between receiving dma_pkt_i and dma_data_i (only happens when attached directly to bsg_cache, because bsg_cache_dma engine has an output-two-fifo on dma_data_o but not on dma_pkt_o.) Maybe we could add comments into code?

gaozihou:
> I've verified the correctness of this modification.
> 
> Another reason for adding this fifo is to avoid one-cycle bubble between receiving dma_pkt_i and dma_data_i (only happens when attached directly to bsg_cache, because bsg_cache_dma engine has an output-two-fifo on dma_data_o but not on dma_pkt_o.) Maybe we could add comments into code?
> 
> Q: Do we need to update testbench here?
> 
> https://github.com/bespoke-silicon-group/basejump_stl/blob/768028fe19b352f476dcae3b9a1183a8e9ccd1bc/testing/bsg_cache/manycore_dram/testbench.v#L282

tommydcjung:
ready 2 go?

tommydcjung:
manycore_dram testbench doesn't work currently, we have better testbench elsewhere.

PR #289, opened 2020-05-10
Synchronous/Asynchronous BSG CAM

dpetrisko:
Adds synchronous content-addressable memory.

This module is more of a true cam than the module currently called bsg_cam_1r1w.v. Should we replace that with these components (minus the sync read register?) as well?


dpetrisko:
I added a parameterizable replacement policy module, so the interface has changed to a true 1r1w CAM at the outmost level. The semantics for clearing individual entries became murky (do we support clearing by tag or clearing by position), so I removed it from the interface for now. When we have a usecase, we can bring it back

dpetrisko:
> add w_set_not_clear_i back in

With clear meaning clear tag == wtag?

taylor-bsg:
w_set_nor_clear_i is the value that you want to write to the valid bit.

dpetrisko:
Right.  My question is that the old semantics were:

```
w_v_i = 4'b0001
set_not_clear = 0
w_tag_i = X
[a|b|c|d]

"Clear d"
```

Now that the replacement module is inside the CAM, the w_v_i is only 1 bit. So the new clear would be:

```
w_v_i = 1'b1
set_not_clear = 0
w_tag_i = d
[a|b|c|d]

"Clear d"
```

Does that make sense?

dpetrisko:
Okay fixed up the interfaces.  
- set_not_clear is back
- the replacement module can accept both reads and writes, up to the policy what to do with them
- Added unmanaged cams for clients that wish to control the replacement at a finer granularity

bsg_mem/bsg_cam_1r1w_tag_array.v
```diff
@@ -13,9 +13,12 @@ module bsg_cam_1r1w_tag_array
   (input                          clk_i
    , input                        reset_i
 
-   // one or zero-hot
-   , input [els_p-1:0]            w_v_i
-   // Tag to set on write
+   // Mutually exclusive set or clear
+   , input                        w_v_i
```
taylor-bsg: Sorry this was not the interface change I was proposing!  I was just proposing a one line change. This "raw" interface to the tag array should still be a vector of w_v_i bits. My point with w_set_not_clear_i is that you have two fields that are settlable for an entry inside the array: valid bit and tag, and that you need to supply the data for both. w_set_not_clear_i is the value you are writing to the valid bit!

It was pretty interesting to read your version here; effectively you are adding extra comparison bandwidth to this module to enable performing associative writes to the cam as well. But I think from the perspective of this module (which is about tag finding), we would actually represent that as an array of read ports -- find the tag for read you want to do, find the tag for the write you want to do.

taylor-bsg: Sorry after this clarification, faa9522 does not make as much sense. But in that commit I did like the work you did to make it clear which of the concurrent ports the input and output signals of a module were associated with. I think you just need a 1-2 line change for w_set_not_clear_i above, and then to handle the LRU update under concurrent reads and allocate, you can follow this mod to the allocator: " think the solution is to have v_read_i and v_alloc_i, and then to add a bsg_lru_pseudo_tree_decode of way_lo, and a bsg_mux_bitwise after the current bsg_mux_bitwise. Then the concurrency in the interface can be handled. If it ends up on the critical path for BP, we can add a parameter to selectively remove the extra logic."

dpetrisko:
Changed the replacement interface to entirely one-hot, added simultaneous alloc/read support to the replacement unit, some general code cleanup

bsg_mem/bsg_cam_1r1w.v
```diff
@@ -1,140 +1,104 @@
...
+   #(.els_p(els_p)
+     ,.scheme_p(repl_scheme_p)
+     )
+   replacement
+    (.clk_i(clk_i)
+     ,.reset_i(reset_i)
+
+     ,.read_v_i(tag_r_match_lo & {els_p{r_v_i}})
+
+     ,.alloc_v_i(tag_w_match_lo & {els_p{w_v_i & w_set_not_clear_i}})
```
taylor-bsg, line 67: alloc_v_i(w_v_i)

bsg_mem/bsg_cam_1r1w.v
```diff
@@ -1,140 +1,104 @@
...
+  logic [els_p-1:0] tag_w_v_li;
+  bsg_cam_1r1w_tag_array
+   #(.width_p(tag_width_p)
+     ,.els_p(els_p)
+     )
+   cam_tag_array
+    (.clk_i(clk_i)
+     ,.reset_i(reset_i)
+
+     ,.w_v_i(tag_w_v_li)
```
taylor-bsg, line 43: wire w_v_li = repl_way_lo | { els_p { w_nuke_i } }
.w_v_i(w_v_li)

bsg_mem/bsg_cam_1r1w.v
```diff
@@ -1,140 +1,104 @@
...
+     ,.w_set_not_clear_i(w_set_not_clear_i)
```
taylor-bsg, line 44: .w_set_not_clear_i(w_v_i & ~ w_nuke_i)

bsg_mem/bsg_cam_1r1w.v
```diff
@@ -1,140 +1,104 @@
...
+     ,.w_tag_i(w_tag_i)
+     ,.w_match_o(tag_w_match_lo)
```
taylor-bsg, line 46: remove w_match_o

bsg_mem/bsg_cam_1r1w.v
```diff
@@ -1,140 +1,104 @@
...
+     ,.r_v_i(mem_r_v_li)
+     ,.r_data_o(r_data_o)
+     );
+
+  logic [els_p-1:0] set_way_lo, clear_way_lo;
+
+  assign set_way_lo   = repl_way_lo    & {els_p{w_v_i &  w_set_not_clear_i}};
+  assign clear_way_lo = tag_w_match_lo & {els_p{w_v_i & ~w_set_not_clear_i}};
+  assign read_v_li    = tag_r_match_lo & {els_p{r_v_i}};
+  assign alloc_v_li   = set_way_lo;
```
taylor-bsg, line 95: alloc_v_li unused

bsg_mem/bsg_cam_1r1w.v
```diff
@@ -1,140 +1,104 @@
...
+  logic [els_p-1:0] mem_w_v_li, mem_r_v_li;
+  bsg_mem_1r1w_one_hot
+   #(.width_p(data_width_p)
+     ,.els_p(els_p)
+     )
+   one_hot_mem
+    (.w_clk_i(clk_i)
+     ,.w_reset_i(reset_i)
+
+     ,.w_v_i(mem_w_v_li)
```
taylor-bsg, line 83: .w_v_i(w_v_li)

bsg_mem/bsg_cam_1r1w.v
```diff
@@ -1,140 +1,104 @@
...
+     ,.w_data_i(w_data_i)
+
+     ,.r_v_i(mem_r_v_li)
```
taylor-bsg, line 86: .r_v_i(tag_r_match_lo)

(no r_v_i needed since it is masked by tags now)

bsg_mem/bsg_cam_1r1w.v
```diff
@@ -1,140 +1,104 @@
...
-* whether multiple entries with the same key are allowed, the match_array is
-* sent to either a one_hot_encoder or a priority_encoder to generate
-* the address of the match location. Also by setting the find_empty_entry_p
-* parameter, if there is an empty entry in the memory, empty_v_o is raised and
-* the first empty entry's address is put on empty_addr_o. This can help other
-* modules interacting with CAM, to easily find an address to write to.
-*
-*/
+ * Asynchronous read 1r1w content addressable memory module.
+ * Each entry has a tag and a data associated with it, and can be independently cleared and set
```
taylor-bsg, line 3: * Read searches the array for any data with tag_i
* Write allocates a new entry, replacing an existing entry with replacement scheme repl_scheme_p
* Write with w_nuke_i flag invalidates the cam

taylor-bsg:
Click on view changes so that you can see the recommendations inline with the code, it will make a lot more sense. =)

bsg_mem/bsg_cam_1r1w.v
dpetrisko: Hm, I get that we won't use it for the bsg_cam_1r1w, but I think it's most flexible to have the option.  It's needed for a 1r1w replacement policy that has multiple matches, or selective invalidations.  If we just leave the port disconnected here, the output comparison will be zero-propagated

bsg_mem/bsg_cam_1r1w.v
dpetrisko: This will unnecessarily write on tag clears

bsg_mem/bsg_cam_1r1w_sync_unmanaged.v
```diff
@@ -0,0 +1,75 @@
...
+   // one or zero-hot
+   , input [els_p-1:0]               w_v_i
+   , input                           w_set_not_clear_i
+   // Tag/data to set on write
+   , input [tag_width_p-1:0]         w_tag_i
+   , input [data_width_p-1:0]        w_data_i
+   // Metadata useful for an external replacement policy
+   // Whether write tag has a match
+   , output [els_p-1:0]              w_match_o
+   // Whether there's an empty entry in the tag array
```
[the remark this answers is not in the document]
dpetrisko: Same concern as above -- this prevents multiple match CAM use-cases

bsg_mem/bsg_cam_1r1w_unmanaged.v
```diff
@@ -0,0 +1,80 @@
...
+   cam_tag_array
+    (.clk_i(clk_i)
+     ,.reset_i(reset_i)
+
+     ,.w_v_i(w_v_i)
+     ,.w_set_not_clear_i(w_set_not_clear_i)
+     ,.w_tag_i(w_tag_i)
+     ,.w_match_o(w_match_o)
+     ,.w_empty_o(w_empty_o)
+
```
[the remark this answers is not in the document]
dpetrisko: I assume w_tag_i is a mistake, we definitely need that in our tag array :)

dpetrisko:
All comments addressed except for combining tag_w_v_li and mem_w_v_li, which would result in unnecessary writes to the data mems on invalidations

bsg_mem/bsg_cam_1r1w.v
taylor-bsg: kk! It was bouncing around in the back of my head as a nice security feature, but I can't think of a concrete use case.

bsg_mem/bsg_cam_1r1w_replacement.v
```diff
@@ -0,0 +1,153 @@
...
+   , input                     reset_i
+
+   // Synchronous update (i.e. indicate that an entry was read)
+   , input [els_p-1:0]         read_v_i
+
+   // May use combination of internal state and empty vector
+   //   to determine replacement
+   // Synchronous update (i.e. indicate that an entry was allocated)
+   , input                     alloc_v_i
+   , input [els_p-1:0]         empty_i
```
taylor-bsg, line 28: shall we call this alloc_empty_i for good measure?

taylor-bsg:
Neat! Kind of a cool module to write -- can't say I've ever seen any code where one hot was so pervasive; I guess that's because usually other people are designing the arrays for us!

The variant where you can both read and write using the tag is kind of awesome (and makes me feel like I am implementing python dictionaries in hardware!), but i think we should wait until we see a real use case before we define that interface. It will be relatively easy to build via the tag array primitive; we would just pass in a parameter which is the number of read ports; and then we do the initial async read for each write to detect if it's in the array in the first place.

bsg_mem/bsg_cam_1r1w.v
```diff
@@ -1,140 +1,95 @@
...
+ * - Read searches the array for any data with r_tag_i
+ * - Write allocates a new entry, replacing an existing entry with replacement
+ *     scheme repl_scheme_p
+ * - Write with w_nuke_i flag invalidates the cam
+ */
```
taylor-bsg, line 8: * - allocation happens in parallel with the read, so it is possible in an LRU scheme, for the currently-being-read item to be selected for over-writing
* - there are no concerns about simultaneous read and writing
* - it is generally discouraged to have multiple identical tags in the array, but if there are, then the data returned on read is the OR of the data. If you actually find that functionality useful, let us know. =)

dpetrisko:
Thanks for all the feedback! Ready to merge whenever

PR #290, opened 2020-05-20
[DEFERRED] Adding sync 1rw bit reshaper

dpetrisko:
This module is used to fold a tall, skinny ram into a wider, shallow ram. This is useful for something like a BHT which is logically very narrow.

Questions:
- Should we expose this at the hard level or the RTL level? If we put this module in the RTL, people will get better synthesis results, but we have to decide widths at the RTL instead of the hard level
- Should we embed the fat RAM in the module? If we need a reshaper module per type of RAM, then there's not really a benefit to having the RAM outside of the module, like bsg_fifo_bypass
- Should we have a byte module similar to this one, or a parameter in this module which does byte masking?
- Should we have a 1r1w no mask variant, which interacts with a bit or byte mask RAM?


dpetrisko:
Another question: Do we want the FSM to reset the memory in this logic?  It's technically orthogonal, but it fits cleanest in here because it's dependent on the fat ram dimensions, not the skinny ram dimensions

dpetrisko:
Okay, changed this to a 1RW reshaper which gets rid of the same_read_write requirement.  Ready for review

dpetrisko:
Closing because we never found a good use-case

PR #291, opened 2020-05-22
[DEFERRED] Adding 1r1w LIFO

dpetrisko:
Useful for e.g. return address stacks


dpetrisko:
Implemented less generically at https://github.com/black-parrot/black-parrot/blob/master/bp_fe/src/v/bp_fe_ras.sv

taylor-bsg:
@dpetrisko is there anything wrong with this module? You have a test, it is seems complete, etc?

dpetrisko:
Nothing wrong per-se but the use-case (RAS) requires checkpoint/restore for performance and can't be built out of this component

PR #292, opened 2020-05-26
Modeling bidirectional wires with delay

hema0730:


bsg_test/bsg_nonsynth_triwire.v
```diff
@@ -0,0 +1,17 @@
+module bsg_nonsynth_triwire #
+  (parameter width_p="inv"
+  ,parameter real transport_delay_p = 0.0)
```
stdavids, line 3: Is there a major benefit to using a real value rather than just an int? It seems like a simulator might struggle if the time scale has the same unit and precision and this was not a whole number.

hema0730: I think you are right and no particular reason to make it a real number. Just wanna have a fine granularity when the unit is big while precision is small. I will simulate it with the same unit and precision and see if a real number resolves to the closest integer.

hema0730: From VCS simulation, when the unit is the same as precision, the delay value will round to the closest integer if it is a real number. For example 1.2 will resolve to 1 while 1.8 will resolve to 2.

taylor-bsg:
Good to go?

stdavids:
Yep, seems good to me

PR #296, opened 2020-06-03
Fixing AMO width bug

### LINKED ISSUE #295: bsg_cache atomics hardcoded for 32-bit data_width
Mask is entire word width (will break for 64-bit width):

https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_cache/bsg_cache.v#L634

AMOs should be sigext:

https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_cache/bsg_cache_decode.v#L34

dpetrisko:
fixes #295


dpetrisko:
Superceded by #298

PR #298, opened 2020-06-04
Adding more atomics to bsg_cache

### LINKED ISSUE #295: bsg_cache atomics hardcoded for 32-bit data_width

dpetrisko:
~Depedent on #296~
Fixes #295 

This PR adds support for the rest of the RISCV fetch-and-op atomics to bsg_cache.  It defines 4 support levels for atomics (inspired by the RISC-V AMO PMAs) which users may select to trade off functionality for PPA.

image

The added AMOs are:
- AMOADD
- AMOXOR
- AMOAND
- AMOOR
- AMOMIN
- AMOMAX
- AMOMINU
- AMOMAXU

in addition to amoswap and amoor, already implemented.

This PR also enables double-word atomics.

Before merging, we should test on a 64-bit cache testbench, but the extra ops are functional for word width


tommydcjung:
I think this is cool and all that... But there is a 3-day moratorium adding new HW. Also, I think there is a simpler way to implement. Let me suggest a better way..

taylor-bsg:
Very cool Dan.  I think Dan is looking at this from the RISC-V compliance
perspective of BP. From the HammerBlade side, we received a request for
this functionality (ie beyond AMO swap) from Cornell but we had a concern
it might add a critical path. I like how Dan offered it as a parameterize
feature to support the old mode. I guess Tommy can advise on PPA impact and
preferred implementation...


M

dpetrisko:
I’m all ears for a more efficient way to implement, but this seems like a minimal hardware solution. 

Indeed, we (BP) need it to be competitive with OpenPiton+Ariane. And we (Basejump) need it to be competitive with other L2-esque caches which implement full atomic sets.

bsg_cache/bsg_cache_pkg.v
```diff
@@ -6,6 +6,13 @@
 
 package bsg_cache_pkg;
 
+  typedef enum logic [1:0] {
+    e_amo_none        = 2'b00
+    ,e_amo_swap       = 2'b01
+    ,e_amo_logical    = 2'b10
+    ,e_amo_arithmetic = 2'b11
+  } amo_support_level_e;
```
tommydcjung, line 14: Instead, create one-hot enums for each amo class (swap, or, xor, and, add, min/max). This way, the interface is more generic and it still supports riscv classification.

bsg_cache/bsg_cache_decode.v
```diff
@@ -31,7 +35,16 @@ module bsg_cache_decode
   assign decode_o.sigext_op = (opcode_i == LB)
     | (opcode_i == LH)
     | (opcode_i == LW)
-    | (opcode_i == LD);
+    | (opcode_i == LD)
+    | (opcode_i == AMOSWAP_W)
```
tommydcjung, line 39: The cache can also support 128-bit data width or higher. Replace this with decode_o.atomic_op

testing/bsg_cache/regression_v2/test_atomic1.py
```diff
@@ -13,13 +13,29 @@ def generate(self):
         index = random.randint(0,1)
         block_offset = random.randint(0,self.block_size_in_words_p-1)
         taddr = self.get_addr(tag,index,block_offset)
-        op = random.randint(0,2)
+        op = random.randint(0,10)
```
tommydcjung, line 16: Keep the existing tests unchanged. And add new tests to test the new features.

bsg_cache/bsg_cache.v
```diff
@@ -592,9 +594,63 @@ module bsg_cache
...
+    logic [slice_width_lp-1:0] slice_data;
+
+    // AMO computation
+    // AMOs are only supported for words and double words
+    if ((i == 2'b10) || (i == 2'b11)) begin
+      logic [slice_width_lp-1:0] atomic_reg_data, atomic_mem_data;
+
+      always_comb begin
+        atomic_reg_data = data_v_r[0+:slice_width_lp];
+        atomic_mem_data = snoop_or_ld_data[0+:slice_width_lp];
```
tommydcjung, line 608: This doesn't seem right... What if atomic op is done on the second 32-bit of 64-bit word? I think atomic operator should be applied with ld_data_final_lo.

tommydcjung:
Create a new testing directory (regression_64) to test all the functionalities for 64-bit config before request for another review.

bsg_cache/bsg_cache_pkg.v
dpetrisko: To clarify, you mean to have 6 entries in an enum, and then pass as parameter something like: atomic_support_level_p = e_amo_swap | e_amo_or | e_amo_xor | e_amo_and?

bsg_cache/bsg_cache_decode.v
dpetrisko: Yep, good call

bsg_cache/bsg_cache.v
```diff
@@ -592,9 +594,63 @@ module bsg_cache
...
 
-    assign sbuf_data_in_mux_li[i] = {(data_width_p/(8*(2**i))){data_v_r[0+:(8*(2**i))]}};
```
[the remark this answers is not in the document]
dpetrisko: Which byte select mux do you mean?  This block is modifying the data input to the sbuf data mux

bsg_cache/bsg_cache_pkg.v
dpetrisko: If we do this, how do you extract the support vector?

```
To set support:

  typedef enum logic [9:0] {
    e_amo_swap        = 10'b0000000001
    ,e_amo_add        = 10'b0000000010
    ,e_amo_xor        = 10'b0000000100
    ,e_amo_and        = 10'b0000001000
    ,e_amo_or         = 10'b0000010000
    ,e_amo_min        = 10'b0000100000
    ,e_amo_max        = 10'b0001000000
    ,e_amo_minu       = 10'b0010000000
    ,e_amo_maxu       = 10'b0100000000

    // Reserved
    ,e_amo_lrsc       = 10'b1000000000
  } amo_support_e;

  localparam amo_support_level_none_lp       = '0;
  localparam amo_support_level_swap_lp       = amo_support_level_none_lp
    | e_amo_swap;
  localparam amo_support_level_logical_lp    = amo_support_level_swap_lp
    | e_amo_xor | e_amo_and | e_amo_or;
  localparam amo_support_level_arithmetic_lp = amo_support_level_logical_lp
    | e_amo_add | e_amo_min | e_amo_max | e_amo_minu | e_amo_maxu;
```

```
To get support I can think of:
|(amo_support_p & e_amo_swap) & decode_v_r.amoswap_op
or
(amo_support_p & e_amo_swap) && decode_v_r.amoswap_op
```

bsg_cache/bsg_cache.v
dpetrisko: Ah yeah, you're right.  I think using ld_data_final_li[i] should be better for critical path, though

dpetrisko:
> Create a new testing directory (regression_64) to test all the functionalities for 64-bit config before request for another review.

I've removed the 64-bit operands from the decoder.  This PR is now only adding the rest of the AMO.Ws to bsg_cache.  Tests are added for the new functionality.  64-bit ops are no more or less tested than they were before.  We can add 64-bit support in a separate PR

bsg_cache/bsg_cache.v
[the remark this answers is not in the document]
dpetrisko: Applying the atomic operation on data_v_r does not work for non-logical operations, such as AMO_MAX. For these ops, the operation must be done on the subword, because they may have a different result for each subword

dpetrisko:
Rebased onto #299 to get the 64-bit testbench (I think this PR now supercedes it).  Added 64-bit test for the added atomics.

Ready for re-review @tommydcjung

dpetrisko:
What else needs to be done for this to merge?

dpetrisko:
ping

dpetrisko:
ping

bsg_cache/bsg_cache_pkg.v
```diff
@@ -6,6 +6,29 @@
 
 package bsg_cache_pkg;
 
+  typedef enum logic [9:0] {
+    e_amo_swap        = 10'b0000000001
```
tommydcjung, line 10: Please change to e_cache_amo_*, so it does not conflict with other enums.

tommydcjung:
Does testing/bsg_cache/regression pass?

I'm getting the following error.
############################################################################
### bsg_fsb_node_trace_replay RECEIVE unmatched (testbench.trace_replay) 
###    
### FAIL (trace mismatch) = 0000000000000001000
###              expected = 0000000000000002234

###              diff     = 0000000000000003234

############################################################################

bsg_cache/bsg_cache.v
```diff
@@ -593,28 +596,74 @@ module bsg_cache
...
+        end
+        else if ((amo_support_p & e_amo_minu) && decode_v_r.amominu_op) begin
+          slice_data = (atomic_reg_data < atomic_mem_data) ? atomic_reg_data : atomic_mem_data;
+        end
+        else if ((amo_support_p & e_amo_maxu) && decode_v_r.amomaxu_op) begin
+          slice_data = (atomic_reg_data > atomic_mem_data) ? atomic_reg_data : atomic_mem_data;
+        end
+        else begin
+          // Normal data replication
+          slice_data = data_v_r[0+:slice_width_lp];
```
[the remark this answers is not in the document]
taylor-bsg: @dpetrisko The implication of the if statement is that the decode_v_r signals are not one hot. But we know they are.  This will end up with inefficient synthesis because it is essentially nesting muxes.  Better would be to write the code for an ALU, where you use the decode signals to control a one-hot mux the selects among operators. You can mask amo_support_p against a vector of the decode bits to detect violation.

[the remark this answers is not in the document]
dpetrisko: @tommydcjung yep, will add

@taylor-bsg Ah, this used to be "unique if", which I believe solves that problem

bsg_cache/bsg_cache_pkg.v
dpetrisko: fixed

dpetrisko:
> Does testing/bsg_cache/regression pass?

Fixed, now passes.  Problem was that I added the AMOs to the trace, but didn't update the support level in the testbench.

However, I tried running bsg_cache/regression_64 and it hung on test_random2.  Interestingly, it also hangs on test_random2 for the original commit the 64-bit testbench was pushed with: 824f4ffd6cfd3c7369aca26480aef526fcba42f2 .  Can you take a look @tommydcjung ?

dpetrisko:
Fixed up the testbenches! regression/regression_v2 and regression_64 w/ new 64-bit atomics all pass

bsg_cache/bsg_cache.v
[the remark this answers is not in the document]
tommydcjung: Doesn't this synthesize one adder for 32-bit and another for 64-bit? same for other operators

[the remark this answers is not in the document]
taylor-bsg: The issue with unique if is that we will have to do a 10-bit compare for each of the bit patterns.  So why are we one-hot encoding then? A cleaner separation of datapath and control by having an explicit will also allow as an extension for more controlled operator merging, as Tommy mentioned. (Aside, not required for this pull request, i'm not sure if it's possible, but it would be pretty cool to generalize into a bsg_alu that would configurably let you remove functions, maybe using bsg_concentrate, and would share logic between operators in a smart way.)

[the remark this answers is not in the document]
taylor-bsg: Following up on Tommy's point about 32-bit and 64-bit both generating hardware, can we instead just have it so that if we are doing 32-bit on a 64-bit AMO-capable v-cache, we shift the bits up to the high word, do the operation, and then shift back? Then we can merge the ALU.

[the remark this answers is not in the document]
dpetrisko: Hm, it's probably possible to merge 32/64 bit ops.  It's a PPA tradeoff, though, as it introduces latency in the 32-64 bit conversion, (shift and mask), where there is currently only op+mux latency.  Since we're concerned about critical path more than area in bsg_cache, I suggest further analysing the tradeoff in a separate change.  Note that the ops being limited to 32/64 bits is already an efficiency boost over the current implementation.

As an aside, the parameterization becomes messier with this combination, as we have to consider caches of size [1-31], [32-63], [64+] as separate cases.  The current parameterization by slice is more elegant

tommydcjung:
vcache_atomic_datapath

You can have a mux select between word and double and feed that into atomic unit

dpetrisko:
Thanks for the feedback @tommydcjung and @taylor-bsg! I've extract the atomic ALU with the bit-shifting trick. I also added a few assertions that users are not trying to use atomic operations on a cache data width <= the atomic op width

bsg_cache/bsg_cache.v
```diff
@@ -592,29 +596,136 @@ module bsg_cache
...
+      .data_i({amo64_in, amo32_in})
+      ,.sel_i(decode_v_r.data_size_op[0])
+      ,.data_o(atomic_reg_data)
+    );
+
+    bsg_mux #(
+      .width_p(64)
+      ,.els_p(2)
+    ) atomic_mem_data_mux (
+      .data_i(ld_data_final_li[3:2])
```
tommydcjung, line 629: Doesn't ld_data_final_li[2] also need to shift?

bsg_cache/bsg_cache_pkg.v
```diff
@@ -6,6 +6,29 @@
 
 package bsg_cache_pkg;
 
+  typedef enum logic [9:0] {
```
tommydcjung, line 9: Can we make this not one-hot, and (amo_support_p & e_cache_amo_swap) can be amo_support_p[e_cache_amo_swap]? and have  amo_support_e in decode_o.

bsg_cache/bsg_cache.v
tommydcjung: Why is the shifting needed in the first place? so make 32-bit path shorter? But what's the point of doing that when there is already longer path?

bsg_cache/bsg_cache_pkg.v
dpetrisko: How would we encode e_cache_amo_or | e_cache_amo_swap in that scheme?

tommydcjung: (1 << e_cache_amo_or) | (1 << e_cache_amo_swap)

dpetrisko: Oh, got it.  Sure, can change that

bsg_cache/bsg_cache.v
dpetrisko: > Doesn't ld_data_final_li[2] also need to shift?

Ope, you're right with the shift.  The tests don't check word-aligned atomics for regression_64.  I'll add that, fail, and then fix the bug.

> Why is the shifting needed in the first place? so make 32-bit path shorter? But what's the point of doing that when there is already longer path?

I think you're right.  Since the load_data is already shifted down and we don't need to sign extend the result, it'll be more efficient to mask the high bits of the ops instead of shifting.  I'll update that as well

bsg_cache/bsg_cache_pkg.v
dpetrisko: fixed

bsg_cache/bsg_cache.v
dpetrisko: Thinking about it more, doing it without the shift would put a zero extension on the ld_data_final_li->wbuf path (to handle amo_minu/maxu), meaning that there ends up being two muxes on the path -- the same as doing two shifts.  The shift approach handles the sign/unsigned ops better

tommydcjung: You can decode amomin/max as sigext_op.

taylor-bsg: Yeah it’s for the signed/unsigned ops.

taylor-bsg: Yeah it’s for the signed/unsigned ops.

bsg_cache/bsg_cache_decode.v
```diff
@@ -62,11 +67,23 @@ module bsg_cache_decode
...
-
+  assign decode_o.amoswap_op = (opcode_i == AMOSWAP_W) | (opcode_i == AMOSWAP_D);
+  assign decode_o.amoadd_op = (opcode_i == AMOADD_W) | (opcode_i == AMOADD_D);
+  assign decode_o.amoxor_op = (opcode_i == AMOXOR_W) | (opcode_i == AMOXOR_D);
+  assign decode_o.amoand_op = (opcode_i == AMOAND_W) | (opcode_i == AMOAND_D);
+  assign decode_o.amoor_op = (opcode_i == AMOOR_W) | (opcode_i == AMOOR_D);
+  assign decode_o.amomin_op = (opcode_i == AMOMIN_W) | (opcode_i == AMOMIN_D);
+  assign decode_o.amomax_op = (opcode_i == AMOMAX_W) | (opcode_i == AMOMAX_D);
+  assign decode_o.amominu_op = (opcode_i == AMOMINU_W) | (opcode_i == AMOMINU_D);
+  assign decode_o.amomaxu_op = (opcode_i == AMOMAXU_W) | (opcode_i == AMOMAXU_D);
```
tommydcjung, line 78: I was trying to say that instead of having these one-hot signals, you can just have amo_support_e.

dpetrisko: Sure, can change that.

bsg_cache/bsg_cache.v
dpetrisko: > You can decode amomin/max as sigext_op.

The sigext refers to the load data, which for atomics is always.  The comparison for minu is itself unsigned

bsg_cache/bsg_cache_decode.v
dpetrisko: Fixed!

dpetrisko:
Feedback addressed

bsg_cache/bsg_cache_decode.v
```diff
@@ -62,11 +67,21 @@ module bsg_cache_decode
...
+      AMOSWAP_W, AMOSWAP_D: decode_o.amo_subop = e_cache_amo_swap;
+      AMOADD_W, AMOADD_D: decode_o.amo_subop = e_cache_amo_add;
+      AMOXOR_W, AMOXOR_D: decode_o.amo_subop = e_cache_amo_xor;
+      AMOAND_W, AMOAND_D: decode_o.amo_subop = e_cache_amo_and;
+      AMOOR_W, AMOOR_D: decode_o.amo_subop = e_cache_amo_or;
+      AMOMIN_W, AMOMIN_D: decode_o.amo_subop = e_cache_amo_min;
+      AMOMAX_W, AMOMAX_D: decode_o.amo_subop = e_cache_amo_max;
+      AMOMINU_W, AMOMINU_D: decode_o.amo_subop = e_cache_amo_minu;
+      AMOMAXU_W, AMOMAXU_D: decode_o.amo_subop = e_cache_amo_maxu;
+      default: decode_o.atomic_op = 1'b0;
```
tommydcjung, line 83: Please add default value for amo_subop (as e_cache_amo_swap) so that it doesn't create inferred latch.

bsg_cache/bsg_cache.v
```diff
@@ -592,29 +597,124 @@ module bsg_cache
...
+      {1'b1, e_cache_amo_add }: atomic_alu_result = atomic_reg_data + atomic_mem_data;
+      {1'b1, e_cache_amo_min }: atomic_alu_result =
+          ($signed(atomic_reg_data) < $signed(atomic_mem_data)) ? atomic_reg_data : atomic_mem_data;
+      {1'b1, e_cache_amo_max }: atomic_alu_result =
+          ($signed(atomic_reg_data) > $signed(atomic_mem_data)) ? atomic_reg_data : atomic_mem_data;
+      {1'b1, e_cache_amo_minu}: atomic_alu_result =
+          (atomic_reg_data < atomic_mem_data) ? atomic_reg_data : atomic_mem_data;
+      {1'b1, e_cache_amo_maxu}: atomic_alu_result =
+          (atomic_reg_data > atomic_mem_data) ? atomic_reg_data : atomic_mem_data;
+      default: atomic_alu_result = '0;
```
tommydcjung, line 658: Is this '0 used anywhere? If not, just set it to atomic_reg_data?

bsg_cache/bsg_cache_decode.v
dpetrisko: oh good call

bsg_cache/bsg_cache.v
dpetrisko: nope, sounds good

bsg_cache/bsg_cache.v
```diff
@@ -655,7 +654,8 @@ module bsg_cache
           (atomic_reg_data < atomic_mem_data) ? atomic_reg_data : atomic_mem_data;
       {1'b1, e_cache_amo_maxu}: atomic_alu_result =
           (atomic_reg_data > atomic_mem_data) ? atomic_reg_data : atomic_mem_data;
-      default: atomic_alu_result = '0;
+      // default is amo_swap
+      default: atomic_alu_result = atomic_reg_data;
```
tommydcjung, line 658: I think we should still keep {1'b1, e_cache_amo_swap}... What's the expected result if the user plugs in unsupported atomic op? It should have zero side effect (does not cause cache miss), maybe it should also return zero instead?

bsg_cache/bsg_cache_decode.v
```diff
@@ -69,6 +69,7 @@ module bsg_cache_decode
   // atomic extension
   always_comb begin
     decode_o.atomic_op = 1'b1;
+    decode_o.amo_subop = e_cache_amo_swap;
```
tommydcjung, line 72: I would prefer having this as part of default statement, but this is okay..
default: begin
decode_o.atomic_op = 1'b0;
decode_o.amo_subop = e_cache_amo_swap;
end

dpetrisko: sure that's fine

bsg_cache/bsg_cache.v
dpetrisko: Hm, my opinion on this is that doing unsupported things in hardware results in undefined behavior (invariant violated -> assertion fired) to minimize area (get rid of an additional mux entry, reduce control logic).  

Reverted to having default zero for the return result.  Since this is a one-hot mux, there shouldn't be any overhead.

Avoiding a cache miss is more tricky and I don't think worth it.  We would have to detect the violation and modify the control logic, adding complexity just for an invariant violation.  And in the end, the user wouldn't be able to tell the difference between using an illegal op and simply getting back a legal zero

bsg_cache/bsg_cache_decode.v
dpetrisko: fixed

bsg_cache/bsg_cache.v
```diff
@@ -592,29 +597,124 @@ module bsg_cache
...
+  end
+  else if (data_width_p >= 32) begin : atomic_32
+    assign atomic_reg_data = data_v_r[0+:32];
+    assign atomic_mem_data = ld_data_final_li[2];
+  end
+
+  // Atomic ALU
+  always_comb begin
+    unique casez({amo_support_p[decode_v_r.amo_subop], decode_v_r.amo_subop})
+      {1'b1, e_cache_amo_swap}: atomic_alu_result = atomic_reg_data;
```
tommydcjung, line 645: I'm skeptical whether this approach will actually not synthesize the unsupported ops. The safer way would be
e_cache_amo_swap: if (amo_support_p[decode_v_r.amo_subop]) atomic_alu_result = atomic_reg_data; else atomic_alu_result = '0;

dpetrisko: I've used this in the past with DC and it has synthesized properly. Any insights @taylor-bsg ?

dpetrisko: Just tried it with the gf14 flow, does not generate extra logic

bsg_cache/bsg_cache.v
```diff
@@ -592,29 +597,124 @@ module bsg_cache
...
+  logic [`BSG_MIN(data_width_p, 64)-1:0] atomic_alu_result;
+  // Final atomic data for store buffer
+  logic [`BSG_MIN(data_width_p, 64)-1:0] atomic_result;
+
+  // Shift data to high bits for operations less than 64-bits
+  // This allows us to share the arithmetic operators for 32/64 bit atomics
+  if (data_width_p >= 64) begin : atomic_64
+    wire [63:0] amo32_reg_in = data_v_r[0+:32] << 32;
+    wire [63:0] amo64_reg_in = data_v_r[0+:64];
+    bsg_mux #(
```
taylor-bsg, line 617: I would just use ? : 

There is some implication with using a bsg_mux that we expect this to turn into a mux, but I would be upset to see if the synthesis tool really did that.

bsg_cache/bsg_cache.v
```diff
@@ -592,29 +597,124 @@ module bsg_cache
...
+      .width_p(64)
+      ,.els_p(2)
+    ) amo_reg_shift_mux (
+      .data_i({amo64_reg_in, amo32_reg_in})
+      ,.sel_i(decode_v_r.data_size_op[0])
+      ,.data_o(atomic_reg_data)
+    );
+
+    wire [63:0] amo32_mem_in = ld_data_final_li[2] << 32;
```
taylor-bsg, line 626: Same

bsg_cache/bsg_cache.v
taylor-bsg: A. What the
B. Cool!
C. Really it works?
D. Add a comment saying it was tested and does not synthesize the extraneous operators in DC version X.Y

bsg_cache/bsg_cache.v
taylor-bsg: 1’b0, ??????: BSG_UNDEFINED_IN_SIM

bsg_cache/bsg_cache_decode.v
```diff
@@ -31,7 +35,8 @@ module bsg_cache_decode
   assign decode_o.sigext_op = (opcode_i == LB)
     | (opcode_i == LH)
     | (opcode_i == LW)
-    | (opcode_i == LD);
+    | (opcode_i == LD)
+    | decode_o.atomic_op;
```
taylor-bsg, line 39: Be pure

bsg_cache/bsg_cache_pkg.v
```diff
@@ -6,6 +6,35 @@
 
 package bsg_cache_pkg;
 
+  typedef enum logic [3:0] {
```
taylor-bsg, line 9: A comment here that the intent is for the low four bits of the OPCODE to match this enum exactly to synthesize decoding, both in this file and in the decode file.

bsg_cache/bsg_cache.v
```diff
@@ -592,29 +597,112 @@ module bsg_cache
...
+          (atomic_reg_data > atomic_mem_data) ? atomic_reg_data : atomic_mem_data;
+      default: atomic_alu_result = '0;
+    endcase
+  end
+
+  // Shift data from high bits for operations less than 64-bits
+  if (data_width_p >= 64) begin : fi
+    wire [63:0] amo32_out = atomic_alu_result >> 32;
+    wire [63:0] amo64_out = atomic_alu_result;
+    bsg_mux #(
```
taylor-bsg, line 654: Fix

bsg_cache/bsg_cache.v
```diff
@@ -592,29 +597,107 @@ module bsg_cache
...
   for (genvar i = 0; i < data_sel_mux_els_lp; i++) begin: sbuf_in_sel
+    localparam slice_width_lp = (8*(2**i));
 
-    assign sbuf_data_in_mux_li[i] = {(data_width_p/(8*(2**i))){data_v_r[0+:(8*(2**i))]}};
+    logic [slice_width_lp-1:0] slice_data;
 
+    // AMO computation
+    // AMOs are only supported for words and double words
+    if ((i == 2'b10) || (i == 2'b11)) begin: atomic_in_sel
+      bsg_mux #(
```
tommydcjung, line 670: also use "? :" here?

dpetrisko: fixed!

dpetrisko:
Good to go

taylor-bsg:
Thank you all!

PR #303, opened 2020-06-17
Fixed two verilator errors for DramSim3

drichmond:
* A variable had both blocking and non-blocking assignments
* A return value was ignored

### DIFF
--- a/bsg_test/bsg_nonsynth_dramsim3.v
+++ b/bsg_test/bsg_nonsynth_dramsim3.v
@@ -73,9 +73,10 @@ module bsg_nonsynth_dramsim3
   
   import "DPI-C" context function 
     void    bsg_dramsim3_exit();
+     bit    init;
 
   initial begin
-    bsg_dramsim3_init(num_channels_p, data_width_p, size_in_bits_p, config_p);
+    init = bsg_dramsim3_init(num_channels_p, data_width_p, size_in_bits_p, config_p);
   end
 
   // memory addr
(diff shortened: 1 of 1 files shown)


bsg_test/bsg_nonsynth_dramsim3.v
```diff
@@ -152,13 +153,13 @@ module bsg_nonsynth_dramsim3
 
       // getting read/write done
       for (integer i = 0; i < num_channels_p; i++) begin
-        read_done[i] = bsg_dramsim3_get_read_done(i);
-        if (read_done[i])
-          read_done_addr[i] = bsg_dramsim3_get_read_done_addr(i);
+        read_done[i] <= bsg_dramsim3_get_read_done(i);
+        if (bsg_dramsim3_get_read_done(i))
```
drichmond, line 157: Not sure if there was a reason this was a blocking assignment outside of this always statement.

drichmond:
@tommydcjung, what do you mean? I asked in slack and you didn't respond. Can you be more explicit?

drichmond:
This passes the testbenches in dramsim3_bandwidth

drichmond:
And the regression tests in replicant

drichmond:
I've looked at one of the waveforms, and I don't see any differences. It's hard, because the patterns are random (I think?) but the clock edges appear to be correct.

is it possible to diff the traces for one of these and verify correctness? These pass in bladerunner with VCS and Verilator

PR #304, opened 2020-06-18
Switch to BSG External version of DRAMSim3

drichmond:
Contains our changes

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,3 @@
 [submodule "imports/ramulator"]
 	path = imports/ramulator
 	url = https://github.com/bespoke-silicon-group/ramulator
-[submodule "DRAMSim3"]
-	path = imports/DRAMSim3
-	url = git@github.com:bespoke-silicon-group/bsg_dramsim3
-        update = none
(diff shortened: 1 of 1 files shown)


drichmond:
This passes the dramsim3_bandwidth and dramsim3_bandwidth2 tests in testing, passes regression in replicant in Verilator and VCS with timing_v0_8_4.

tommydcjung:
Shouldn't you be creating a separate branch in this fork , instead of updating the master?

drichmond:
Is that what you want me to do? Happy to do it

tommydcjung:
It will be better if there are any new changes coming from the original.

drichmond:
Good point. When you say original, you mean the UMD repository, right?

drichmond:
Done! And I reverted master to its original state

tommydcjung:
I don't think the new branch should contain this commit, which basically resets all files.
https://github.com/bsg-external/DRAMsim3/commit/4d755992be83bd8e6f67d0b7ca8db92a3aae761c

drichmond:
How on earth...

drichmond:
I'm not seeing that. How are you seeing that?

PR #305, opened 2020-06-22
set max input to 32

tommydcjung:
It might not be a good setting for synthesis, but still can be needed for testbench/simulation.


tommydcjung:
I guess it affects reading the file and elaborating, but generate if-else shouldn't be much of problem?

drichmond:
I honestly don't know. Do you want me to test compile times in VCS and Verilator?

dpetrisko:
I think it depends if Verilator uses #ifdef or if else for generate statements. Interesting question

taylor-bsg:
I think Dan was saying that compile times for case statements can be out of
control for Verilator.

Presumably Tommy can do a quick test.

M

dpetrisko:
Yeah that's #209.  The problem there is each line of the case statement is actually used, so it ends up being a super large C function.  Presumably, parameters are optimized better, because unused parameter paths are "dead code" in the model.  But a quick experiment could verify

tommydcjung:
As far as I know, this has no impact on DC.

PR #307, opened 2020-06-25
Fixing critical path in bsg_encode_one_hot

dpetrisko:
The recursive nature of bsg_encode_one_hot leads to huge delays on the v_o signal for this module.  For a 16-entry priority encoder, I've seen 10-15 FO4 on the v_o port. I've squashed this path in the past by ignoring the v_o port and doing an OR on the bits outside of the module, but it makes sense for this fix to live inside the module.

My theory as to why the QoR is so bad is due to a lack of flattening.  Otherwise, I would think an OR reduce should synthesize fine, recursive or not

### DIFF
--- a/bsg_misc/bsg_id_pool.v
+++ b/bsg_misc/bsg_id_pool.v
@@ -50,6 +50,7 @@ module bsg_id_pool
   ) pe0 (
     .i(~allocated_r | dealloc_decode)
     ,.o(one_hot_out)
+    ,.v_o()
   );
 
   bsg_encode_one_hot #(
--- a/bsg_misc/bsg_id_pool.v
+++ b/bsg_misc/bsg_id_pool.v
@@ -44,13 +44,15 @@ module bsg_id_pool
   logic alloc_v_lo;
   logic [els_p-1:0] one_hot_out;
 
+   // We use this v_o instead of the v_o of bsg_encode_one_hot
+   //   because it has better critical path
   bsg_priority_encode_one_hot_out #(
     .width_p(els_p)
     ,.lo_to_hi_p(1)
   ) pe0 (
     .i(~allocated_r | dealloc_decode)
     ,.o(one_hot_out)
-    ,.v_o()
+    ,.v_o(alloc_v_lo)
   );
 
   bsg_encode_one_hot #(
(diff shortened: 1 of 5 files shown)


stdavids:
Oh nevermind, you need `vs` I missed the address reformation.

dpetrisko:
>Have you tested your theory by ungrouping this module to see if the path is any better?
No, will do

> Did you test to make sure that this fixed the path while not ungrouping?
Yes, it does

proftaylor:
Can you show a timIng report with the critical path? v_o is recursively
constructed into a tree so this should not have much depth. (But, we are
not using complementary logic and we are not using wide fan in gates, and
the code does not adjust the construction of the tree according to the
input arrival times of the signals.)


M

taylor-bsg:
This will generate a lot of excess logic, because it computes v_o at every level.  But per discussion with Dan and Scott, the root underlying problem is that the recursive implementation leads to excessive modularization. Instead I suggest reimplementing the align clause with a double nested loop. 

Regarding the path for v_o more generally, I suggest punching out v_o from bsg_priority_encode_one_hot_out -- it is either the low or high bit of the scan that is compute, depending on the setting of lo to hi.

dpetrisko:
Adding prefixed version of bsg_encode_one_hot! I'll try it out and see if the critical paths are better

bsg_misc/bsg_id_pool.v
```diff
@@ -50,6 +50,7 @@ module bsg_id_pool
   ) pe0 (
     .i(~allocated_r | dealloc_decode)
     ,.o(one_hot_out)
+    ,.v_o()
   );
```
taylor-bsg, line 54: .v_o(alloc_v_lo)

dpetrisko: fixed!

taylor-bsg:
Okay looks good. The bsg_encode_one_hot is a little sub-optimal because it does a OR reduce for the v bit at each level.  Maybe the synthesizer would figure out it is unnecessary, or maybe it won't. The use of the always_comb for loop is pretty interesting, but makes visibility a little difficult. I tried coding it an alternative way via generate statements writing into a "tableau" that tracks, which I think ends up a bit cleaner. I will submit it in a little while.

dpetrisko:
I think the v reduction is necessary? Because we use it to compute the prefix_addr at each level, it’s not just for v_o

taylor-bsg:
Yeah, but you can compute it just using the previous v bits in the tree. If
you look at the debug printout in the pull request I did, you can see how
the v tree works.

M

PR #308, opened 2020-06-26
Added DPI-accessible cycle counter

drichmond:
On DPI-based stacks it is useful to have a cycle counter that can be read via DPI. This module and C++ class provide that functionality.

For example, in the Manycore Verilator simulation we need this to understand how long a program takes to execute.


drichmond:
This works on the dpi-verilator platform in bsg_replicant.

dpetrisko:
LGTM!

PR #314, opened 2020-06-30
credit interface

tommydcjung:


taylor-bsg:
Pass in credits as array; don't add one if it's credit interface.

PR #315, opened 2020-06-30
Removed verilator requirement from DPI classes

drichmond:


dpetrisko:
Is there / should there be a corresponding change to add this check to a Verilator+dpi finish call?

drichmond:
Oh, that's a good idea. I can put something in the dpi-verilator platform in replicant.

PR #316, opened 2020-07-01
bsg_fifo_1r1w_small bsg_two_fifo for unhardened

tommydcjung:


tommydcjung:
bsg_two_fifo is slightly better in terms of area.

dpetrisko:
Maybe also sub bsg_one_fifo?

taylor-bsg:
Nice Tommy! Thanks!

I vaguely remember optimizing critical paths in bsg_two_fifo as well.

taylor-bsg:
I will approve and merge. Tommy, it would be nice if we can incorporate Dan's suggestion. I did confirm that bsg_fifo_1r1w_small does seem to work for the case of 1 element, so we might as well slide in the optimized code from bsg_one_fifo. This use of specialization is one of the core features in BaseJump STL.

PR #317, opened 2020-07-01
Allowing readmemh or readmemb in test rom

dpetrisko:
title


drichmond:
Rad. 

Have you tested that this works? I can run it in Verilator in a bit

dpetrisko:
Testing it now. It compiles though, which is half the battle :P

drichmond:
More than half, IMO

dpetrisko:
Works in both VCS and Verilator

dpetrisko:
good to go?

PR #318, opened 2020-07-01
Crossbar refactor

tommydcjung:


bsg_noc/bsg_router_crossbar_o_by_i.v
```diff
@@ -0,0 +1,132 @@
+/**
```
taylor-bsg, line 1: comment that describes the usage of the module, especially the source routing.

bsg_noc/bsg_router_crossbar_o_by_i.v
```diff
@@ -0,0 +1,132 @@
...
+    , input [i_els_p-1:0] valid_i
+    , input [i_els_p-1:0][i_width_p-1:0] data_i // lower bits = dest id.
+    , output [i_els_p-1:0] credit_ready_and_o // this can be either credits or ready_and_i on inputs based on i_use_credits_p
+
+    // crossbar output 
+    , output [o_els_p-1:0] valid_o
+    , output [o_els_p-1:0][o_width_lp-1:0] data_o
+    , input [o_els_p-1:0] ready_and_i
+  );
+
```
taylor-bsg, line 34: for now I would add an initial begin assert verifying that o_els_p != 1, and stating the semantics are TBD based on an actual use case, especially when dropping packets, since technically no routing bits are required for a 1-output router.

bsg_noc/bsg_router_crossbar_o_by_i.v
```diff
@@ -0,0 +1,139 @@
...
+
+  // parameter checking
+  initial begin
+    assert(o_els_p > 1) else $error("o_els_p needs to be greater than 1.");
+  end
+
```
taylor-bsg, line 41: // for now we leave this case unhandled
// awaiting an actual use case so we can
// determine whether the code is cleaner
// with 0-bit or 1-bit source routing.

bsg_noc/bsg_router_crossbar_o_by_i.v
```diff
@@ -0,0 +1,139 @@
+/**
+ *    bsg_router_crossbar_o_by_i.v
+ *
+ *    This module connects N inputs to M outputs with a crossbar network.
+ */
```
taylor-bsg, line 5: The drop header parameter can be optionally used to combine multiple crossbars into a network and implement source routing.

PR #321, opened 2020-07-09
Added error/failure case for when DPI scope not found

drichmond:


drichmond:
Relevant to [#347](https://github.com/bespoke-silicon-group/bsg_manycore/pull/347). Provides a better user warning when the DPI scope is not found.

dpetrisko:
LGTM!

PR #322, opened 2020-07-10
Created DPI GPIO module

drichmond:
We use this to enable/disable tracers from C/C++ in replicant.

### DIFF
--- a/bsg_test/bsg_nonsynth_dpi_gpio.v
+++ b/bsg_test/bsg_nonsynth_dpi_gpio.v
@@ -22,6 +22,8 @@
 module bsg_nonsynth_dpi_gpio
    #(parameter int width_p = -1
      ,parameter bit [width_p-1:0] init_o_p = '0
+     ,parameter bit use_input_p = '0
+     ,parameter bit use_output_p = '0
      ,parameter bit debug_p = '0
      )
    (output bit [width_p-1:0] gpio_o
@@ -32,6 +34,9 @@ module bsg_nonsynth_dpi_gpio
 
    if(width_p <= 0)
      $fatal(1, "BSG ERROR (%M): width_p must be greater than 0");
+
+   if(~use_input_p & ~use_output_p)
+     $fatal(1, "BSG ERROR (%M): GPIO must be configured to use input, output, or both");
 
    // Print module parameters to the console and set the intial debug
    // value. We use init_b to track whether the module has been
@@ -45,6 +50,8 @@ module bsg_nonsynth_dpi_gpio
       $display("BSG INFO:     Instantiation: %M");
       $display("BSG INFO:     width_p:       %d", width_p);
       $display("BSG INFO:     init_o_p:      %b", init_o_p);
+      $display("BSG INFO:     use_input_p:   %b", use_input_p);
+      $display("BSG INFO:     use_output_p:  %b", use_output_p);
       $display("BSG INFO:     debug_p:       %d", debug_p);
 
    end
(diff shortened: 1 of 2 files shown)


bsg_test/bsg_nonsynth_dpi_gpio.v
```diff
@@ -0,0 +1,133 @@
...
+//     a state change occurs (0->1 or 1->0) then module will print DEBUG
+//     ENABLED / DEBUG DISABLED. No messages are printed if a state
+//     change does not occur.
+//   int bsg_dpi_width(): returns the parameter width_p
+module bsg_nonsynth_dpi_gpio
+   #(parameter int width_p = -1
+     ,parameter bit [width_p-1:0] init_o_p = '0
+     ,parameter bit debug_p = '0
+     )
+   (output bit [width_p-1:0] gpio_o
```
dpetrisko, line 27: Are only 1 of these expected to be connected?  If so, use inout?

drichmond: I expect they could both be connected - and coming from an FPGA background (where inout isn't architecturally supported) I defaulted to in/out

But your comment is good. What do you envision?

drichmond:
> Looks cool. Could you describe an example use case? I'm curious how you actually hook this up somewhere

Yeah, for sure. This module is intended to be used to set/read "pins" using DPI. It is similar to the functionality that is provided by the AWS-FPGA interface `sv_set_virtual_dip_switch`, that we use to do clock switching in VCS.

We use it in [#606](https://github.com/bespoke-silicon-group/bsg_replicant/pull/606/) to control trace generation. Two output pins enable and disable trace/log generation. 

Verilog: 

https://github.com/bespoke-silicon-group/bsg_replicant/blob/b7046915c1bc51532b68776ca71b186bac977ee2/libraries/platforms/aws-fpga/hardware/cl_manycore.sv#L1331

C++: https://github.com/bespoke-silicon-group/bsg_replicant/blob/b7046915c1bc51532b68776ca71b186bac977ee2/libraries/features/tracer/simulation/bsg_manycore_tracer.cpp#L53-L58

bsg_test/bsg_nonsynth_dpi_gpio.v
drichmond: See page 9 here: https://www.xilinx.com/support/documentation/ip_documentation/axi_gpio/v2_0/pg144-axi-gpio.pdf

That was my mental model when creating this module. Obviously I left out `t` because that's not necessary in simulation.

dpetrisko: Oh, cool.  Maybe we can add a parameter to specify single or dual channel, flagging an error if the opposite configuration is used?

For example, configure as a strict input, and then error if we use get?

drichmond: Sounds reasonable. How about `use_input_p` and `use_output_p`, and then fail if the relevant method is used but the parameter isn't?

A single parameter leads to an enum (input only, output only, dual) and that has to be documented. Numerous parameters leads to increase (and an unused fourth state) but I feel like that is more intuitive.

dpetrisko: Yep, that makes sense to me! Probably an error message for the ~input/~output state

drichmond: Addressed! See the code for details

PR #323, opened 2020-07-12
Adding write mask to the DRAMSim3 module

farzamgl:
This PR adds a byte masked version of the nonsynth DRAMSim3 module.

### DIFF
--- a/bsg_test/bsg_nonsynth_dramsim3.v
+++ b/bsg_test/bsg_nonsynth_dramsim3.v
@@ -10,6 +10,7 @@ module bsg_nonsynth_dramsim3
     , parameter num_columns_p="inv"
     , parameter address_mapping_p="inv"
     , parameter size_in_bits_p=0
+    , parameter masked_p=0
     , parameter debug_p=0
     , parameter init_mem_p=0 // zero out values in memory at the beginning
     , parameter string config_p="inv"
@@ -30,6 +31,7 @@ module bsg_nonsynth_dramsim3
 
     , input [num_channels_p-1:0] data_v_i
     , input [num_channels_p-1:0][data_width_p-1:0] data_i
+    , input [num_channels_p-1:0][data_mask_width_lp-1:0] mask_i
     , output logic [num_channels_p-1:0] data_yumi_o 
 
     , output logic [num_channels_p-1:0] data_v_o
@@ -197,9 +199,10 @@ module bsg_nonsynth_dramsim3
   logic [num_channels_p-1:0] read_v_li;
   logic [num_channels_p-1:0][channel_addr_width_p-1:0] read_addr_li;
   logic [num_channels_p-1:0] write_v_li;
+  logic [num_channels_p-1:0][data_mask_width_lp-1:0] mask_li;
 
   for (genvar i = 0; i < num_channels_p; i++) begin: channels
-    bsg_nonsynth_mem_1r1w_sync_dma
+    bsg_nonsynth_mem_1r1w_sync_mask_write_byte_dma
       #(.width_p(data_width_p)
         ,.els_p((size_in_bits_p/num_channels_p)/data_width_p)
         ,.id_p(base_id_p+i)
@@ -207,19 +210,22 @@ module bsg_nonsynth_dramsim3
     channel
       (.clk_i(clk_i)
       ,.reset_i(reset_i)
-     
+
       ,.r_v_i(read_v_li[i])
       ,.r_addr_i(read_addr_li[i][channel_addr_width_p-1:byte_offset_width_lp])
 
-      ,.w_v_i(write_v_li[i])    
-      ,.w_addr_i(ch_addr_i[i][channel_addr_width_p-1:byte_offset_width_lp]) 
+      ,.w_v_i(write_v_li[i])
+      ,.w_addr_i(ch_addr_i[i][channel_addr_width_p-1:byte_offset_width_lp])
       ,.w_data_i(data_i[i])
+      ,.w_mask_i(mask_li[i])
 
       ,.data_o(data_o[i])
     );
 
     assign read_v_li[i] = read_done[i];
     assign read_addr_li[i] = read_done_ch_addr[i];
+
+    assign mask_li[i] = masked_p ? mask_i[i] : data_mask_width_lp'($signed(1));
   
     assign write_v_li[i] = data_v_i[i] & v_i[i] & write_not_read_i[i] & yumi_o[i];
     assign data_yumi_o[i] = data_v_i[i] & write_not_read_i[i] & yumi_o[i];
@@ -268,6 +274,8 @@ module bsg_nonsynth_dramsim3
     end
   end
 
+  assert property (@(posedge mask_i, negedge mask_i) masked_p)
+    else $warning("Ignoring mask_i because masked_p is not set");
 
   // final
   final begin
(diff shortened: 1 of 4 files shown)


farzamgl:
> I am little constrained in what I can see from the mobile app, but I would assume this module is 95% the same as the non-masked version? Is there any reason not to just merge them, either by making the other module be a wrapper around this one, or just by adding the mask bits to the original interface (assuming HBM2 has mask bits, I think this is preferred since it signals the availability of these bits.)

Yes, it's mostly the same module just the memory is swapped with a masked one. As for the second approach, I don't think DRAMSim supports write masks as far as I've looked at the code but maybe @tommydcjung or @mrutt92 can verify this.

mrutt92:
I agree with @taylor-bsg that the two modules should be consolidated. DRAMSim is only responsible for the timing of transactions - I would just assume the mask bits are available with `data_i`

drichmond:
I agree with the discussion above. I've confirmed that HBM supports byte-masking (though it does not if ECC is enabled?). 

Both Intel and Xilinx parts support byte masking.

@mrutt92 Can you verify what you mean when you say "I would just assume the mask bits are available with data_i". Do you mean that they're pins that are read at the same time as data_i?

mrutt92:
> @mrutt92 Can you verify what you mean when you say "I would just assume the mask bits are available with data_i". Do you mean that they're pins that are read at the same time as data_i?

That's what I mean. The JEDEC doc shows timing diagrams for write operations which include a signal called DATA. The footnote for these diagrams state that data includes DM (data mask).

From the doc, figures 45-50 (at least), there's a footnote.

`3. WDQS = WDQS[3:0]; DATA = DQ[127:0], DM[15:0], DBI[15:0].`

farzamgl:
Thank you Max!
So are we going to just add the mask bits to the original interface and have 1 module? If there's a special case where user never uses write masks, those bits can be tied to one.

drichmond:
Channeling my inner @dpetrisko 

Why not make the use of byte-masking a parameter? If the default is 0 (not used) then the module would be backward compatible.

farzamgl:
> Channeling my inner @dpetrisko
> 
> Why not make the use of byte-masking a parameter? If the default is 0 (not used) then the module would be backward compatible.

Well, it's a bit awkward because there's still gonna be an unused `mask_i` port which can be confusing to the user.

dpetrisko:
>Channeling my inner @dpetrisko 
:) 

In this case I support the wrapper, since it's more in line with modules like: https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_mem/bsg_nonsynth_mem_1r1w_sync_dma.v#L32

taylor-bsg:
Hi all,

This is a very interesting case, and lots of great opinions here from
everybody.

I think this case is a little different than with the rest of BaseJump STL,
since it is targeting an existing interface rather than trying to
synthesize new logic. So while there is a benefit to having a spectrum of
RAMs to attain new levels of specialization; I think in terms of the DRAM
interface, byte masking is part of the DRAM interface, so I think our DRAM
interface module should just have it, and we should avoid the wrapper
interface with the slightly simpler interface since it does not add much
value.

As Farzam said, in BaseJump STL, we do prefer to avoid "utility" modules
that have lots of extra inputs and outputs that are selected via a
parameter to determine whether they are used or not.  But I think in this
case, since DRAM is a standard interface, I think having a parameter that
signals "hey, I do not intend to use this masking feature" is reasonable
since it can be used by infrastructure potentially to make the
implementation more simulation efficient.


M

mrutt92:
Channeling my inner me - whatever you do please be sure to fix the file list or sv source in the projects that use the original module (bsg_replicant, bsg_manycore, etc.).

farzamgl:
Alright, I added the mask port and the parameter to the original module. The default value for the parameter is 0 so all the tests that do not use a mask still work as-is.

bsg_test/bsg_nonsynth_dramsim3.v
```diff
@@ -199,24 +201,47 @@ module bsg_nonsynth_dramsim3
...
+        (.clk_i(clk_i)
+        ,.reset_i(reset_i)
+
+        ,.r_v_i(read_v_li[i])
+        ,.r_addr_i(read_addr_li[i][channel_addr_width_p-1:byte_offset_width_lp])
+
+        ,.w_v_i(write_v_li[i])
+        ,.w_addr_i(ch_addr_i[i][channel_addr_width_p-1:byte_offset_width_lp])
+        ,.w_data_i(data_i[i])
+        ,.w_mask_i(mask_i[i])
```
dpetrisko, line 220: Might as well just use byte dma for both, since sync_dma is just a wrapper itself: https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_mem/bsg_nonsynth_mem_1r1w_sync_dma.v#L39

something like mask_li = mask_i | {{mask_width}~masked_p}}

dpetrisko: Also, an assertion if w_mask_i is not constant and the parameter is low?

farzamgl: Good point!

farzamgl: Can SystemVerilog assertions detect signal edges in an array? Or do I have to do the register->xor->or thing?

farzamgl: Yeah apparently it does support that.

bsg_test/bsg_nonsynth_dramsim3.v
```diff
@@ -10,6 +10,7 @@ module bsg_nonsynth_dramsim3
     , parameter num_columns_p="inv"
     , parameter address_mapping_p="inv"
     , parameter size_in_bits_p=0
+    , parameter masked_p=0
```
tommydcjung, line 13: Do we really need masked_p? the existing code will need to add mask_i port anyway, where {n{1'b1}} can be provided.

farzamgl: I don't have a strong opinion on this, but the good thing about this is that it's backward compatible so you won't need to change the code in manycore. On the other side, it's a bit weird to have an unused input port.

tommydcjung: Updating the existing codes will be a one-liner... I think it's worse to have a parameter that changes whether the port is used or not.

drichmond: I had the same question, and @taylor-bsg weighed in here: https://github.com/bespoke-silicon-group/basejump_stl/pull/323#issuecomment-657722026

tommydcjung: Does it make the implementation more simulation efficient? nope.

taylor-bsg: It's definitely an interesting corner case -- with the passage of time it will become more clear what the right approach is when faced with similar such situations.

bornaehsani:
Looks like this PR broke the DMA operations in the Hammerblade Manycore. CUDA regression tests using DMA operations pass with fast_n_fake machines, but are failing in timing_v0_8_4 machine after this merge. @mrutt92 Do you have any ideas?

mrutt92:
Is it only DMA that fails?

drichmond:
Yeah - I'm experimenting with this now. 

If I switch to bsg_nonsynth_mem_1r1w_sync_dma instead of bsg_nonsynth_mem_1r1w_sync_mask_write_byte_dma this works. 

Maybe the issue wasn't introduced by @farzamgl's change. Maybe it existed in the masked version before this change.

drichmond:
I'm learning the API here. If the two versions differ, it could just be an API mismatch, instead of a functional bug

dpetrisko:
Are we sure it’s the write mask change?  Possibly this is the culprit

https://github.com/bespoke-silicon-group/basejump_stl/pull/326

drichmond:
Yes, removing this PR fixes the issue. 

I diagnosed it to the use of $signed.

dpetrisko:
Oh yeah I see it. Should be ‘1

PR #324, opened 2020-07-16
crossbar router credit interface, add assertion about no FIFO space.

tommydcjung:


drichmond:
Hold up - i'm testing this to make sure the assertion would have found the bug we just discussed

bsg_noc/bsg_router_crossbar_o_by_i.v
```diff
@@ -82,6 +82,15 @@ module bsg_router_crossbar_o_by_i
         ,.data_i(fifo_yumi_li[i])
         ,.data_o(credit_ready_and_o[i])
       );
+
+      // synopsys translate_off
+      always_ff @ (negedge clk_i) begin
+        if (~reset_i & valid_i[i]) begin
+          assert(fifo_ready_lo[i]) else $error("Trying to enque when there is no space in FIFO, while using credit interface. i =%d", i);
```
drichmond, line 89: Do you mean ~ready_i?

drichmond: (or similar)

This assert isn't firing

tommydcjung: It should fire...

drichmond: Oh, I misread it.

This assert doesn't fire in Verilator, which is worrisome. It does fire, if I modify the if statement. 

@dpetrisko Have you seen verilator do this before?

drichmond: Edit: it does fire if I modify the if statement and use $fatal, instead of assert.

dpetrisko: weird, $error should work in verilator

drichmond:
```          assert(fifo_ready_lo[i]) else $error(1, "Trying to enque when there is no space in FIFO, while using credit interface. i =%d", i);

           if (~fifo_ready_lo[i])
            $fatal(1, "Trying to enque when there is no space in FIFO, while using credit interface. i =%d", i);

        end
```

For example, the assert does not fire here - but the fatal does

tommydcjung: Does it fire in VCS?

drichmond: No, but VCS was setting the parameters correctly (we suspect)

tommydcjung: If you set the parameter wrong in VCS, it should fire.

drichmond: It fires, but it does not terminate simulation

drichmond: It seems like `assert` is the culprit here: 

"Verilator is beginning to add support for assertions. Verilator currently only converts assertions to simple
"if (...) error" statements, and coverage statements to increment the line counters described in the coverage
section."

tommydcjung: I'm not doing anything novel here.. this pattern is used everywhere else. Maybe this is a bug in verilator?

drichmond: Unsupported, or bug. Depends on what you want to call it.

drichmond: I think this PR is good. It's necessary. 

If we want to refactor all assertions in basejump to make Verilator happy, then that is absolutely a different PR. 

I would like to hear @taylor-bsg ack this conversation before we merge.

tommydcjung: Instead of changing everything in basejump_stl for this obscure reason, I think more valuable things is to report this bug to verilator community. It may get fixed or not, but it can benefit everyone else using verilator.

dpetrisko: Wait @drichmond , are you compiling with --assert for verilator?

dpetrisko:
```
     +1800-2017ext+<ext>        Use SystemVerilog 2017 with file extension <ext>
    --assert                    Enable all assertions
    --autoflush                 Flush streams after all $displays
```

:)

drichmond: I tried that. It causes another error while compiling the RTL, so I assumed it was the wrong path

dpetrisko: We compile BP with that flag, so that's a different issue

drichmond: With #325, Verilator now throws an assert when this condition happens

PR #326, opened 2020-07-17
Fixed the signal alignment on dramsim3's read port

farzamgl:
Since the `bsg_nonsynth_dramsim3` module uses a synchronous memory for storing the data, we only need to shift the valid and address signals by one cycle in order to align them with the output data.
Also fixes the fails in `random_full` and `constrained_random` test suites in the `dramsim3_bandwidth` testbench.

### DIFF
--- a/bsg_test/bsg_nonsynth_dramsim3.v
+++ b/bsg_test/bsg_nonsynth_dramsim3.v
@@ -226,18 +226,16 @@ module bsg_nonsynth_dramsim3
   
   end
 
-  // this register is needed because we use blocking assignment 
-  // for read_done which drives read_v_li.
-  // without this register data_v_o mirrors read_v_li instead
-  // of being its value from the previous cycle.
-  logic [num_channels_p-1:0] data_v_r;
-  logic [num_channels_p-1:0] [channel_addr_width_p-1:0] read_done_ch_addr_r;
-
+  // aligning valid and address with the output data
   always_ff @(posedge clk_i) begin
-    data_v_r <= read_v_li;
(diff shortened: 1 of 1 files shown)


bsg_test/bsg_nonsynth_dramsim3.v
```diff
@@ -226,18 +226,10 @@ module bsg_nonsynth_dramsim3
...
   end
 
-  // this register is needed because we use blocking assignment 
-  // for read_done which drives read_v_li.
-  // without this register data_v_o mirrors read_v_li instead
-  // of being its value from the previous cycle.
-  logic [num_channels_p-1:0] data_v_r;
-  logic [num_channels_p-1:0] [channel_addr_width_p-1:0] read_done_ch_addr_r;
-
+  // aligning valid and address with the output data
```
tommydcjung, line 229: This makes more sense.

tommydcjung:
It might make sense to figure out why this suddenly has become a problem...

bsg_test/bsg_nonsynth_dramsim3.v
tommydcjung: I think there should be data_v_r, which resets to zero with reset_i. and assign that to data_v_o.
might also be same for read_done_ch_addr_o

tommydcjung: Basically, prevent x coming out of data_v_o after reset.

farzamgl: done.

PR #327, opened 2020-07-17
[RFC] Adding Baseline CI 

dpetrisko:
This is just a starting point, single manual test.  Few questions:

- I made jobs manual by default, so devs can just run relevant jobs.  Should we run all tests always?
- We can also build a dependency chain to run relevant tests. Maybe using flist?
- Should we integrate with bsg_ppa_regression?
- Can we standardize on all tests being verilator compatible? This will make it much more accessible to outside devs, and allow us to use the shared gitlab machines (i.e. not our own).
- An admin can add a gitlab "hook" such that it's impossible to merge failing CI.  Should we do this?

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,29 @@
+variables:
+    GIT_STRATEGY: fetch
+    GIT_SUBMODULE_STRATEGY: recursive
+
+stages:
+  - manual
+
+.job_template: &job_definition
+  when: manual
+  stage: manual
+  only:
+    refs:
+      - /^ci_.*$/
+      - master
+  artifacts:
+    when: always
+    paths:
+      - "*.log"
+  dependencies: []
+
+bsg_cache_job:
+  <<: *job_definition
+  tags:
+    - bsg
+    - vcs
+  script:
+    # TODO: Check for pass/fail
+    - make -C testing/bsg_cache/regression &> bsg_cache_regression.log
+
--- a/testing/bsg_cache/regression/Makefile
+++ b/testing/bsg_cache/regression/Makefile
@@ -1,7 +1,8 @@
 #
 # Makefile
 #
-include ../../../../bsg_cadenv/cadenv.mk
+BSG_CADENV_DIR ?= ../../../../bsg_cadenv
+include $(BSG_CADENV_DIR)/cadenv.mk
 export BASEJUMP_STL_DIR = $(abspath ../../..)
 
 WAYS_P ?= 2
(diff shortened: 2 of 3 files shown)


drichmond:
> This is just a starting point, single manual test. Few questions:
> 
> * I made jobs manual by default, so devs can just run relevant jobs.  Should we run all tests always?
What's the longest running test? How long does it take to run all tests, in a serial fashion?
> * We can also build a dependency chain to run relevant tests. Maybe using flist?
I'd be curious to hear more about this approach :p 
> * Should we integrate with bsg_ppa_regression?
I like this idea. 
> * Can we standardize on all tests being verilator compatible? This will make it much more accessible to outside devs, and allow us to use the shared gitlab machines (i.e. not our own).
I'd like to see this. As we discussed in the HB meeting this week, there's value to having the code linted by verilator, let alone run by it.
> * An admin can add a gitlab "hook" such that it's impossible to merge failing CI.  Should we do this?
I think so.

dpetrisko:
> What's the longest running test? How long does it take to run all tests, in a serial fashion?

TBD!  One of the main hurdles we'll need to face is setting up the tests so they can run just with 'make all'.  Currently, many of the tests need manual massaging to get up and running.

> I'd be curious to hear more about this approach :p

:) It sounds fancier than it is.  Basically, if we identify the dependencies of each test using json (like in bsg_ppa regression), we could run only the tests which list the changed file as a dependency.

I'll add that we need toggle coverage as a first class metric.  So if a contributor adds a new feature and toggle coverage goes down, we'll know that added tests are insufficient.

I'll also add an action item to standardize the makefile commands so that they automatically sweep parameterizations.  I've noticed many existing tests rely on the user to change parameters (even though the set of tests is automated as a target).

drichmond:
Sounds like a great start.

I'd be interested in the Makefile commands discussion.


> Last, well also need to standardize "BSG_PASS" and "BSG_FAIL" as detectable pass/fail to help out the CI jobs

Yes, we do this in replicant. I've been thinking about doing this for all print messages. BSG_INFO, BSG_DEBUG, BSG_PASS, BSG_FAIL, BSG_ERROR, BSG_WARN, etc.

drichmond:
One idea for the prints: What about a set of files in BaseJump that define print wrappers in various languages. We have one for C/C++ that @mrutt92 developed. I can create one for Make (it's basically all-but-collated). I'm sure we can create one for Python, and then one for Verilator. 

Then error/fail can be built on top of that?

dpetrisko:
Closing for lack of interest, collaborator help welcome

PR #328, opened 2020-07-24
Fixed the fake mask in the unmasked mode DRAMSim3

farzamgl:
The fake mask input to the DMA memory module was wrong in the #323.

### DIFF
--- a/bsg_test/bsg_nonsynth_dramsim3.v
+++ b/bsg_test/bsg_nonsynth_dramsim3.v
@@ -274,9 +274,6 @@ module bsg_nonsynth_dramsim3
     end
   end
 
-  assert property (@(posedge mask_i, negedge mask_i) masked_p)
-    else $warning("Ignoring mask_i because masked_p is not set");
-
   // final
   final begin
     bsg_dramsim3_exit();
--- a/bsg_test/bsg_nonsynth_dramsim3.v
+++ b/bsg_test/bsg_nonsynth_dramsim3.v
@@ -225,7 +225,7 @@ module bsg_nonsynth_dramsim3
     assign read_v_li[i] = read_done[i];
     assign read_addr_li[i] = read_done_ch_addr[i];
 
-    assign mask_li[i] = masked_p ? mask_i[i] : data_mask_width_lp'($signed(1));
+    assign mask_li[i] = masked_p ? mask_i[i] : '1;
   
     assign write_v_li[i] = data_v_i[i] & v_i[i] & write_not_read_i[i] & yumi_o[i];
     assign data_yumi_o[i] = data_v_i[i] & write_not_read_i[i] & yumi_o[i];


bsg_test/bsg_nonsynth_dramsim3.v
```diff
@@ -225,7 +225,7 @@ module bsg_nonsynth_dramsim3
     assign read_v_li[i] = read_done[i];
     assign read_addr_li[i] = read_done_ch_addr[i];
 
-    assign mask_li[i] = masked_p ? mask_i[i] : data_mask_width_lp'($signed(1));
+    assign mask_li[i] = masked_p ? mask_i[i] : {data_mask_width_lp{1'b1}};
```
mrutt92, line 228: why not just `'1`?

mrutt92: http://www.asic-world.com/systemverilog/literal_values1.html#Integer_and_Logic_Literals

farzamgl: I remember reading somewhere that if you don't specify the width of the signal it defaults to 32 or some other number based on the synthesis tool. But it was a long time ago and I think it was about Verilog, not SystemVerilog. I just wanted to make sure that doesn't happen regardless of the mask width.

dpetrisko:
- ‘1 will autoinfer the width. 1 is automatically 32 bits (ironically, that’s the original bug)
- it was tested with the mask parameter enabled (on BP), not on manycore where the mask parameter is disabled
- For new modules or old modules with testbenches, basejump stl policy is full toggle coverage. It’s tough to enforce this for new features of old modules, because that means small features require disproportionate work, which discourages contributions

drichmond:
> * ‘1 will autoinfer the width. 1 is automatically 32 bits (ironically, that’s the original bug)
> * it was tested with the mask parameter enabled (on BP), not on manycore where the mask parameter is disabled

@farzamgl, wasn't there a failure message in the testbench in basejump?

farzamgl:
> > * ‘1 will autoinfer the width. 1 is automatically 32 bits (ironically, that’s the original bug)
> > * it was tested with the mask parameter enabled (on BP), not on manycore where the mask parameter is disabled
> 
> @farzamgl, wasn't there a failure message in the testbench in basejump?

Yes. There was a failure message in 3/5 of the tests of the 1/3 dramsim3 testbenches which was overlooked by we when I made a small change at the end. The fail was not obvious because the last test succeeded and you have to actually look at all the individual 5 logs to check the other ones.

farzamgl:
> * ‘1 will autoinfer the width. 1 is automatically 32 bits (ironically, that’s the original bug)
So '1 is always safe for all widths in both Verilog and SystemVeilog?

drichmond:
Is it valid verilog syntax?

dpetrisko:
It's as safe as '0, which is accepted by every tool that I know of. Strictly speaking, I think it is SystemVerilog 2005

dpetrisko:
Style guide says to use '0 for zero, so I think '1 should also be fine.

taylor-bsg:
Hi,

Thanks for the discussion on this.

I updated the style guide (which is for SystemVerilog only), to say that '1
is okay.

M

farzamgl:
Good to know. Changed the commit to '1

dpetrisko:
Really that’s the benefit of #327

drichmond:
Yes, big fan of #327.  Would it catch the issue here? Would it see the testbench failure?

dpetrisko:
Not specifically, since it's only set up for bsg_cache, but assuming we add it to the regression

drichmond:
Who wants to merge this? it works in all the situations i've tested.

drichmond:
LGTM

PR #330, opened 2020-07-31
Parallel In Serial Out Update

gaozihou:
Original version PISO purports to be doubly helpful, but it waits for the output yumi signal to be asserted before asserting the input ready signal, creating a dependence from output to input. This avoids a bubble between successive parallel words, but is a little distasteful.

This PR eliminates the dependency from output to input by adding an extra data buffer between input and output.

Tests passed for 96 different configurations.


bsg_dataflow/bsg_parallel_in_serial_out.v
```diff
@@ -2,210 +2,230 @@
...
+    ,.data_o (fifo_data_lo[els_p-1:1])
+    ,.yumi_i (fifo_yumi_li)
+    );
 
     /**
-     * State Machine Logic
+     * State machine on the input data side
+     *
+     * When 2 element buffer is used, and two parallel data arrive consecutively,
+     * the first word of second parallel data may be accepted by fifo0 repetitively.
```
dpetrisko, line 142: repetitively -> repeatedly

bsg_dataflow/bsg_parallel_in_serial_out.v
```diff
@@ -2,210 +2,230 @@
...
-     * is going to accept our data (ie. yumi_i). If we are done transmitting
-     * data, we should stall the counter on the last bit. 
+     * The shift_ctr_r register stores the word we are transmitting. Whenever
+     * we reset or done transmitting data, we clear the shift_ctr_r register.
+     * When data fifo has valid data output, we will increment the register if 
+     * the outside world is going to accept our data (ie. yumi_i).
      */
 
-   // synopsys sync_set_reset "reset_i"
+    // synopsys sync_set_reset "reset_i"
```
dpetrisko, line 197: Can instead use: https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_misc/bsg_counter_clear_up_one_hot.v

and a one-hot mux down below for faster hardware

bsg_dataflow/bsg_parallel_in_serial_out.v
```diff
@@ -2,210 +2,230 @@
...
-        assign data_li = data_i;
+    logic wait_fifo1_r;
+
+    assign fifo0_v_li = valid_i & ~wait_fifo1_r;
+    assign fifo1_v_li = valid_i;
+    assign ready_o = fifo1_ready_lo;
+
+    if (use_minimal_buffering_p == 0)
+      begin: twobuf
+        // synopsys sync_set_reset "reset_i"
```
dpetrisko, line 156: Maybe use bsg_dff_reset_set_clear here, it could be more clear

bsg_dataflow/bsg_parallel_in_serial_out.v
```diff
@@ -2,210 +2,230 @@
...
+        always_ff @(posedge clk_i)
+            if (reset_i)
+                wait_fifo1_r <= 1'b0;
+            else
+                // fifo0_ready_lo is guaranteed to be 1'b1 when wait_fifo1_r == 1'b0
```
dpetrisko, line 161: Sorry, I don't follow where these invariants come from.  Maybe a table could help?

bsg_dataflow/bsg_parallel_in_serial_out.v
gaozihou: Thanks Dan!

bsg_dataflow/bsg_parallel_in_serial_out.v
gaozihou: Thanks! Replaced with bsg_dff_reset_set_clear.

bsg_dataflow/bsg_parallel_in_serial_out.v
gaozihou: Great idea! Table added to illustrate possible states of registers.

bsg_dataflow/bsg_parallel_in_serial_out.v
gaozihou: Great suggestion! I cleaned up the code with Basejump IPs for counter and mux.

Yeah I agree that one-hot coding results in faster hardware. I guess binary coding results in smaller hardware? (like fewer registers) 

(For now the PISO is using bsg_counter_clear_up and bsg_mux, would be very easy to convert to one-hot version if needed)

dpetrisko: Yeah, the downside is that there are more registers for counter (N vs lgN).  The upside is you get rid of a comparator and decoder on the paths.  Can do some PPA analysis later, but we can merge this version for now!

How about a comment to analyse this tradeoff for later?

gaozihou: Thanks Dan! Comments added.

bsg_dataflow/bsg_parallel_in_serial_out.v
```diff
@@ -2,136 +2,42 @@
...
-     * being accepted from the outside word, we can accept data on that cycle. 
-     */
-    assign ready_o = (state_r == eRX) || done_tx_n;
+    (input                           clk_i
+    ,input                           reset_i
 
+    // Data Input Channel (Ready and Valid)
+    ,input                           valid_i
+    ,input  [els_p-1:0][width_p-1:0] data_i
+    ,output                          ready_o
```
taylor-bsg, line 31: ready_and_o per updated SystemVerilog Coding guidelines

taylor-bsg (commented):
is this ready to be integrated?

bsg_dataflow/bsg_parallel_in_serial_out.v
gaozihou: I am a little worried about backward compatibility issue, since it is also instantiated in other repos (found in bsg_replicant, bsg_fpga, bsg_motherboards, black-parrot).
For clarity I added **ready_and_o** to comments, so that users could see it when instantiating.

taylor-bsg:
I think it’s okay — we will discover the places where we are not versioning
controlling.m very quickly, and the fix is minor.

M

PR #331, opened 2020-08-06
Added new address mapping to DRAMSim3

farzamgl:
- Added the `rochrababgco` address mapping to the translation modules
- Fixed a small error in Verilog parameter validation vs DRAMSim3

### DIFF
--- a/bsg_test/bsg_dramsim3_pkg.v
+++ b/bsg_test/bsg_dramsim3_pkg.v
@@ -26,6 +26,14 @@ package bsg_dramsim3_hbm2_8gb_x128_pkg;
   parameter string config_p="HBM2_8Gb_x128.ini";
   parameter address_mapping_p=bsg_dramsim3_pkg::e_ro_ra_bg_ba_ch_co;
 
+  typedef struct packed {
+    logic [$clog2(num_rows_p)-1:0] ro;
+    logic [$clog2(num_bg_p)-1:0] bg;
+    logic [$clog2(num_ba_p)-1:0] ba;
+    logic [$clog2(num_columns_p)-1:0] co;
+    logic [$clog2(data_width_p>>3)-1:0] byte_offset;
+  } dram_ch_addr_s;
+
 endpackage // bsg_dramsim3_hbm2_8gb_x128_pkg
 
 package bsg_dramsim3_hbm2_4gb_x128_pkg;
@@ -41,6 +49,15 @@ package bsg_dramsim3_hbm2_4gb_x128_pkg;
   parameter longint size_in_bits_p=2**35; // 4GB (32Gb)
   parameter string config_p="HBM2_4Gb_x128.ini";
   parameter address_mapping_p=bsg_dramsim3_pkg::e_ro_ra_bg_ba_ch_co;
+
+  typedef struct packed {
+    logic [$clog2(num_rows_p)-1:0] ro;
+    logic [$clog2(num_bg_p)-1:0] bg;
+    logic [$clog2(num_ba_p)-1:0] ba;
+    logic [$clog2(num_columns_p)-1:0] co;
+    logic [$clog2(data_width_p>>3)-1:0] byte_offset;
+  } dram_ch_addr_s;
+
 endpackage // bsg_dramsim3_hbm2_8gb_x128_pkg
 
 package bsg_dramsim3_lpddr3_8gb_x32_1600_pkg;
@@ -56,6 +73,14 @@ package bsg_dramsim3_lpddr3_8gb_x32_1600_pkg;
   parameter longint size_in_bits_p=2**34; // 2GB (16Gb)
   parameter string config_p="LPDDR3_8Gb_x32_1600.ini";
   parameter address_mapping_p=bsg_dramsim3_pkg::e_ro_ch_ra_ba_bg_co;
+
+  typedef struct packed {
+    logic [$clog2(num_rows_p)-1:0] ro;
+    logic [$clog2(num_ba_p)-1:0] ba;
+    logic [$clog2(num_columns_p)-1:0] co;
+    logic [$clog2(data_width_p>>3)-1:0] byte_offset;
+  } dram_ch_addr_s;
+
 endpackage // bsg_dramsim3_lpddr3_8gb_x32_1600_pkg
 
 `define dramsim3_ba_width(num_ba_mp) $clog2(num_ba_mp)
--- a/bsg_test/bsg_dramsim3_pkg.v
+++ b/bsg_test/bsg_dramsim3_pkg.v
@@ -18,8 +18,10 @@ package bsg_dramsim3_hbm2_8gb_x128_pkg;
   parameter int data_width_p=256;
   parameter int num_channels_p=8;
   parameter int num_columns_p=64;
+  parameter int num_rows_p=32768;
   parameter int num_ba_p=4;
   parameter int num_bg_p=4;
+  parameter int num_ranks_p=1;
   parameter longint size_in_bits_p=2**36; // 8GB (64Gb)
   parameter string config_p="HBM2_8Gb_x128.ini";
   parameter address_mapping_p=bsg_dramsim3_pkg::e_ro_ra_bg_ba_ch_co;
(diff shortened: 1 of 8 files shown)


bsg_test/bsg_nonsynth_dramsim3_map.v
```diff
@@ -39,5 +53,31 @@ module bsg_nonsynth_dramsim3_map
...
+      else if (i >= mem_ra_pos_lp)
+        assign mem_addr_o[i] = ch_addr_i[ra_pos_lp+i-mem_ra_pos_lp];
+      else if (i >= mem_ba_pos_lp)
+        assign mem_addr_o[i] = ch_addr_i[ba_pos_lp+i-mem_ba_pos_lp];
+      else if (i >= mem_bg_pos_lp)
+        assign mem_addr_o[i] = ch_addr_i[bg_pos_lp+i-mem_bg_pos_lp];
+      else if (i >= mem_co_pos_lp)
+        assign mem_addr_o[i] = ch_addr_i[co_pos_lp+i-mem_co_pos_lp];
+      else
+        assign mem_addr_o[i] = 1'b0;
```
tommydcjung, line 79: Can we just do "assign mem_addr_o = {ch_addr_i[], ch_addr_i[], ch_addr_i[], ...};" ?

farzamgl: Not really because SystemVerilog doesn't like signals with zero widths and some of the fields can be empty depending on the parameters.

testing/bsg_test/dramsim3_bandwidth/testbench.v
```diff
@@ -1,7 +1,7 @@
 `include "bsg_nonsynth_dramsim3.svh"
 
 
-`define dram_pkg bsg_dramsim3_hbm2_8gb_x128_pkg
+`define dram_pkg bsg_dramsim3_lpddr3_8gb_x32_1600_pkg
```
taylor-bsg, line 4: Instead of flip-flopping back and forth by manually editing the file, does it make sense to just add two more tests and then include the common source file?

farzamgl: Oh, that's actually wrong. I forgot to revert that line back. That testbench is not for lpddr3.

mrutt92: Or define it from the command line.

bsg_test/bsg_nonsynth_dramsim3_map.v
dpetrisko: Not necessarily suggesting it, but I recently discovered that you can use zero width signal in concatenations:

https://verificationacademy.com/forums/systemverilog/zero-bit-width-array

farzamgl: Yeah but you cannot address a zero width part of an array with [... +: 0]

testing/bsg_test/dramsim3_bandwidth/testbench.v
farzamgl: The problem with that is that the python files that generate the trace are all hardcoded for HBM2.

bsg_test/bsg_nonsynth_dramsim3_map.v
dpetrisko: The idea would be to do something like this {actual_slice!=0{ch_addr_i[0+:BSG_MAX(1, actual_slice)}}

farzamgl: I see. Yeah, I think this way is a bit cleaner.

farzamgl:
Can we merge this, if there aren't any specific issues?

drichmond:
Has anyone run this PR in HammerBlade?

bsg_test/bsg_dramsim3_pkg.v
```diff
@@ -68,22 +75,11 @@ endpackage // bsg_dramsim3_hbm2_8gb_x128_pkg
...
-    logic [`dramsim3_ro_width(ch_addr_width_mp, data_width_mp, num_ba_mp, num_bg_mp, num_columns_mp)-1:0] ro; \
-    logic [`dramsim3_bg_width(num_bg_mp)-1:0] bg; \
-    logic [`dramsim3_ba_width(num_ba_mp)-1:0] ba; \
-    logic [`dramsim3_co_width(num_columns_mp)-1:0] co; \
-    logic [`dramsim3_byte_offset_width(data_width_mp)-1:0] byte_offset; \
-  } typename_mp
+`define dramsim3_ra_width_pkg(dram_pkg) \
+  `dramsim3_ra_width(dram_pkg::num_ranks_p)
 
-`define declare_dramsim3_ch_addr_s_with_pkg(typename_mp, dram_pkg) \
```
drichmond, line 83: How do I replace this in HammerBlade?

drichmond: @farzamgl

farzamgl: I see. I removed this struct because some of the fields can be empty(for example if there's only 1 rank), and we cannot have an empty signal in the struct.

drichmond: Removing things isn't the intent of your PR, right? You only changed the address map?

tommydcjung: I think this would also work in replicant? https://github.com/bespoke-silicon-group/bsg_manycore/blob/master/testbenches/common/v/spmd_testbench.v#L881

farzamgl: It wasn't the intent, but during the address mapping changes I realized that we could have 0-bit wide signals in that struct(depending on the configuration), so I had to remove it.

drichmond: > I think this would also work in replicant? https://github.com/bespoke-silicon-group/bsg_manycore/blob/master/testbenches/common/v/spmd_testbench.v#L881

My initial (rapid) reaction is yes. Though I would like to remove the magic constants.

farzamgl: I think you can use the `dramsim3_*_width` macros for the struct definition in HB code.

mrutt92: I made those macros for the convenience of designs that wanted to use the HBM simulation. They're not used in the design itself because they match DRAMSim3's address configuration, which is specified either in the C++ source or some config file. I didn't anticipate the empty field problem. Maybe the thing to do here is follow what @tommydcjung did in that spmd_testbench except define them in this header with some general name. Magic numbers are OK as long as they're defined in exactly one place, right? That way people who want to use this don't have to go rooting through our projects to find the address map they need.

drichmond: @farzamgl do you need to remove this macro for the correctness of your code?

dpetrisko: I’m going to hop in without full context here and say this sounds like a problem which can be solved with unions

farzamgl: > @farzamgl do you need to remove this macro for the correctness of your code?

My code doesn't use any structs for the address, but yes it was creating issues in its current form(in the testbench).

> I’m going to hop in without full context here and say this sounds like a problem which can be solved with unions

@dpetrisko Is there a way to define the struct with unions so it doesn't create 0-bit signal errors?

tommydcjung: I agree with @mrutt92 and I think we should just be agile and move on.

tommydcjung (approved):

drichmond (changes requested):
We need to find a satisfactory solution, before moving on. Not all questions have been addressed.

bsg_test/bsg_dramsim3_pkg.v
dpetrisko:
```
typedef union packed
{
  struct packed
  {
     logic [`BSG_MAX(col_width_p, 1)-1:0] col;
     logic [`BSG_MAX(row_width_p, 1)-1:0] row;
   } addr_with_row_and_col;
  struct packed
  {
     logic [`BSG_MAX(row_width_p)-1:0] row;
  } addr_without_col;
} dram_addr
```

@farzamgl  something like this?

farzamgl: I don't this is practical. There's going to be 16 possible combinations in the union.
I think generating the struct in your own testbench is a cleaner way of doing it.

dpetrisko: The struct definitions exists either way

bsg_test/bsg_dramsim3_pkg.v
```diff
@@ -32,33 +34,38 @@ package bsg_dramsim3_hbm2_4gb_x128_pkg;
...
+  parameter int num_rows_p=16384;
   parameter int num_ba_p=4;
   parameter int num_bg_p=4;
+  parameter int num_ranks_p=1;
   parameter longint size_in_bits_p=2**35; // 4GB (32Gb)
   parameter string config_p="HBM2_4Gb_x128.ini";
   parameter address_mapping_p=bsg_dramsim3_pkg::e_ro_ra_bg_ba_ch_co;
 endpackage // bsg_dramsim3_hbm2_8gb_x128_pkg
 
+package bsg_dramsim3_lpddr3_8gb_x32_1600_pkg;
```
tommydcjung, line 46: How about inside each package define the address mapping struct? I'm against the union idea, because like @farzamgl said, it's not a practical solution..

mrutt92: Union's seem like a good thing to use if you're trying to share structs on the same logic, which is not the case here. Defining a struct for each package seems more like what we want.

farzamgl: This is a better solution imo, but you have to change the code in HB.

mrutt92: Stuff happens.

mrutt92: It's why version control exists :-)

tommydcjung: Yeah we can update the HB to use this better solution. Not a big deal.

drichmond (approved):
To make my position extremely clear: BaseJump should evolve separately from HB, and HB is responsible for adapting.

That being said - we have a limited amount person-hours (we're all working on new features, deadlines, etc), so if there is a way to avoid introducing cascading changes they should be discussed. The excellent discussion has proven this isn't possible, but it seems like a good compromise has been found.

The irony is that we've all spent more time discussing this bug -- and in that time we collectively could have implemented and tested a fix in HB.

The current solution looks good, and thanks to @farzamgl for helping me get it into HB.

PR #333, opened 2020-08-20
Locking arbiter with implicit unlock

Songchun-Li:
This is a new locking arbiter based on fixed priority arbiter for BP Stream Protocol, and has following features:
1. The channel is locked when a request is granted and the corresponding lock input signal is high.
2. Lock output signal keeps high when it is locked, even if no new request on the locked channel is granted.
3. The locked channel will be unlock when a request on the locked channel is granted and the lock input signal is low.

The module is tested with a trace-replay based test-bench.

BTW, I think it is a good idea to rename this new arbiter and the original locking arbiter, like 
bsg_locking_arb_fixed              --> bsg_arb_fixed_unlock
bsg_locking_arb_fixed_new.    --> bsg_arb_fixed_lock


dpetrisko:
>The locked channel will be unlock when a request on the locked channel is granted and the lock input signal is low.
I'm not sure if we want to require granting on low as the unlock condition. It may be better to simply unlock if lock goes low. This would allow a transaction to abort

dpetrisko:
>BTW, I think it is a good idea to rename this new arbiter and the original locking arbiter, like
>bsg_locking_arb_fixed --> bsg_arb_fixed_unlock
>bsg_locking_arb_fixed_new. --> bsg_arb_fixed_lock

@stdavids What about combining these three variants into bsg_arb_fixed, similar to the other arbiters in basejump having multiple operating modes? So we would parameterize on unlock_type_p = none, implicit, explicit where the interface is:
- "None" has lock disconnected
- "Implicit" has a lock signal per input lock the arb when granted, and unlock when that lock lowers
- "Explict" automatically locks on grant and has a single lock signal unlock the bus

```
  ( input   clk_i
  , input   reset_i

  , input        [num_locks_p-1:0] switch_lock_i
  , input        [inputs_p-1:0] reqs_i
  , output logic [inputs_p-1:0] grants_o

  , input   ready_i
  , output logic lock_o
  );  
```

Songchun-Li:
The unlock logic is changed to unlock when the lock_i at the locked channel is low. 

Besides, the lock_o signal is updated. Now, it passes the lock_i signal on the granted channel when it is unlocked, also the one on the locked channel when it is locked at the same cycle.

The test trace is also updated and the arbiter passed the test.

Songchun-Li:
The new locking arbiter is now able to work at either implicit or explicit operating mode, but I am not sure whether it is a good idea to include a none operating mode, since it will just instantiate a bsg_arb_fixed and kind of overlaps with the original module

Songchun-Li:
The test trace for explicit mode is added. And the tests can be ran by 
```
make sim MODE=implicit
make sim MODE=explicit
```

bsg_misc/bsg_locking_arb_fixed_new.v
```diff
@@ -0,0 +1,72 @@
...
+        ,.sel_one_hot_i(grants_o)
+        ,.data_o(lock_selected_lo)
+        );
+      // The lock_i can be passed to lock_o at the same cycle
+      assign lock_o = (is_locked) ? lock_onlocked_lo : lock_selected_lo;
+    end
+  else  // Explicit operating mode
+    begin
+      assign unlock = switch_lock_i;
+      assign lock_o = 1'b0;
```
dpetrisko, line 50: Shouldn't this be high during a transaction (&req_mask_r)?

bsg_misc/bsg_locking_arb_fixed_new.v
```diff
@@ -0,0 +1,72 @@
...
+  );  
+
+  wire [inputs_p-1:0] not_req_mask_r, req_mask_r;
+  wire unlock;  
+  
+  if (unlock_type_p == 0) // Implicit operating mode
+    begin
+      wire is_locked = |not_req_mask_r;
+      wire lock_onlocked_lo, lock_selected_lo;
+      bsg_mux_one_hot #(.width_p(1) ,.els_p(inputs_p) )
```
dpetrisko, line 29: Aren't these muxes just bitwise ands?

bsg_misc/bsg_locking_arb_fixed_new.v
```diff
@@ -0,0 +1,72 @@
...
+      // The lock_i can be passed to lock_o at the same cycle
+      assign lock_o = (is_locked) ? lock_onlocked_lo : lock_selected_lo;
+    end
+  else  // Explicit operating mode
+    begin
+      assign unlock = switch_lock_i;
+      assign lock_o = 1'b0;
+    end
+
+  bsg_dff_reset_en #( .width_p(inputs_p) )
```
dpetrisko, line 53: Can you write a comment on the operation here? (an ASCII diagram would help).  The ~not_req_mask is a little mind-bending

bsg_misc/bsg_locking_arb_fixed_new.v
```diff
@@ -0,0 +1,72 @@
+// unlock_type_p = implicit / explicit
+// 0 implicit: lock when lock and req is granted. unlock when ~lock
+// 1 explicit: automatically locks on grant and has a single lock signal high
+
+module bsg_locking_arb_fixed_new #( parameter inputs_p="inv"
```
dpetrisko, line 5: You can just name bsg_locking_arb_fixed and replace the old one

bsg_misc/bsg_locking_arb_fixed_new.v
```diff
@@ -0,0 +1,72 @@
...
+      assign lock_o = (is_locked) ? lock_onlocked_lo : lock_selected_lo;
+    end
+  else  // Explicit operating mode
+    begin
+      assign unlock = switch_lock_i;
+      assign lock_o = 1'b0;
+    end
+
+  bsg_dff_reset_en #( .width_p(inputs_p) )
+    req_words_reg
```
dpetrisko, line 54: Does this have a bubble between consecutive, different channel transaction?

Songchun-Li: Currently, it is unlocked at the next cycle after receiving low lock_i from the locked channel, but there is no bubble between consecutive, different channel transactions, because we implemented the bp_stream with lock_o low for the last packet. In this way, the arbiter is unlocked when the consecutive, different channel request arrives, and no bubble is presented.

bsg_misc/bsg_locking_arb_fixed_new.v
Songchun-Li: Yep, the diagram with a 4-bit example is added. The not_req_mask_r is a little mind-bending, but that was what it used to be in the locking_arb, so I just fit the new operating mode in it.

bsg_misc/bsg_locking_arb_fixed_new.v
Songchun-Li: Set it to zero because there was no lock_o for the original explicit mode. Fixed in the new commit

Songchun-Li:
It has been updated and ready for review

bsg_misc/bsg_locking_arb_fixed.v
```diff
@@ -1,24 +1,64 @@
...
   );  
 
   wire [inputs_p-1:0] not_req_mask_r, req_mask_r;
+  wire unlock;  
+  
+  if (unlock_type_p == 0) // Implicit operating mode
+    begin
+      wire is_locked = |not_req_mask_r;
+      wire lock_onlocked_lo, lock_selected_lo;
+      assign lock_onlocked_lo = |(switch_lock_i & req_mask_r);
```
dpetrisko, line 29: minor: wire lock_onlocked_lo =

dpetrisko (approved):

dpetrisko:
Closing after offline discussion where original arbiter is sufficient

PR #334, opened 2020-08-29
Fixing saturation of read and write counters in DRAM controller

farzamgl:
If these counters that are used to enforce the minimum timing delays of the DRAM model overflow, it can add additional delays between issued DRAM operations.


drichmond:
I can merge, but this isn't my area of expertise. I'm guessing Chun is pretty busy right now, so is there anyone else who can approve?

farzamgl:
It's not urgent, so we can merge whenever it's reviewed.

hema0730 (approved):
Great catch! Typo introduced by copy paste.

PR #338, opened 2020-09-18
Update noc testbench filelist

gaozihou:
Resolve undefined macro problem in testbenches, so that they can run without error.
Keep filelists up-to-date.


dpetrisko (approved):
LGTM!

PR #342, opened 2020-09-30
Prevent erroring on trace-replay nop

dpetrisko:
Missing case statement leads to ```### bsg_trace_replay UNKNOWN op 0 (testbench.trace_replay)``` for nops and terminates the simulation.  Credit to @gaozihou for the debug!


taylor-bsg (commented):
So basically we had no nops in any of our traces, hence why this bug was never exposed?

tommydcjung:
I think some testbenches that use nop still use bsg_nonsynth_fsb_node_trace_replay, which didn't have this problem.

taylor-bsg (approved):

Issue #341, opened 2020-09-28
[bsg_idiv_iterative] might be unnecessary buffer.


tommydcjung:
https://github.com/bespoke-silicon-group/basejump_stl/blob/5ac2a5fd51b3369745f4aab5f7dc13cc3484495d/bsg_misc/bsg_idiv_iterative.v#L138

taylor-bsg:
I agree -- the bsg_buf and bsg_buf_ctrl's can be removed from this module.
These are inherited from days when CAD tools didn't work that well. Perhaps
that will result in some area savings in BP and the manycore?

M

taylor-bsg:
see also https://github.com/bespoke-silicon-group/basejump_stl/issues/193

taylor-bsg:
https://github.com/bespoke-silicon-group/basejump_stl/pull/343

Issue #193, opened 2019-11-22
clean up bsg_idiv


taylor-bsg:
https://github.com/bespoke-silicon-group/basejump_stl/blob/91c5ab0e37ce868bb71a0616298cc1c70733c668/bsg_misc/bsg_idiv_iterative.v#L46

clean up the module to remove buffering of datapath signals, and structural instantiation of single-bit control registers.

dpetrisko:
> structural instantiation of single-bit control registers.

What's the style guide ruling on this? Control registers in sequential logic blocks, structural datapath?

taylor-bsg:
Open to suggestions, I think the style guide ruling is basically that you
should have a schematic for your hardware and that the relationship between
between the schematic and your code should be clear.  We don’t have any
specific requirement that you structurally instantiate anything instead of
writing RTL, but hierarchy should exist around logic that needs to be
placed in a particular place, or on datapath blocks where critical paths
are likely to approach or exceed 30 FO4. Async reset flops should be
instantiated structurally. (Hoping that I am not regretting this relaxing
of the rules.)

PR #343, opened 2020-09-30
[bsg_idiv_iterative] remove unnecessary buffer

tommydcjung:


tommydcjung:
https://github.com/bespoke-silicon-group/basejump_stl/issues/341

taylor-bsg:
See also https://github.com/bespoke-silicon-group/basejump_stl/issues/193

bsg_misc/bsg_idiv_iterative.v
```diff
@@ -134,46 +134,35 @@ module bsg_idiv_iterative #(parameter width_p=32)
    //   assign add_in1 = (opB ^ {width_p+1{opB_inv}}) & {width_p+1{opB_clr_l}};
```
taylor-bsg, line 134: This code is for the case where we are bit-stacking the divide unit which is probably not the comment case. Maybe add a bitstack_p parameter, default 0, and then we can make the default these two assign statements and comment out the bit-stack version so that we don't have to worry about unnecessary hierarchy introducing extra gates?

tommydcjung: I see, but what's the reason to keep the bit-stack version?

taylor-bsg (approved):

bsg_misc/bsg_idiv_iterative.v
taylor-bsg: If we had a design that needed ultra dense dividers i suppose...

PR #350, opened 2020-10-27
PISO Break Unnecessary Constraint on v_i and data_i

gaozihou:
Current design requires that when v_i==1, value of data_i must hold until ready_and_o==1. This requirement may not always be fulfilled when attached to other BaseJump modules.

Updated design breaks the constraint by moving "extra buffering" from first word to last word, therefore data_i always enqueue into PISO within one cycle.


dpetrisko (approved):

hema0730 (approved):
LGTM!

PR #351, opened 2020-11-06
Remove inside in generate if

dalance:
In LRM, `generate if` takes `constant_expression`, but `inside` is `expression`.
So `inside` can't be used in `generate if`.

> LRM A.4.2
> if_generate_construct ::=
if ( constant_expression ) generate_block [ else generate_block ]

### DIFF
--- a/bsg_test/bsg_nonsynth_dpi_from_fifo.v
+++ b/bsg_test/bsg_nonsynth_dpi_from_fifo.v
@@ -67,7 +67,7 @@ module bsg_nonsynth_dpi_from_fifo
    bit    rx_r = 0;
    
    // Check if width_p is a ctype width. call $fatal, if not.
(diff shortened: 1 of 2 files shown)


taylor-bsg:
Thanks! We will take a look.

dpetrisko:
@dalance Out of curiosity, do you know of a tool for which this usage is unsupported?

dalance:
@dpetrisko This issue  was detected at [sv-tests](https://github.com/SymbiFlow/sv-tests).
In the testcase, [sv-parser](https://github.com/dalance/sv-parser), which is my original SV parser, failed to parse this code.

In commercial EDA tools, Cadence Xcelium seems to detect this case as syntax error.

taylor-bsg (approved):
Looks good! Take note @drichmond

PR #356, opened 2020-11-24
Adding gf12 macro defines for individual design memories

dpetrisko:
This PR adds hooks to the hardened memories so that external designs can define their own macros without having to change the files here.  This has several advantages:
- Not having to maintain separate incompatible tapeout branches for each design
- Not all memories have an optimal configuration; this allows us to specialize per design
- Adds hooks for users who are straight compiling instead of using a "swap" methodology

### DIFF
--- a/hard/gf_14/bsg_mem/bsg_mem_1r1w_sync_macros.vh
+++ b/hard/gf_14/bsg_mem/bsg_mem_1r1w_sync_macros.vh
@@ -1,4 +1,7 @@
 
+`ifndef BSG_MEM_1R1W_SYNC_MACROS_VH
+`define BSG_MEM_1R1W_SYNC_MACROS_VH
+
 `define bsg_mem_1r1w_sync_macro(words,bits,mux)\
   if (harden_p && els_p == words && width_p == bits)          \
     begin: macro                                              \
@@ -20,3 +23,5 @@
     );                                                        \
   end
       
+`endif
+
--- a/hard/gf_14/bsg_mem/bsg_mem_1r1w_sync_macros.vh
+++ b/hard/gf_14/bsg_mem/bsg_mem_1r1w_sync_macros.vh
@@ -1,0 +1,22 @@
+
+`define bsg_mem_1r1w_sync_macro(words,bits,mux)\
+  if (harden_p && els_p == words && width_p == bits)          \
+    begin: macro                                              \
+    gf14_1r1w_d``words``_w``bits``_m``mux                     \
+      mem (                                                   \
+      .CLKA   (clk_i)                                         \
+      ,.CLKB  (clk_i)                                         \
+      ,.CENA  (~r_v_i)                                        \
+      ,.AA    (r_addr_i)                                      \
+      ,.QA    (r_data_o)                                      \
+      ,.CENB  (~w_v_i)                                        \
+      ,.AB    (w_addr_i)                                      \
+      ,.DB    (w_data_i)                                      \
+      ,.EMAA  (3'b011)                                        \
+      ,.EMAB  (3'b011)                                        \
+      ,.EMASA (1'b0)                                          \
+      ,.STOV  (1'b0)                                          \
+      ,.RET1N (1'b1)                                          \
+    );                                                        \
+  end
+      
--- a/hard/gf_14/bsg_mem/bsg_mem_1rw_sync.v
+++ b/hard/gf_14/bsg_mem/bsg_mem_1rw_sync.v
@@ -40,7 +40,10 @@ module bsg_mem_1rw_sync #( parameter width_p = -1
 
   wire unused = reset_i;
 
-  // TODO: Define more hardened macro configs here
+  // TODO: Set hardened macro configs in this define
+  `ifdef BSG_MEM_HARD_1RW_SYNC_MACROS
+  `BSG_MEM_HARD_1RW_SYNC_MACROS
+  `endif
   `bsg_mem_1rw_sync_macro(512,64,4) else
   `bsg_mem_1rw_sync_macro(256,96,2) else
   `bsg_mem_1rw_sync_macro(1024,46,4) else
--- a/hard/gf_14/bsg_mem/bsg_mem_1rw_sync.v
+++ b/hard/gf_14/bsg_mem/bsg_mem_1rw_sync.v
@@ -1,23 +1,5 @@
 
-`define bsg_mem_1rw_sync_macro(words,bits,mux)       \
-  if (harden_p && els_p == words && width_p == bits) \
-    begin: macro                                     \
-      gf14_1rw_d``words``_w``bits``_m``mux           \
-        mem                                          \
-          ( .CLK   ( clk_i  )                        \
-          , .A     ( addr_i )                        \
-          , .D     ( data_i )                        \
-          , .Q     ( data_o )                        \
-          , .CEN   ( ~v_i   )                        \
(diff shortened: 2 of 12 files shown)


hard/gf_14/bsg_mem/bsg_mem_1rw_sync_mask_write_bit.v
```diff
@@ -73,31 +77,6 @@ module bsg_mem_1rw_sync_mask_write_bit #( parameter width_p = -1
   `bsg_mem_1rw_sync_mask_write_bit_macro(64,58,2) else
   `bsg_mem_1rw_sync_mask_write_bit_banked_macro(64,116,2,1) else
   
-  // HACKED VERSION OF THE FOLLOWING
```
dpetrisko, line 76: I also removed this hack.  It was relevant for the bp cce directory in the original tapeout, but has been superceded by 'banked_macro'

stdavids (approved):

tommydcjung:
I think it would be nice to have a header file that has all the macros "bsg_mem_1r1w_sync_macro" and etc, and the existing hard files can include that. And in the design folders, the user can create design-specific hard files using the macros in the same header file.

dpetrisko:
>I think it would be nice to have a header file that has all the macros "bsg_mem_1r1w_sync_macro" and etc, and the existing hard files can include that. And in the design folders, the user can create design-specific hard files using the macros in the same header file.

So this would basically be
bsg_mem_1r1w_sync_macros.vh:
```
`define bsg_mem_1r1w_sync_macro(words,bits,mux)\
  if (harden_p && els_p == words && width_p == bits)     \
    begin: macro                                            \
    gf14_1r1w_d``words``_w``bits``_m``mux                     \
      mem (                                                    \
      .CLKA   (clk_i)                                           \
      ,.CLKB  (clk_i)                                           \
      ,.CENA  (~r_v_i)                                         \
      ,.AA    (r_addr_i)                                      \
      ,.QA    (r_data_o)                                      \
      ,.CENB  (~w_v_i)                                        \
      ,.AB    (w_addr_i)                                      \
      ,.DB    (w_data_i)                                      \
      ,.EMAA  (3'b011)                                        \
      ,.EMAB  (3'b011)                                        \
      ,.EMASA (1'b0)                                          \
      ,.STOV  (1'b0)                                          \
      ,.RET1N (1'b1)                                          \
    );                                                        \
  end
```

bsg_mem_1r1w_sync.v:
```
  `bsg_mem_1r1w_sync_macro(32,92,1) else
    begin: notmacro
    bsg_mem_1r1w_sync_synth #(
      .width_p(width_p)
      ,.els_p(els_p)
      ,.read_write_same_addr_p(read_write_same_addr_p)
    ) synth (.*); 
  end

   //synopsys translate_off
   initial
     begin
        $display("## %L: instantiating width_p=%d, els_p=%d, read_write_same_addr_p=%d, harden_p=%d (%m)",width_p,els_p,read_write_same_addr_p,harden_p);
     end

   always_ff @(negedge clk_i)
     if (w_v_i)
       begin
          assert ((reset_i === 'X) || (reset_i === 1'b1) || (w_addr_i < els_p))
            else $error("Invalid address %x to %m of size %x\n", w_addr_i, els_p);

          assert ((reset_i === 'X) || (reset_i === 1'b1) || ~(r_addr_i == w_addr_i && w_v_i && r_v_i && !read_write_same_addr_p && !disable_collision_warning_p))
            else
              begin
                 $error("X'ing matched read address %x (%m)",r_addr_i);
              end
       end
   //synopsys translate_on
```

and then designs would either use the one in basejump_stl or in some <bsg_design>/v/hard.  Yeah, I think I like that better.  There's code duplication, but only of the wrapper.  The "default" rams would get copied over when the v/hard directory is created

hard/gf_14/bsg_mem/bsg_mem_1rw_sync.v
```diff
@@ -78,7 +60,7 @@ module bsg_mem_1rw_sync #( parameter width_p = -1
         end // block: s1r1w
       else
         begin: notmacro
-          bsg_mem_1rw_sync_synth # (.width_p(width_p), .els_p(els_p))
+          bsg_mem_1rw_sync_synth # (.width_p(width_p), .els_p(els_p), .latch_last_read_p(latch_last_read_p))
```
dpetrisko, line 63: This parameter was missing before...

hard/gf_14/bsg_mem/bsg_mem_1rw_sync_mask_write_bit.v
```diff
@@ -73,33 +31,8 @@ module bsg_mem_1rw_sync_mask_write_bit #( parameter width_p = -1
...
-              , .EMA   ( 3'b010                                )
-              , .EMAW  ( 2'b00                                 )
-              , .EMAS  ( 1'b0                                  )
-              );
-        end: bank
-    end: macro
-  else
     begin: notmacro
-      bsg_mem_1rw_sync_mask_write_bit_synth #(.width_p(width_p), .els_p(els_p))
+      bsg_mem_1rw_sync_mask_write_bit_synth #(.width_p(width_p), .els_p(els_p), .latch_last_read_p(latch_last_read_p))
```
dpetrisko, line 35: This one was also missing

hard/gf_14/bsg_mem/bsg_mem_1rw_sync_mask_write_byte.v
```diff
@@ -74,7 +30,7 @@ module bsg_mem_1rw_sync_mask_write_byte #( parameter els_p = -1
   `bsg_mem_1rw_sync_mask_write_byte_banked_macro(1024,256,8,1) else
   // no hardened version found
     begin : notmacro
-      bsg_mem_1rw_sync_mask_write_byte_synth #(.data_width_p(data_width_p), .els_p(els_p))
+      bsg_mem_1rw_sync_mask_write_byte_synth #(.data_width_p(data_width_p), .els_p(els_p), .latch_last_read_p(latch_last_read_p))
```
dpetrisko, line 33: Same here

dpetrisko:
I updated with macro headers and hardened bodies.  I haven't ported to my design branch yet, but I think this will end up cleaner (and not require any bsg_14 changes)

hard/gf_14/bsg_mem/bsg_mem_1r1w_sync_macros.vh
```diff
@@ -0,0 +1,22 @@
+
```
tommydcjung, line 1: might want header guard.

tommydcjung (approved):

dpetrisko:
Good call! Added the guards

dpetrisko:
Tested with the gf14 flow, works well and is very clean.  PR should be good to go

stdavids:
Thanks for the feedback Tommy!

Okay so the new flow is to basically always link the *_ macro.vh files which live in basejump stl. And then to give designs control over the list of memories to harden, we copy the .v file into the target design (I like the v/hard/... idea) and link that file rather than the one in basejump stl?

dpetrisko:
yep, exactly!

stdavids:
Tested and changes to the cad flow are upstream so I am merging this commit. The scope of this PR is quite small only changing hard 14nm memories and so it is good to get this PR merged to get the cad flow and basejump working together.

If this methodology works well then we can investigate moving this idea into other hard memories like tsmc40.

taylor-bsg:
Great work everybody!

PR #358, opened 2020-12-04
Fix wormhole streamers to use passthrough_dynamic SIPO/PISO

muwyse:
This change modifies the wormhole streamer modules to use passthrough_dynamic variants of SIPO/PISO when protocol data width < link flit width. The streamer protocol client provides an expected number of data flits and then the wormhole streamer computes the number of protocol data flits per PISO/SIPO transaction.

Without this feature, the protocol client would need to send extra null data packets (SIPO) or drain null data packets (PISO) to completely fill or drain the SIPO/PISO, respectively.

~~This change allows the wormhole_stream_out module to drain invalid client data flits. This can occur when the input link flit width is greater than the output client protocol data width, if the client protocol message does not have a full link flit width of data. Without the drain, the client would be responsible for knowing the link flit width and draining extra, invalid data packets.~~

### DIFF
--- a/bsg_noc/bsg_wormhole_stream_in.v
+++ b/bsg_noc/bsg_wormhole_stream_in.v
@@ -165,7 +165,7 @@ module bsg_wormhole_stream_in
       // set value is zero based and equal to (protocol beats - 1)
       // SIPO passthrough sends same cycle it receives last data input
       logic [pr_len_width_p-1:0] pr_data_cnt;
-      wire pr_data_consumed = (pr_data_cnt == '0);
+      wire pr_data_last = (pr_data_cnt == '0);
       bsg_counter_set_down
        #(.width_p(pr_len_width_p)
          ,.init_val_p('0)
@@ -176,7 +176,7 @@ module bsg_wormhole_stream_in
          ,.reset_i(reset_i)
          ,.set_i(hdr_v_i & hdr_ready_and_o)
          ,.val_i(pr_data_beats_i)
-         ,.down_i(data_v_i & data_ready_and_o & ~pr_data_consumed)
+         ,.down_i(data_v_i & data_ready_and_o & ~pr_data_last)
          ,.count_r_o(pr_data_cnt)
          );
 
--- a/bsg_noc/bsg_wormhole_stream_in.v
+++ b/bsg_noc/bsg_wormhole_stream_in.v
@@ -68,10 +68,10 @@ module bsg_wormhole_stream_in
    , input [hdr_width_p-1:0]     hdr_i
    , input                       hdr_v_i
    , output                      hdr_ready_and_o
-   // number of protocol message data flits in arriving wormhole message
+   // number of protocol message data packets in arriving wormhole message
    // arrives late when hdr_v_i & hdr_ready_and_o
-   // value is flits-1 (i.e., zero based)
-   , input [pr_len_width_p-1:0]  pr_data_flits_i
+   // value is len-1 (i.e., zero based)
+   , input [pr_len_width_p-1:0]  pr_data_beats_i
 
    // The protocol data information
    , input [pr_data_width_p-1:0] data_i
@@ -140,10 +140,9 @@ module bsg_wormhole_stream_in
     //   protocol data, aggregate it, and then send it out on the wormhole network
     begin : narrow
       // flit_width_p > pr_data_width_p -> multiple protocol data per link flit
-      // and the protocol data may not completely fill the SIPO. Need to track how
-      // many protocol data flits remain for each SIPO transaction
+      // and the protocol data may not completely fill the SIPO.
 
-      // number of protocol data flits per full link flit
+      // number of protocol data per full link flit
       localparam [len_width_p-1:0] max_els_lp = `BSG_CDIV(flit_width_p, pr_data_width_p);
       localparam lg_max_els_lp = `BSG_SAFE_CLOG2(max_els_lp);
       // SIPO len_i is zero-based, i.e., input is len-1
@@ -153,11 +152,11 @@ module bsg_wormhole_stream_in
       logic sipo_first_lo;
       logic [lg_max_els_lp-1:0] sipo_len_li;
 
-      // count of protocol data flits to send
-      // last flit is consumed when pr_data_cnt is 0
+      // count of protocol data packets to send
+      // last data is consumed when pr_data_cnt is 0
       // set late when hdr_v_i & hdr_ready_and_o
       // set value is provided by input client
-      // set value is zero based and equal to (protocol flits - 1)
+      // set value is zero based and equal to (protocol beats - 1)
       // SIPO passthrough sends same cycle it receives last data input
       logic [pr_len_width_p-1:0] pr_data_cnt;
       wire pr_data_consumed = (pr_data_cnt == '0);
@@ -170,17 +169,15 @@ module bsg_wormhole_stream_in
         (.clk_i(clk_i)
          ,.reset_i(reset_i)
          ,.set_i(hdr_v_i & hdr_ready_and_o)
-         ,.val_i(pr_data_flits_i)
+         ,.val_i(pr_data_beats_i)
          ,.down_i(data_v_i & data_ready_and_o & ~pr_data_consumed)
          ,.count_r_o(pr_data_cnt)
          );
 
-      // for each SIPO transaction, provide number of protocol flits to expect
-      assign sipo_len_li = (data_v_i & sipo_first_lo)
-                           ? (pr_data_cnt >= sipo_full_len_lp)
-                             ? sipo_full_len_lp
-                             : lg_max_els_lp'(pr_data_cnt)
-                           : '0;
+      // for each SIPO transaction, provide number of protocol data to expect
+      assign sipo_len_li = (pr_data_cnt >= sipo_full_len_lp)
+                           ? sipo_full_len_lp
+                           : lg_max_els_lp'(pr_data_cnt);
 
       bsg_serial_in_parallel_out_passthrough_dynamic
        #(.width_p(pr_data_width_p)
--- a/bsg_noc/bsg_wormhole_stream_in.v
+++ b/bsg_noc/bsg_wormhole_stream_in.v
@@ -29,15 +29,15 @@
 //   ---------------------------------------------------------------
 //   - header flits do not contain any data
 //
-// Data can be sent the same or any cycle after header, but only 1 message at
+// Data can be sent any cycle after header, but only 1 message at
 //   a time is supported.
 // - Legal: H     H
 //            D-D   D-D
-// - Legal: H   H
-//          D-D D-D
 // - Illegal: H H
 //            D-D-D-D
 //   (Header sent too early)
+//
+// This module does not accept data until the header sends
 //
 module bsg_wormhole_stream_in
  #(// The wormhole router protocol information
@@ -90,6 +90,9 @@ module bsg_wormhole_stream_in
 
   logic [flit_width_p-1:0] hdr_lo;
   logic hdr_v_lo, hdr_ready_and_li;
+  logic hdr_v_li, hdr_ready_and_lo;
+  assign hdr_v_li = is_hdr & hdr_v_i;
+  assign hdr_ready_and_o = is_hdr & hdr_ready_and_lo;
 
   // Header is input all at once and streamed out 1 flit at a time
   bsg_parallel_in_serial_out_passthrough
@@ -101,8 +104,8 @@ module bsg_wormhole_stream_in
      ,.reset_i(reset_i)
 
      ,.data_i(hdr_i)
-     ,.v_i(hdr_v_i)
-     ,.ready_and_o(hdr_ready_and_o)
+     ,.v_i(hdr_v_li)
+     ,.ready_and_o(hdr_ready_and_lo)
 
      ,.data_o(hdr_lo)
      ,.v_o(hdr_v_lo)
@@ -112,6 +115,9 @@ module bsg_wormhole_stream_in
 
   logic [flit_width_p-1:0] data_lo;
   logic data_v_lo, data_ready_and_li;
+  logic data_v_li, data_ready_and_lo;
+  assign data_ready_and_o = is_data & data_ready_and_lo;
+  assign data_v_li = is_data & data_v_i;
 
   // Protocol data is 1 or multiple flit-sized. We accept a large protocol data
   //   and then stream out 1 flit at a time
@@ -127,8 +133,8 @@ module bsg_wormhole_stream_in
          ,.reset_i(reset_i)
 
          ,.data_i(data_i)
-         ,.v_i(data_v_i)
-         ,.ready_and_o(data_ready_and_o)
+         ,.v_i(data_v_li)
+         ,.ready_and_o(data_ready_and_lo)
 
          ,.data_o(data_lo)
          ,.v_o(data_v_lo)
@@ -188,14 +194,12 @@ module bsg_wormhole_stream_in
          ,.reset_i(reset_i)
 
          ,.data_i(data_i)
-         ,.v_i(data_v_i)
-         ,.ready_and_o(data_ready_and_o)
+         ,.v_i(data_v_li)
+         ,.ready_and_o(data_ready_and_lo)
 
          ,.first_o(sipo_first_lo)
 
-         // TODO: need sipo_len_li on data_v_i & first_o
-         // should this module provide any buffering for data arriving before
-         // header or sipo_len_li is valid?
+         // len_i must be valid when v_i & first_o
          ,.len_i(sipo_len_li)
 
          ,.data_o(data_lo)
--- a/bsg_noc/bsg_wormhole_stream_in.v
+++ b/bsg_noc/bsg_wormhole_stream_in.v
@@ -27,6 +27,7 @@
 //   ---------------------------------------------------------------
 //   | data   | data  | data  | data  | protocol info | len   cord |
 //   ---------------------------------------------------------------
+//   - header flits do not contain any data
 //
 // Data can be sent the same or any cycle after header, but only 1 message at
 //   a time is supported.
@@ -54,6 +55,7 @@ module bsg_wormhole_stream_in
    // Higher level protocol information
    , parameter pr_hdr_width_p  = "inv"
    , parameter pr_data_width_p = "inv"
+   , parameter pr_len_width_p  = "inv"
 
    // Size of the wormhole header + the protocol header. The data starts afterwards.
    // Users may set this directly rather than relying on the protocol header derived default
@@ -66,6 +68,10 @@ module bsg_wormhole_stream_in
    , input [hdr_width_p-1:0]     hdr_i
    , input                       hdr_v_i
    , output                      hdr_ready_and_o
+   // number of protocol message data flits in arriving wormhole message
+   // arrives late when hdr_v_i & hdr_ready_and_o
+   // value is flits-1 (i.e., zero based)
+   , input [pr_len_width_p-1:0]  pr_data_flits_i
 
    // The protocol data information
    , input [pr_data_width_p-1:0] data_i
@@ -83,7 +89,7 @@ module bsg_wormhole_stream_in
   localparam [len_width_p-1:0] hdr_len_lp = `BSG_CDIV(hdr_width_p, flit_width_p);
 
   logic [flit_width_p-1:0] hdr_lo;
-  logic hdr_ready_lo, hdr_v_lo, hdr_ready_and_li;
+  logic hdr_v_lo, hdr_ready_and_li;
 
   // Header is input all at once and streamed out 1 flit at a time
   bsg_parallel_in_serial_out_passthrough
@@ -96,13 +102,12 @@ module bsg_wormhole_stream_in
 
      ,.data_i(hdr_i)
      ,.v_i(hdr_v_i)
-     ,.ready_and_o(hdr_ready_lo)
+     ,.ready_and_o(hdr_ready_and_o)
 
      ,.data_o(hdr_lo)
      ,.v_o(hdr_v_lo)
      ,.ready_and_i(hdr_ready_and_li)
      );
-  assign hdr_ready_and_o = hdr_ready_lo;
   assign hdr_ready_and_li = is_hdr & link_ready_and_i;
 
   logic [flit_width_p-1:0] data_lo;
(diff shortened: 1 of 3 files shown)


bsg_noc/bsg_wormhole_stream_out.v
```diff
@@ -103,8 +106,39 @@ module bsg_wormhole_stream_out
   assign data_v_li = is_data & link_v_i;
   // Protocol data is less than a single flit-sized. We accept a large
   //   wormhole flit, then let the client process it piecemeal
-  if (flit_width_p >= pr_data_width_p)
+  if (flit_width_p > pr_data_width_p)
     begin : narrow
+      // flit_width_p > pr_data_width_p -> multiple protocol data per link flit
```
dpetrisko, line 111: Could optimize to just reset on the piso on data consumed. Degenerate case is pr_len = 1, PISO els = 1000

bsg_noc/bsg_wormhole_stream_out.v
```diff
@@ -66,6 +68,7 @@ module bsg_wormhole_stream_out
    , output [hdr_width_p-1:0]     hdr_o
    , output                       hdr_v_o
    , input                        hdr_ready_and_i
+   , input [pr_len_width_p-1:0]   data_len_i
```
dpetrisko, line 71: comment that this is a late signal, computed from hdr_o

dpetrisko (changes requested):
Can you add a test configuration to the testbench for this case?

bsg_noc/bsg_wormhole_stream_out.v
```diff
@@ -27,6 +27,7 @@
 //   ---------------------------------------------------------------
 //   | data   | data  | data  | data  | protocol info | len   cord |
 //   ---------------------------------------------------------------
+//   - header flits do not contain any data
```
dpetrisko, line 30: Add this assumption to wormhole_stream_in as well?

bsg_noc/bsg_wormhole_stream_out.v
```diff
@@ -103,8 +106,39 @@ module bsg_wormhole_stream_out
...
+  if (flit_width_p > pr_data_width_p)
     begin : narrow
+      // flit_width_p > pr_data_width_p -> multiple protocol data per link flit
+      // and it is possible that last link flit is not completely filled with valid
+      // protocol data. If this happens, need to drain protocol data from PISO in
+      // last link data flit.
+
+      // data_len_i = number of data_o flits that will be consumed by client
+      // counter is set when header is acked by client
+      logic [pr_len_width_p-1:0] pr_data_cnt;
```
dpetrisko, line 118: This might belong better in stream_control.
is_hdr
is_data
is_drain

dpetrisko (changes requested):
~Can you add a test for this configuration to the testbench?~ Duplicate comment

bsg_noc/bsg_wormhole_stream_out.v
dpetrisko: also, is this the len in bytes or the number of pr_data_beats?

dpetrisko:
Hey @gaozihou , could you take a look at this? The context is for a wormhole streamer where the protocol request size is not a divisor of the flit width.

flit_width = 32
pr_width = 8

for a request with pr_size = 24 bits, the PR side will ack three times.  But then there's still some data remaining in the PISO. We need to drain this data before the next request. @muwyse approach here actually acks each individual piece of data, I suggest resetting the piso to bulk clear it out, but either approach should work.

bsg_noc/bsg_wormhole_stream_out.v
muwyse: Agreed, a reset/flush signal to PISO is better than draining one protocol data flit at a time due to degenerate cases where link flit >> protocol data flit

bsg_noc/bsg_wormhole_stream_out.v
muwyse: Currently this is pr_data_beats that will be consumed by protocol client.

bsg_noc/bsg_wormhole_stream_out.v
muwyse: Since it is not supported in wormhole_stream_out, then perhaps, yes. I suppose it is possible to use the wormhole_stream_in module though and not have the other end connected to a wormhole_stream_out where this restriction wouldn't be needed. But, adding it makes both modules consistent.

bsg_noc/bsg_wormhole_stream_out.v
muwyse: My understanding of wormhole_stream_control is that it deals only with link packets. It doesn't know about protocol sizes/packets. The is_data_o signal means that the current link flit has data, but that could be one of the partially full flits we are considering in this PR.

That being said, we could expose the protocol info to the control module and have it implement the counter then output a drain signal after the required protocol data flits were consumed.

muwyse:
> Hey @gaozihou , could you take a look at this? The context is for a wormhole streamer where the protocol request size is not a divisor of the flit width.
> 
> flit_width = 32
> pr_width = 8
> 
> for a request with pr_size = 24 bits, the PR side will ack three times. But then there's still some data remaining in the PISO. We need to drain this data before the next request. @muwyse approach here actually acks each individual piece of data, I suggest resetting the piso to bulk clear it out, but either approach should work.

I agree that resetting the PISO is better than individual acks. I will try to code this up.

bsg_noc/bsg_wormhole_stream_out.v
muwyse: modified both stream in and out to flush (i.e., trigger reset on the SIPO or PISO)

bsg_noc/bsg_wormhole_stream_out.v
muwyse: added and hopefully clarified naming

muwyse:
> Can you add a test configuration to the testbench for this case?

Yes, still need to do testbench stuff, including fixing up existing for new protocol data beats/length input.

bsg_noc/bsg_wormhole_stream_out.v
muwyse: Added. Can relax in both modules later if desired, but it will add complexity.

bsg_noc/bsg_wormhole_stream_out.v
dpetrisko: Yeah, I think exposing is fine since stream_control is only going to be used in this module, which knows about both protocols

dpetrisko:
> Yeah, I think exposing is fine since stream_control is only going to be used in this module, which knows about both protocols

The other alternative here is to make parallel_in_serial_out_passthrough_dynamic and serial_in_parallel_out_passthrough_dynamic variants, which is essentially what we're hacking around here

gaozihou:
1. The proposed interface looks nice! I was convinced that `pr_data_len_i` is better than something like `last_word_i` signal, because it does not require counter outside streamer and it is easier to use. I also like the user-friendly setup that `pr_data_len_i` comes together with `hdr_v_i / hdr_v_o`.
My only comment on the interface is, would it be better to make `pr_dat_len_i == actual_len-1`? This will match the definition in wormhole network / dynamic piso / sipo... It also requires less bits to represent (like `actual_len == 8` case), and eliminate potential bugs due to setting `pr_len_width_p` incorrectly.
(We don't need to take care of the `actual_data_len == 0` case, because in that case streamer-controller will keep `is_data = 0`)

2. In terms of implementation details, I would vote for @dpetrisko's dynamic passthrough piso / sipo. First reason is that it reuses the counter inside piso / sipo, no need to add extra counter in streamer. Second reason is that the implementation is a little easier, like pseudo code below. Third reason (may not be a reason) is that piso / sipo passthrough is not used anywhere else, we may take the vantage to tweak it as much as we need...
```
// In stream_in
      bsg_serial_in_parallel_out_passthrough_dynamic
       #(.width_p(pr_data_width_p)
         ,.els_p(data_len_lp)
         )
       data_sipo
        (.clk_i(clk_i)
         ,.reset_i(reset_i)

         ,.data_i(data_i)
         // is_last_data already exists in stream control module, need to expose to outside world
         // pr_data_len_i is registered as pr_data_len_r
         ,.len_i(is_last_data? (pr_data_len_r % data_len_lp) : data_len_lp)
         ,.v_i(data_v_i)
         ,.ready_and_o(data_ready_and_o)

         ,.data_o(data_lo)
         ,.v_o(data_v_lo)
         ,.ready_and_i(data_ready_and_li)
         );

// in stream_out
      bsg_parallel_in_serial_out_passthrough_dynamic
       #(.width_p(pr_data_width_p)
         ,.els_p(data_len_lp)
         )
       data_piso
        (.clk_i(clk_i)
         ,.reset_i(reset_i)

         ,.data_i(link_data_i)
         ,.v_i(data_v_li)
         ,.len_i(is_last_data? (pr_data_len_r % data_len_lp) : data_len_lp)
         ,.ready_and_o(data_ready_lo)

         ,.data_o(data_o)
         ,.v_o(data_v_o)
         ,.ready_and_i(data_ready_and_i)
         );
```

dpetrisko:
#363

muwyse:
@dpetrisko, does #363 replace this PR?

The changes to the wormhole streamers will be to use the passthrough_dynamic SIPO/PISO?

dpetrisko:
I would rebase on #363 and replace with the passthrough_dynamic, the toplevel interface change stays the same

gaozihou:
Just merged #363 into wormhole_streamer branch, should be okay to rebase now

bsg_noc/bsg_wormhole_stream_in.v
```diff
@@ -148,12 +194,20 @@ module bsg_wormhole_stream_in
          ,.v_i(data_v_i)
          ,.ready_and_o(data_ready_and_o)
 
+         ,.first_o(sipo_first_lo)
+
+         // TODO: need sipo_len_li on data_v_i & first_o
+         // should this module provide any buffering for data arriving before
+         // header or sipo_len_li is valid?
+         ,.len_i(sipo_len_li)
```
muwyse, line 202: @gaozihou @dpetrisko Thoughts on this? With the current interface, the SIPO cannot accept data until the len_i input is known, which is expected to be derived from the protocol header. It is up to the protocol client to obey this, and the earliest data can be presented is the cycle after hdr_v_i & hdr_ready_and_o, since SIPO len_i is derived from the registered value of the protocol data counter.

dpetrisko: Why does SIPO len have to be computed from the registered value? Why can't it be same cycle?

bsg_noc/bsg_wormhole_stream_in.v
```diff
@@ -66,6 +68,10 @@ module bsg_wormhole_stream_in
    , input [hdr_width_p-1:0]     hdr_i
    , input                       hdr_v_i
    , output                      hdr_ready_and_o
+   // number of protocol message data flits in arriving wormhole message
+   // arrives late when hdr_v_i & hdr_ready_and_o
+   // value is flits-1 (i.e., zero based)
+   , input [pr_len_width_p-1:0]  pr_data_flits_i
```
dpetrisko, line 74: nit, but I would call this beats instead of flits, since the protocol is more of a bus than a network

bsg_noc/bsg_wormhole_stream_out.v
```diff
@@ -103,12 +109,51 @@ module bsg_wormhole_stream_out
...
+         ,.down_i(data_v_o & data_ready_and_i & ~pr_data_consumed)
+         ,.count_r_o(pr_data_cnt)
+         );
+
+      // for each PISO transaction, provide number of protocol flits to expect
+      assign piso_len_li = (data_ready_and_i & piso_first_lo)
+                           ? (pr_data_cnt >= piso_full_len_lp)
+                             ? piso_full_len_lp
+                             : lg_max_els_lp'(pr_data_cnt)
+                           : '0;
```
dpetrisko, line 152: Could you explain the '0 case here?

bsg_noc/bsg_wormhole_stream_in.v
```diff
@@ -129,16 +134,57 @@ module bsg_wormhole_stream_in
...
+         ,.down_i(data_v_i & data_ready_and_o & ~pr_data_consumed)
+         ,.count_r_o(pr_data_cnt)
+         );
+
+      // for each SIPO transaction, provide number of protocol flits to expect
+      assign sipo_len_li = (data_v_i & sipo_first_lo)
+                           ? (pr_data_cnt >= sipo_full_len_lp)
+                             ? sipo_full_len_lp
+                             : lg_max_els_lp'(pr_data_cnt)
+                           : '0;
```
dpetrisko, line 183: Could you explain the '0 case here?

bsg_noc/bsg_wormhole_stream_out.v
muwyse: It is not needed, will remove. I just dislike non-zero or hard to determine default values. The top condition is the condition where len_i is captured.

bsg_noc/bsg_wormhole_stream_in.v
muwyse: Same as above.

bsg_noc/bsg_wormhole_stream_in.v
muwyse: done

bsg_noc/bsg_wormhole_stream_in.v
muwyse: It could, but requires more logic to determine if the len_i should come from the register or bypass it and come directly from pr_data_beats_i.

I see this as the same top level question: what is the interface/contract between the protocol client and the stream_in module?

dpetrisko: I think that extra logic is fine. The computation length is on the order of 10-bit adders, which are not too big. Since wormhole streamers will be connected between a wormhole router and another module, they often will be going through a CDC, which causes the input path to be very short.  That said, the comment saying that not too much computation should be done on the length is important.

Is protocol beats the right input here? We must be computing that outside of the module, since the protocol information comes in log2(bytes)?

Another possibility, we save pr_len_in_bytes = 1 << pr_len_in_log_bytes and subtract pr_width_in_bytes? This is also only a 7-bit counter for 1024 bytes, so not too bad for typical usecases.

Just thinking out loud here though, I'm not sure what the most best interface is. I believe most buses use burst_size=num_beats like we have here

gaozihou: @muwyse I guess the hdr->data ordering requirement could be enforced by controlling the hdr_v_i / hdr_ready_o / data_v_i / data_ready_o? When counter value is 0 we accept hdr, when non-zero we accept data..

gaozihou: @dpetrisko I guess bypassing the counter value does not break the hdr->data ordering completely? Data flits still cannot arrive earlier than hdr... We still need to enforce data_ready_o to maintain the ordering.

By definition of ready-and interface, ready_o should not depend on value of v_i. When bypassing counter value, we are creating a loop from hdr_v_i to data_ready_o, which may not be favorable?

dpetrisko: What would you recommend @gaozihou? My thought is that the dependency isn't ideal, but is technically compliant since it's two separate interfaces.

The alternative is that we don't enforce the hdr->data ordering.

So there are kind of two paths:
- We merge the version with ordering enforcement, we find a critical path, and then we make a change possibly loosening this constraint
- We merge the version with no ordering enforcement, modules are written that depend on sending data before header, then we change the interface and these new modules break

So the bypass is probably the best bad option...

muwyse: Clarification: if sipo_len_li can be set in same cycle as hdr_v_i & hdr_ready_and_o, the comparison in the assignment changes to use pr_data_beats_i instead of pr_data_cnt. I think it is sensible to allow that bypass of the counter.

I also think it is reasonable to require the message length to be known before the data SIPO accepts any data. Otherwise we need a different interface on the SIPO since it currently relies on knowing the len_i input when v_i & first_o are both raised (i.e., on the first data input to SIPO).

Another option is to add a buffer in the wormhole_stream_in module of configurable (including zero) size for data packets, which would be up to max_els_p entries. This buffer could accept data packets from the client as soon as they arrive (and even bypass if the buffer is empty) and its output connects to the SIPO. The SIPO would wait to accept data until the len_i input is known good, and then the buffer starts draining to the SIPO.

My main question in the ordering not enforced case is how does the SIPO handle the partial full case when data arrives before it knows len_i?

I would say we require hdr->data ordering at first, and can loosen later if it is a big need.

dpetrisko: >Clarification: if sipo_len_li can be set in same cycle as hdr_v_i & hdr_ready_and_o, the comparison in the assignment changes to use pr_data_beats_i instead of pr_data_cnt. I think it is sensible to allow that bypass of the counter.

Yeah let's do this in the interest of getting something in. This is the minimal buffering solution, and my gut feeling is the critical path will be fine in all current use-cases

> Another option is to add a buffer in the wormhole_stream_in module of configurable (including zero) size for data packets, which would be up to max_els_p entries. This buffer could accept data packets from the client as soon as they arrive (and even bypass if the buffer is empty) and its output connects to the SIPO. The SIPO would wait to accept data until the len_i input is known good, and then the buffer starts draining to the SIPO.

Maybe add this as a comment block for how to support hdr / data independence

dpetrisko:
@muwyse @gaozihou Is this good to go or are there open questions?

gaozihou:
@dpetrisko I guess the hdr->data ordering requirement could be enforced by controlling the hdr_v_i / hdr_ready_o / data_v_i / data_ready_o? When counter value is 0 we accept hdr, when non-zero we accept data..

For example:
When len_r == 0, we accept hdr_in (hdr_ready_o = 1 & data_ready_o = 0 & hdr_v_i = v_i & data_v_i = 0); 
When len_r != 0, we accpet data_in (hdr_ready_o = 0 & data_ready_o = 1 & hdr_v_i = 0 & data_v_i = v_i). 
In this way we still have the hdr->data dependency, but it will never introduce bug when user forget about the requirement..

dpetrisko:
> When len_r == 0, we accept hdr_in (hdr_ready_o = 1 & data_ready_o = 0 & hdr_v_i = v_i & data_v_i = 0);
> When len_r != 0, we accpet data_in (hdr_ready_o = 0 & data_ready_o = 1 & hdr_v_i = 0 & data_v_i = v_i).
> In this way we still have the hdr->data dependency, but it will never introduce bug when user forget about the ?> requirement..

I think this is a good solution.  In the current iteration, all wormhole packet formats have headers in seperate flits than data. So a bubble on the data port will not be a true performance overhead. Is that right?

muwyse:
Added gates on hdr/data input handshakes using is_hdr and is_data signals for stream in to enforce hdr->data ordering

dpetrisko (approved):
LGTM!

dpetrisko:
@gaozihou , look good to you?

gaozihou:
@dpetrisko @muwyse Sorry for the delay...

It looks good to me, just two minor things:
1. I would suggest using pr_data_consumed and ~pr_data_consumed instead of is_hdr and is_data. They may not be equivalent, depending on the implementation detail of passthrough PISO and SIPO. There can be the case that is_hdr==0 while pr_data_consumed==1. is_hdr and is_data is on the "inner" side of PISO and SIPO, while pr_data_consumed is on the "outside" where hdr_v_i, data_v_i, hdr_ready_o and data_ready_o live in.
2. It would be safer to add similar logic to stream_out on hdr_v_o, data_v_o, hdr_ready_i and data_ready_i. (Current implementation relies on the fact that passthrough PISO does not have any internal buffer, so doing without this logic is fine)

gaozihou (approved):
I approve the changes, with only minor issues being addressed

muwyse:
> 1. I would suggest using pr_data_consumed and ~pr_data_consumed instead of is_hdr and is_data. They may not be equivalent, depending on the implementation detail of passthrough PISO and SIPO. There can be the case that is_hdr==0 while pr_data_consumed==1. is_hdr and is_data is on the "inner" side of PISO and SIPO, while pr_data_consumed is on the "outside" where hdr_v_i, data_v_i, hdr_ready_o and data_ready_o live in.

First, a clarification: In stream_in, pr_data_consumed should really be called pr_data_last in the current code (I just made this change and pushed). The last protocol data beat is consumed when the count is 0 (pr_data_last == 1) and the data_v/ready handshake occurs. This signal is only used for decrementing the protocol data beats counter, to avoid a decrement when count == 0. The protocol data beats counter is used to supply the length input to the SIPO.

You are referring to the gating of the header and data valid/ready signals? Yes, pr_data_consumed (pr_data_last) can be high when is_hdr == 0. This occurs when the last protocol data beat is ready on the input, but the link does not accept the output. If the suggestion is to say that hdr_v_li = pr_data_last & hdr_v_i, I think this is incorrect. We need to finish consuming/sending data before switching to header. This would also mismatch with the stream control unit.

Could you explain more what the specific problem could be? For passthrough SIPO, my understanding is that by definition the last serial input is consumed in the same cycle it is sent as part of the output. I think passthrough for the PISO just means that the input parallel data is not buffered, also by definition. So what implementation details could change where using is_hdr/is_data can cause a problem, and would those changes be a different passthrough SIPO/PISO, or would it be a change from a passthrough to non-passthrough SIPO/PISO? I don't think we need to support the latter.

gaozihou:
Oh, I just noticed the counter is counting down from len-1 to 0 (not from len to 0), then you are right pr_data_consumed is not a proper signal to use, otherwise it will cause problem.

In terms of the specific problem, for our current dynamic passthrough SIPO, "the last serial input is consumed in the same cycle it is sent as part of the output" does not hold when a single-flit packet is coming. This is because the dynamic SIPO needs that len_i signal to know whether this is the "last serial input", which creates a loop path from c_i -> len_i -> ready_and_o, violating the definition of ready_and interface. Therefore, single-flit packet going in will be consumed even though it is not sent out in same cycle.

Our current passthrough PISO, passthrough SIPO and dynamic passthrough PISO are all fine. The issue on dynamic passthrough SIPO is not a big deal because it is only used in stream_in data_i port. The reason why I mentioned this is because in general I would not mix the "wormhole-side" logic signals with "parallel-side" logic signals (I discussed this with Dan a while ago), and since I though there was a ready-to-use pr_data_consumed and it would be a one-line change... Yeah if we count down from len-1 to 0 then using is_data and is_hdr is more efficient.

gaozihou:
I guess the definition of passthrough might be tricky. In my opinion "passthrough" is more like data_o shows up in same cycle as data_i updates, and there is a "passthrough" path from input side to output side... 

For example, our first version passthrough PISO did have buffering inside, so data_i is consumed at the first cycle, sending out first flit data_o in same cycle and register all others. I guess this is also an implementation of "passthrough"?

muwyse:
The code of the passthrough dynamic SIPO looks like it does output a single flit transaction on the same cycle it enters. https://github.com/muwyse/basejump_stl/blob/78b4e2d13fb3c5e51c17068e904f4a7399fcd225/bsg_dataflow/bsg_serial_in_parallel_out_passthrough_dynamic.v#L120

That being said, if the output link side is not ready, I think the problem could appear, but I actually think if it did it would be a violation of the stream_in input side handshaking with the protocol client. The requirement is that the protocol header is consumed and then in following cycles data is consumed. So, the protocol client should not be providing data after the single input packet handshakes (otherwise the protocol message is not single flit, or the client is erroneously sending data).

The SIPO will hold the single data packet until the link output is not ready, but the protocol client won't raise data_v_i again because it has already sent the single data input. The next protocol transaction must send the header first, before any new data can be sent.

So, yes, the data_ready_o signal will be raised (we could call this erroneously), but the protocol client is violating the interface if it sends more data.

Maybe the solution is to add a set/clear dff (call it pr_data_consumed) indicating when the protocol data has all been consumed by the SIPO. This is set when (data_v_i & data_ready_and_o & pr_data_last), and it is cleared on (is_data & link_v_o & link_ready_and_i & pr_data_consumed). data_ready_and_o then is gated with ~pr_data_consumed. I think hdr gating can continue to use is_hdr.

dpetrisko:
Would it be appropriate to add an assertion to validate the necessary behavior of the passthrough PISO? I think it would be something like asserting that v_i == v_o always?

gaozihou:
Thanks @muwyse! The good news is, in dynamic passthrough SIPO, when it holds the single-flit data in its buffer (in waiting state), ready_and_o is always set to 0 (https://github.com/muwyse/basejump_stl/blob/78b4e2d13fb3c5e51c17068e904f4a7399fcd225/bsg_dataflow/bsg_serial_in_parallel_out_passthrough_dynamic.v#L124), so user would not accidentally stream in more data flits, which means current implementation is totally fine. 

I am fine with merging current version since it works totally fine. 

Yeah the reason why I though it was counting down from len to 0 was because of the "no data" special case. **Currently the pr_data_beats_i is represented by beats-1, so we are expecting random pr_data_beats_i value from user in the no-data case?** (I know it does not really matter because is_data will never be asserted). If we set counter to beats-0, then pr_data_last==(counter==1) and pr_data_consumed==(counter==0), and the special case is automatically addressed.

dpetrisko:
Cool, merging and then we can take a look at the original streamer PR. Thanks @gaozihou and @muwyse

PR #359, opened 2020-12-07
Adding hold on valid to rr_arb

dpetrisko:
This adds a hold on valid mode to the round robin arbiter, where it will not switch its grant unless acked.

### DIFF
--- a/bsg_misc/bsg_round_robin_arb.py
+++ b/bsg_misc/bsg_round_robin_arb.py
@@ -222,7 +222,10 @@ else
         last_n = reset_on_sr? (inputs_p-2) :
                ( yumi_i     ?tag_o : last_r );  
       end else if( hold_on_valid_p ) begin: hold_on_last_n_gen
-        last_n = yumi_i ? tag_o : v_o ? (tag_o == '0) ? (lg_inputs_p)'(inputs_p-1) : (tag_o-1'b1) : last_r;
+        // Need to manually handle wrap around on non-power of two case, else reuse subtraction
+        last_n = yumi_i ? tag_o
+               : v_o ? ((~`BSG_IS_POW2(inputs_p) && tag_o == '0) ? (lg_inputs_p)'(inputs_p-1) : (tag_o-1'b1))
+                     : last_r;
       end else
         last_n = (yumi_i ? tag_o:last_r);
 
--- a/bsg_misc/bsg_round_robin_arb.py
+++ b/bsg_misc/bsg_round_robin_arb.py
@@ -127,6 +127,13 @@ print """module bsg_round_robin_arb #(inputs_p      = %s
                                      ,lg_inputs_p   =`BSG_SAFE_CLOG2(inputs_p)
                                      ,reset_on_sr_p = 1'b0
                                      ,hold_on_sr_p  = 1'b0
+                                     // Hold on valid sets the arbitration policy such that once
+                                     // a output tag is selected, it remains selected until it is
+                                     // acked. This is consistent with BaseJump STL handshake
+                                     // assumptions. Notably, this parameter is required to work
+                                     // with bsg_parallel_in_serial_out_passthrough. This policy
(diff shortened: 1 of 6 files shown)


bsg_misc/bsg_round_robin_arb.py
```diff
@@ -213,6 +214,8 @@ def print_reset_on_logic(reqs_w):
       end else if( reset_on_sr_p ) begin: reset_on_last_n_gen
         last_n = reset_on_sr? (inputs_p-2) :
                ( yumi_i     ?tag_o : last_r );  
+      end else if( hold_on_valid_p ) begin: hold_on_last_n_gen
+        last_n = yumi_i ? tag_o : v_o ? (tag_o-1'b1) : last_r;
```
gaozihou, line 218: I guess `tag_o-1'b1` may not be valid when inputs_p is non-pow2? For example, when `inputs_p==5` and `tag_o==3'b000`, then `tag_o-1'b1==3'b111`, and the state machine enters default state ,making `sel_one_hot_n= {5{1'bx}}`... Maybe we should do the following?
```
last_n = yumi_i ? tag_o : v_o ? ((tag_o == '0)? (lg_inputs_p)'(inputs_p-1) : (tag_o-1'b1)) : last_r;
```
This error is not triggered in the testbench, because there is not data injected into router from proc, so tag_o is never set to 0...

gaozihou (changes requested):

bsg_misc/bsg_round_robin_arb.py
dpetrisko: Good catch!

dpetrisko: Fixed!

gaozihou (approved):

taylor-bsg:
comments that explain the new functionality, limitations/assumptions, and interactions with piso passthrough

dpetrisko:
>comments that explain the new functionality, limitations/assumptions, and interactions with piso passthrough

I would actually argue that hold_on_valid should be the default behavior, and doesn't need special explanation. As we discussed, we're now taking the stance that all modules in basejump_stl must not drop valid once risen. If that's the case, we should add this tag-1'b1 clause to the other arbitration modes and not have a hold_on_valid parameter at all.

taylor-bsg:
I have a simple rule .. if you spend more than 10 seconds thinking about
it, it deserves a comment saying what you thought about. How long did we
spend?? :-)

taylor-bsg:
No specific objections but changing the default kind of mandates a PPA
sweep. Also the arbiter is used in other places besides routers, something
to think about...

dpetrisko:
Added comment block and created issue #362 for further investigation!

dpetrisko:
@taylor-bsg Good to merge?

dpetrisko:
@taylor-bsg

dpetrisko:
@taylor-bsg

bsg_misc/bsg_round_robin_arb.v
```diff
@@ -2428,6 +2437,8 @@ else
       end else if( reset_on_sr_p ) begin: reset_on_last_n_gen
         last_n = reset_on_sr? (inputs_p-2) :
                ( yumi_i     ?tag_o : last_r );  
+      end else if( hold_on_valid_p ) begin: hold_on_last_n_gen
+        last_n = yumi_i ? tag_o : v_o ? (tag_o == '0) ? (lg_inputs_p)'(inputs_p-1) : (tag_o-1'b1) : last_r;
```
taylor-bsg, line 2441: - add parens to clarify operator ordering
- add comment
// potential optimizations: specialize power of 2 case

taylor-bsg (approved):

bsg_misc/bsg_round_robin_arb.v
dpetrisko: fixed the formatting! I just tossed the optimization in there too, since it's really simple

PR #361, opened 2020-12-11
Dynamic PISO reduce internal buffering

gaozihou:
Reduce internal buffering from `2*els_p` to `els_p+1`, at the cost of extra combinational logic

Not tested yet

### DIFF
--- a/bsg_dataflow/bsg_parallel_in_serial_out_dynamic.v
+++ b/bsg_dataflow/bsg_parallel_in_serial_out_dynamic.v
@@ -8,21 +8,21 @@
 `include "bsg_defines.v"
 
 module bsg_parallel_in_serial_out_dynamic
-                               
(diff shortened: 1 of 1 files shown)


dpetrisko:
High level, but I’m curious if hard coding the length of a dynamic PISO would synthesize to the same as the static PISO...

gaozihou:
@dpetrisko Interesting question! I think it will be very very similar.
I am not sure about the following expression:
```
data_o = (go_fifo_v_lo & (count_lo==max_els_p-1))? fifo_data_lo[max_els_p-1] : fifo_data_lo[count_lo];
```
Will the tool simplify it to the following?
```
data_o = fifo_data_lo[count_lo]
```

dpetrisko:
LGTM once it's tested

PR #364, opened 2020-12-14
Adding type definition for bsg_dramsim3_pkg

dpetrisko:
This is a scoping issue that I believe is technically undefined behavior in the LRM, since the enum is potentially undefined at elaboration unless it happens to elaborate in the correct order. Adding the type allows it to compile with latest Verilator.


tommydcjung (approved):

dpetrisko:
good to merge?

drichmond:
I think so

tommydcjung:
okay I tested in vcs. so I will just get this merged.

PR #371, opened 2020-12-23
DRAMSim3 moves macros from source to header

### LINKED ISSUE #368: macro defines in package
https://github.com/bespoke-silicon-group/basejump_stl/blob/d4cd62042e6f83b953a6bad75c2257f10ce0c853/bsg_test/bsg_dramsim3_pkg.v#L86

This means that the correct operation of this module will depend on its include order on the VCS command line. We should factor out into a .vh file and include appropriately.

mrutt92:
Fixes #368

### DIFF
--- a/bsg_test/bsg_dramsim3.vh
+++ b/bsg_test/bsg_dramsim3.vh
@@ -1,0 +1,30 @@
+`ifndef BSG_DRAMSIM3_VH
+`define BSG_DRAMSIM3_VH
+
+`define dramsim3_ba_width(num_ba_mp) $clog2(num_ba_mp)
+`define dramsim3_bg_width(num_bg_mp) $clog2(num_bg_mp)
+`define dramsim3_co_width(num_columns_mp) $clog2(num_columns_mp)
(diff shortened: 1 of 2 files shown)


taylor-bsg:
@mrutt92 will this break anything when I merge it?

mrutt92:
Nope.

dpetrisko (approved):

dpetrisko:
I think it requires minor changes in BlackParrot. Everywhere bsg_dramsim3_pkg is currently used needs an `include bsg_dramsim3.vh now. Not an objection to the merge, I wrote a note to remind myself when we bump basejump.  (I think bsg_manycore uses global includes, so as long as this is added to the makefile flist, should be fine)

mrutt92:
What's the status on this?

farzamgl (approved):
Everything should work with minimal change on BlackParrot, so there's no problem in my view.

mrutt92:
@taylor-bsg I don't have permissions to merge this but it has been approved by the BP folks.

PR #374, opened 2020-12-31
Adding cache block data width support to bsg_cache

Songchun-Li:
This PR aims at fixing [#372](https://github.com/bespoke-silicon-group/basejump_stl/issues/372) 

The change is minimal and merely about block offset of address.

PR #377, opened 2021-01-08
Fixing the bugs for LD/SD when data_width_p > 64

Songchun-Li:
## Motivation
To implement BP Stream Protocol between L1 and L2, we would like to make L2 cache slice have parameterizable data width, ranging from dword_width_p (64 bits) to cce_block_width_p (512 bits) with 2^n. However, there is a bug in the store buffer write mask for D (8 bytes) size when data_width_p is larger than 64 btis.

## Problem Description
When there is a SD operation, we expect data_mem_w_mask_li to be `8'hff` to write 64 bits, or 8 bytes data into the data_mem, like the following screenshot where the data_width_p==64.
image

And when the data_width_p==128,  the data_mem_w_mask_li should be either  `16'h00ff` or `16'hff00` to write the lower half or the upper half of the 128 bits for a single SD op. However, the data_mem_w_mask_li becomes `16'hffff` and overwrites the other half mistakenly.
image

## Root Cause
Tracing the signals backward, the mask signal comes from sbuf_mask_in_mux_li. And there are at most 4 elements in sbuf_mask_in_mux_li, which are for W (1 byte), H(2 bytes), W(4 bytes), D(8 bytes) from index 0 to 3 correspondingly. Therefore, the focus is on `sbuf_mask_in_mux_li[data_sel_mux_els_lp-1]`, the max_size clause. (data_sel_mux_els_lp == 4 for 64 bits)
```
    if (i == data_sel_mux_els_lp-1) begin: max_size
      assign sbuf_mask_in_mux_li[i] = {data_mask_width_lp{1'b1}};    
    end 
    else begin: non_max_size

      logic [(data_width_p/slice_width_lp)-1:0] decode_lo;

      bsg_decode #(
        .num_out_p(data_width_p/slice_width_lp)
      ) dec (
        .i(addr_v_r[i+:`BSG_MAX(lg_data_mask_width_lp-i,1)])
        ,.o(decode_lo)
      );

      bsg_expand_bitmask #(
        .in_width_p(data_width_p/slice_width_lp)
        ,.expand_p(2**i)
      ) exp (
        .i(decode_lo)
        ,.o(sbuf_mask_in_mux_li[i])
      );

    end
```
The max_size clause optimizes out the decoding and expanding logic based on data_width_p used to be less than or equal to 64 bits for the max_size, and just sets all the bits for simplicity. For example, for data_width_p==64, max_size is D, then data_mask_width_lp ==(64>>3)==8, so we have the mask to be 8'hff.  For data_width_p==32, max_size is W, data_mask_width_lp ==(32>>3)==4, so we have the mask to be 4'hf.
However, when data_width_p==128, the new case we want to support, data_mask_width_lp==(128>>3)==16 and setting all the 16 bits to 1, while the max_size is still D, so the data_mem_w_mask_li of 16'hffff gives a bug.

## Solution
A simple solution is to add additional condition for the data_width_p for the max_size clauses
` if (i == data_sel_mux_els_lp-1 && data_width_p <= 64) begin: max_size`
to disable the max_size clause when data_width_p is larger than 64.
The expected sbuf_mask_in should be like this:
image

### DIFF
--- a/bsg_cache/bsg_cache.v
+++ b/bsg_cache/bsg_cache.v
@@ -681,7 +681,7 @@ module bsg_cache
 
     assign sbuf_data_in_mux_li[i] = {(data_width_p/slice_width_lp){slice_data}};
 
-    if (i == data_sel_mux_els_lp-1) begin: max_size
+    if (i == data_sel_mux_els_lp-1 && data_width_p <= 64) begin: max_size
       assign sbuf_mask_in_mux_li[i] = {data_mask_width_lp{1'b1}};    
     end 
     else begin: non_max_size
--- a/bsg_cache/bsg_cache.v
+++ b/bsg_cache/bsg_cache.v
@@ -681,29 +681,23 @@ module bsg_cache
 
     assign sbuf_data_in_mux_li[i] = {(data_width_p/slice_width_lp){slice_data}};
 
-    if (i == data_sel_mux_els_lp-1 && data_width_p <= 64) begin: max_size
-      assign sbuf_mask_in_mux_li[i] = {data_mask_width_lp{1'b1}};    
-    end 
-    else begin: non_max_size
+    logic [(data_width_p/slice_width_lp)-1:0] decode_lo;
 
-      logic [(data_width_p/slice_width_lp)-1:0] decode_lo;
+    bsg_decode #(
+      .num_out_p(data_width_p/slice_width_lp)
+    ) dec (
+      .i(addr_v_r[i+:`BSG_MAX(lg_data_mask_width_lp-i,1)])
+      ,.o(decode_lo)
+    );
 
-      bsg_decode #(
-        .num_out_p(data_width_p/slice_width_lp)
-      ) dec (
-        .i(addr_v_r[i+:`BSG_MAX(lg_data_mask_width_lp-i,1)])
-        ,.o(decode_lo)
-      );
+    bsg_expand_bitmask #(
+      .in_width_p(data_width_p/slice_width_lp)
+      ,.expand_p(2**i)
+    ) exp (
+      .i(decode_lo)
+      ,.o(sbuf_mask_in_mux_li[i])
+    );
 
-      bsg_expand_bitmask #(
-        .in_width_p(data_width_p/slice_width_lp)
-        ,.expand_p(2**i)
-      ) exp (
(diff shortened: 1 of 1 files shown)


dpetrisko (approved):
LGTM, but @tommydcjung may have a preference on implementation

tommydcjung:
I think that part of code was written before I fixed bsg_decode to support num_out_p==1, would it work if you get rid of max_size section?

Songchun-Li:
Yep, removing the max_size section also works

tommydcjung:
can you add a small test that replicates this bug as a part of regression?

taylor-bsg:
And modify pull request so that the diff does not show code that was not actually modified?

Songchun-Li:
> And modify pull request so that the diff does not show code that was not actually modified?

It seems github does not support hiding whitespace changes by default, and we have to check this box manually for each review.
image

Source: https://github.blog/2018-05-01-ignore-white-space-in-code-review/

taylor-bsg:
Yes, this is why I am asking you to modify your pull request so that you
are not changing whitespace for lines of code you did not modify

Songchun-Li:
Got it. The unnecessary whitespace changes are reverted.

Songchun-Li:
While working on the test, I spot another similar bug in [ld_data_final_li](https://github.com/bespoke-silicon-group/basejump_stl/blob/959aedfe1414453f5578419cb4019bee3873387b/bsg_cache/bsg_cache.v#L780) for LD when data_width>64.

When data_width_p >= 64, there are 4 elements in sbuf_mask_in_mux_li, which are for W (1 byte), H(2 bytes), W(4 bytes), D(8 bytes) from index 0 to 3 correspondingly. And LD uses the 3rd one in the max_size section. The max_size section optimizes out the decoding and padding logic and works perfectly when data_width_p==64.

However, if data_width_p>64, for example, data_width_p==128, the dword indicates by the address can be either the upper 64 bits or the lower 64 bits. And the target dword should also be mux-ed, put to the lowest 64 bits and paddding out the remaining higher bits, just like what is done in the else-section.

Therefore, I remove the if-section like how we solve the similar bug in this PR.

Songchun-Li:
The changes in this PR is moved to #374, for the convenience of reviewing updates in testing. Closing this one.

PR #376, opened 2021-01-07
VCS 2020 Variable initialization related fix

ptpan:
There is an open discussion [here](https://github.com/bespoke-silicon-group/bsg_replicant/issues/638) about the simulation error using VCS 2020, and this PR attempts to fix that issue by initializing the clock generator output to 1. This should work with the async reset generator (triggered by the negative edge of input clock) to correctly assert the reset signal from simulation zero.

PR #378, opened 2021-01-08
Switch bsg_nonsynth_reset_gen input/output to bit instead of (implicit) logic to spurious pos/neg edges at T=0

drichmond:
Switched input and output of nonsynth_reset_gen to bit (instead of logic)

In newer versions of VCS, casting between logic and bit types triggers new positive/negative edges at T=0 when 'x' -> '1/0'. 

To solve this (and be backward compatible) this PR changes the type of `clk_i` and `async_reset_o` to bit so that users can use bits in the instantiating module and avoid negative/positive edges, or use logic and maintain the old behavior.

Replaces #376

### DIFF
--- a/bsg_test/bsg_nonsynth_reset_gen.v
+++ b/bsg_test/bsg_nonsynth_reset_gen.v
@@ -20,6 +20,14 @@ module bsg_nonsynth_reset_gen #(parameter num_clocks_p=1
 
    genvar i;
 
+   // This module relies on the input clock wires starting at 0 at
+   // time 0 and not transitioning from X to 0 at time 0, and causing
+   // a spurious negedge. To accomplish this in > VCS 2020, the input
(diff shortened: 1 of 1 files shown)


dpetrisko (approved):

taylor-bsg:
Add a comment explaining why input bit is used.

// This module relies on the input clock wires starting at 0 at time 0
// and not transitioning from X to 0 at time 0, causing a spurious negedge
// to accomplish this in > VCS 2020, we have to be sure that the input is declare as bit.
// Moreover, the expectation is that this module's input comes from a clock generation
// module that uses bit, such as bsg_nonsynth_clk_gen

drichmond:
OK! That comment?

taylor-bsg:
yes

drichmond:
Done (with small modifications)

taylor-bsg (approved):

PR #379, opened 2021-01-09
Switch to assert final for bsg_circular_ptr, to satisfy VCS 2020

drichmond:
When switching to VCS 2020, from VCS 2019 this assertion starts to fire. Whatever causes it does not seem to change the correctness, but it does cause error messages.

`assert final`, which waits until signals have stabilized, fixes the problem

Happy to add comments, etc.


drichmond:
This passes on all of the replicant tests, with both VCS and Verilator

dpetrisko (approved):

taylor-bsg:
Brilliant .. this will have to become a best practice in our BSG SystemVerilog doc...

PR #384, opened 2021-01-14
Fix tag trace replay, tag data set to low when not sending

gaozihou:
tag_data_o should set to 1'b0 when not sending (when valid_o == 0).

This module was working before, because the previous version PISO kept data_o=last_bit_sent (which was 1'b0) when valid_o=0. The new version PISO keeps data_o=first_bit_sent (which is 1'b1), exposing the issue.


tommydcjung:
I think this clarifies the interface more. bsg_tag_master does not have v_i port so it just takes whatever data presented. maybe v_o and yumi_i on the bsg_tag_trace_replay is kinda unnecessary. we don't want the data_o to be stuck at 1'b1, because you want to stream enough 0s to bring the bsg_tag_master out of eStuck state.

stdavids (approved):

PR #387, opened 2021-01-21
Fixing the write burst control logic in bsg_dmc_controller

farzamgl:
This PR fixes a handshaking issue in the write data and mask pipeline in bsg_dmc_controller.
When processing multiple consecutive write commands, we have to check if the write PISOs are ready to accept new data before issuing the write command and dequeuing it's {data, mask} pair from the input SIPO.

### DIFF
--- a/bsg_dmc/bsg_dmc_controller.v
+++ b/bsg_dmc/bsg_dmc_controller.v
@@ -140,8 +140,6 @@ module bsg_dmc_controller
 
   logic        cwd_valid;
   logic  [7:0] cwd_tick;
-  logic        wburst_valid;
-  logic  [7:0] wburst_tick;
 
   logic        cas_valid;
   logic  [7:0] cas_tick;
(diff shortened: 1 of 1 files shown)


bsg_dmc/bsg_dmc_controller.v
```diff
@@ -622,11 +605,12 @@ module bsg_dmc_controller
       tx_mask_piso_data_li[k] <= tx_mask[k*dfi_mask_width_lp+:dfi_mask_width_lp];
     end
   end
+  always_ff @(posedge dfi_clk_i) begin
+    tx_data_piso_valid_li <= (tx_sipo_yumi_cnt_li == ui_burst_length_lp);
+    tx_mask_piso_valid_li <= (tx_sipo_yumi_cnt_li == ui_burst_length_lp);
+  end
```
hema0730, line 611: I think it's more accurate to assign "shoot & (n_cmd==WRITE)" to valid_i signals.

farzamgl: Is it functionally different or it's just a clarification change?

hema0730: Just to make valid signal more strict to avoid being asserted accidentally.

akashnt4:
@dpetrisko , @farzamgl , @taylor-bsg ,

Resurrecting this thread. 
Relevant scenario at the unit level would be to keep writing to the same row so that there are no ACTIVATE/ PRECHARGE command in between write accesses (PISO will be fed and accessed continuously) . But I'm unable to see any overwriting happening at the unit while trying out the said scenario.

Would you be able to help me with the history and if possible, wave dump for a failing case? Thanks!

gaozihou:
Hi @akashnt4, I think bsg_dmc is relying on a specific behavior of PSIO. The PISO will not assert ready_o until the previous packet finishes sending (usually takes a few cycles), so although it seems like "PISO will be fed and accessed continuously", only the first packet gets accepted (when valid_i==1 and ready_o==1).

@farzamgl found this behavior after #330 get merged in Oct 2020. In order to make bsg_dma work I merged #350 in March 2021, so it should be working correctly now.

dpetrisko:
Hey @gaozihou I don’t fully understand the issue, but could we run into problems if els_p == 1? If we rely on there being a gap between consecutive ready_o?

gaozihou:
@dpetrisko If els_p==1, then the PISO will instantiate bsg_two_fifo internally, so it will be fine.

gaozihou:
@dpetrisko to summarize the issue, currently bsg_dmc is not checking PISO's ready_o signal. It knows ready_o==1'b1 because the buffer (fifo) size is large enough. However, instead of asserting v_i for single cycle, it asserts v_i for els_p number of cycles, where els_p is the DRAM burst length. This means the PSIO has to de-assert ready_o for (els_p-1) number of cycles to be compatible with it. For a PISO the (els_p-1) gap will not result in bubble on the other side because it takes els_p cycles to stream out anyway.

This is not an ideal behavior though...

dpetrisko:
Should we close this then?

taylor-bsg:
I think it's still a bug .. it is asserting valid to the tx_piso even when
it does not have new data to send?

@gaozihou @akashnt4

does it suffice to replace this

else if(wburst_valid) begin
wburst_tick <= wburst_tick - 1;
if(wburst_tick == 0) wburst_valid <= 0;
end

with:

 else wburst_valid <= 0;

gaozihou:
@taylor-bsg Yeah I think this would work, or more generically:
```
else if (tx_data_piso_ready_lo)
    wburst_valid <= 0;
```
I think for this DMC, tx_data_piso_ready_lo and tx_mask_piso_ready_lo should have same value all the time, so we could check any of them.

taylor-bsg:
I think the logic would not work correctly if ready were stalled, the data
would fall off the end. So maybe instead an assert that
tx_data_piso_ready_lo is high if wburst_valid is high? And we still clear
valid?

gaozihou:
Yeah I agree it's better to
1. clear valid, and
2. add assertions on tx_data_piso_ready_lo and tx_mask_piso_ready_lo, as long as the PISOs are not pass-through (I think it is a reasonable assumption).

akashnt4:
This PR is redundant as required updates are part of #584 . Can be closed once #584 is mainline.

taylor-bsg:
| 20 | PR [#387](https://github.com/bespoke-silicon-group/basejump_stl/pull/387) — Fixing the write burst control logic in bsg_dmc_controller | Close — superseded by merged #584 | The author explicitly requested closure after #584 landed; it merged in November 2024. Current write-burst valid is cleared after the burst. Keep #746 for the separate CALR bank-selection bug. [PR #584](https://github.com/bespoke-silicon-group/basejump_stl/pull/584); [PR #746](https://github.com/bespoke-silicon-group/basejump_stl/pull/746). |

PR #388, opened 2021-01-22
[WIP] Add Optional Encoding for BSG Link

gaozihou:
Fully functional, need to add comments...

This is the synthesizable version bsg_link, will have an FPGA version downstream module (which has dual clk_i ports).


bsg_link/bsg_link_ddr_downstream.v
```diff
@@ -49,8 +49,10 @@ module bsg_link_ddr_downstream
   // Set use_extra_data_bit_p=1 to utilize this extra bit
```
taylor-bsg, line 49: // add comments explaining encoding scheme and also motivation

taylor-bsg: // also an explaination of the support for channel bonding

taylor-bsg (commented):
Perhaps ready should be pulled low if reset is asserted? Just thinking aloud here.

gaozihou:
Yeah prof I think this is an interesting point. I pulled ready_o high in order to match the behavior of bsg_two_fifo (and many other modules in bsg_dataflow). When reset is asserted, ready_o is 1'b1, so it accepts incoming packets (and ignore them). v_o is 1'b0 so no outgoing packets.

The good thing is that outputs do not change when coming out of reset, reducing transient power consumption to the whole chip... It also consumes garbage packets (if any) during reset
The downside is that if we de-assert bsg_link reset after de-assert core logic reset, packets may get dropped...

tommydcjung (approved):

gaozihou:
Three main components in this PR:
1. Adding new files for SDR link (no harm)
2. Modifying existing files for DDR link encode (interface and default behavior do not change)
3. Adding testbenches (no harm)

PR #393, opened 2021-02-02
Removing Dividend and Divisor Registers from Iterative Division Module

barnold3:
@dpetrisko @taylor-bsg change has been verified


taylor-bsg:
Awesome! Can you minimize the number of white space changes so that git is only showing substantive differences? This is a great help.

barnold3:
Should be clearer now

barnold3:
Should be clearer now @taylor-bsg

bsg_misc/bsg_idiv_iterative.v
```diff
@@ -50,57 +53,51 @@ module bsg_idiv_iterative #(parameter width_p=32, parameter bitstack_p=0)
...
    //if the divisor is zero
-   wire         zero_divisor_li   =  ~(| divisor_r);
+   wire         zero_divisor_li   =  ~(| opA);
+   
+   wire [width_p:0]  add_out;
+   wire [width_p:0]  opA_mux_in_0;
+   wire              latch_msb_divisor;
 
+   assign opA_mux_in_0[width_p-1:0] = add_out[width_p-1:0];
+   assign opA_mux_in_0[width_p] = (latch_msb_divisor & ~signed_div_r) ? 1'b0 : add_out[width_p];
```
taylor-bsg, line 73: instead of a ? 0 : b, use

b & ~a

(see also below)

bsg_misc/bsg_idiv_iterative.v
```diff
@@ -50,57 +53,51 @@ module bsg_idiv_iterative #(parameter width_p=32, parameter bitstack_p=0)
...
-       ,.clk_i(clk_i)
-       );
-
    //if the divisor is zero
-   wire         zero_divisor_li   =  ~(| divisor_r);
+   wire         zero_divisor_li   =  ~(| opA);
+   
+   wire [width_p:0]  add_out;
+   wire [width_p:0]  opA_mux_in_0;
+   wire              latch_msb_divisor;
```
taylor-bsg, line 70: output of locally instantiated module should be suffixed _lo (applies to the three added control signals) per BaseJump Standard Verilog guidelines

bsg_misc/bsg_idiv_iterative_controller.v
```diff
@@ -100,23 +102,33 @@ module bsg_idiv_iterative_controller #(parameter width_p=32)
...
 
     case (state)
 
     WAIT: begin
-       if (v_i) next_state = START;
-       latch_inputs_o = 1'b1;
+       if (v_i) begin
+          opA_ld_o     = 1'b1;
+          opC_ld_o     = 1'b1;
+          opA_sel_o    = 1'b1;
```
taylor-bsg, line 116: Seems like the two sel_o blocks can be pulled out of the if (v_i) statement to reduce potential critical paths.

bsg_misc/bsg_idiv_iterative_controller.v
```diff
@@ -100,23 +102,33 @@ module bsg_idiv_iterative_controller #(parameter width_p=32)
...
+          opC_sel_o    = 3'b100;
+          next_state   = START;
+       end
+       latch_signed_div_o = 1'b1;
```
taylor-bsg, line 120: conversely, since latch_signed_div_o changes state inside the module, it probably should be inside the if (v_i) block.

bsg_misc/bsg_idiv_iterative.v
```diff
@@ -17,6 +17,9 @@
 // 39           negate remainder (if necessary)
 // 40           negate quotient (if necessary)
 // -------------------------------------------
+//
+// Schematic: https://docs.google.com/presentation/d/1F7Lam7fMCp-v9K1PsjTvypWHJFFXfqoX6pJrmgf-_JE/
+//
```
taylor-bsg, line 22: output parameter ready_o should be renamed ready_and_o, to be consistent with our attempts to clarify interfaces that are ready_and versus ready_then (also in BSG SystemVerilog style guide)

bsg_misc/bsg_idiv_iterative.v
```diff
@@ -50,57 +53,51 @@ module bsg_idiv_iterative #(parameter width_p=32, parameter bitstack_p=0)
    assign quotient_o = opC[width_p-1:0];
 
    wire         signed_div_r;
```
taylor-bsg, line 55: Deletion of divisor_msb and dividend_msb seem like an unnecessary code change that percolates through the rest of the code? Instead I would take the _i versions as input, a deminimis amount of logic delay.

I will take the code that I think can be reverted with ITEM_1

bsg_misc/bsg_idiv_iterative.v
```diff
@@ -50,57 +53,51 @@ module bsg_idiv_iterative #(parameter width_p=32, parameter bitstack_p=0)
...
+   assign opA_mux_in_0[width_p] = (latch_msb_divisor & ~signed_div_r) ? 1'b0 : add_out[width_p];
+   
    wire         opA_sel;
    wire [width_p:0]  opA_mux;
-   wire [width_p:0]  add_out;
    bsg_mux  #(.width_p(width_p+1), .els_p(2)) muxA
-       (.data_i({ {divisor_msb, divisor_r}, add_out } )
-       ,.data_o(opA_mux)
-       ,.sel_i(opA_sel)
+          ( .data_i( {{divisor_i[width_p-1], divisor_i}, opA_mux_in_0} )
```
taylor-bsg, line 78: ITEM_1

bsg_misc/bsg_idiv_iterative.v
```diff
@@ -50,57 +53,51 @@ module bsg_idiv_iterative #(parameter width_p=32, parameter bitstack_p=0)
...
-
    //if the divisor is zero
-   wire         zero_divisor_li   =  ~(| divisor_r);
+   wire         zero_divisor_li   =  ~(| opA);
+   
+   wire [width_p:0]  add_out;
+   wire [width_p:0]  opA_mux_in_0;
+   wire              latch_msb_divisor;
 
+   assign opA_mux_in_0[width_p-1:0] = add_out[width_p-1:0];
```
taylor-bsg, line 72: ITEM_1

bsg_misc/bsg_idiv_iterative.v
```diff
@@ -50,57 +53,51 @@ module bsg_idiv_iterative #(parameter width_p=32, parameter bitstack_p=0)
...
-       ,.sel_i(opA_sel)
+          ( .data_i( {{divisor_i[width_p-1], divisor_i}, opA_mux_in_0} )
+           ,.data_o(opA_mux)
+           ,.sel_i(opA_sel)
      );
 
+   wire [width_p:0]  opB_mux_in_0;
+   wire              latch_msb_dividend;
+   
+   assign opB_mux_in_0[width_p:1] = add_out[width_p-1:0];
```
taylor-bsg, line 86: ITEM_1

bsg_misc/bsg_idiv_iterative_controller.v
```diff
@@ -100,23 +102,33 @@ module bsg_idiv_iterative_controller #(parameter width_p=32)
       opC_ld_o       = 1'b1;
       adder_cin_o    = !add_neg_last;
       neg_ld         = 1'b0;
-      latch_inputs_o = 1'b0;
+      latch_msb_divisor_o  = 1'b0;
+      latch_msb_dividend_o = 1'b0;
```
taylor-bsg, line 106: variable is always 0?

bsg_misc/bsg_idiv_iterative.v
```diff
@@ -50,57 +53,51 @@ module bsg_idiv_iterative #(parameter width_p=32, parameter bitstack_p=0)
...
+          ( .data_i( {opC, add_out, opB_mux_in_0} )
            ,.data_o(  opB_mux )
            ,.sel_one_hot_i(opB_sel)
      );
 
    wire [2:0]   opC_sel;
    wire [width_p:0]  opC_mux;
    bsg_mux_one_hot #(.width_p(width_p+1), .els_p(3)) muxC
-          ( .data_i( {{dividend_msb, dividend_r},add_out, {opC[width_p-1:0], ~add_out[width_p]}} )
+          ( .data_i( {{dividend_i[width_p - 1], dividend_i}, add_out, {opC[width_p-1:0], ~add_out[width_p]}} )
```
taylor-bsg, line 100: ITEM_1

bsg_misc/bsg_idiv_iterative.v
```diff
@@ -50,57 +53,51 @@ module bsg_idiv_iterative #(parameter width_p=32, parameter bitstack_p=0)
    assign quotient_o = opC[width_p-1:0];
 
    wire         signed_div_r;
-   wire [width_p-1:0]  dividend_r;
-   wire [width_p-1:0]  divisor_r;
-   wire divisor_msb  = signed_div_r & divisor_r[width_p-1];
-   wire dividend_msb = signed_div_r & dividend_r[width_p-1];
 
-   wire latch_inputs;
+   wire latch_signed_div;
```
taylor-bsg, line 57: signals that are outputs of locally instantiated modules should end in _lo

bsg_misc/bsg_idiv_iterative.v
```diff
@@ -211,7 +208,9 @@ module bsg_idiv_iterative #(parameter width_p=32, parameter bitstack_p=0)
       ,.opC_sel_o                (opC_sel)
       ,.opC_ld_o                 (opC_ld)
 
-      ,.latch_inputs_o           (latch_inputs)
+      ,.latch_signed_div_o       (latch_signed_div)
+      ,.latch_msb_divisor_o      (latch_msb_divisor)
```
taylor-bsg, line 212: probably don't need the _msb outputs (ITEM_1)

bsg_misc/bsg_idiv_iterative_controller.v
taylor-bsg: ITEM_1

bsg_misc/bsg_idiv_iterative_controller.v
```diff
@@ -100,23 +102,33 @@ module bsg_idiv_iterative_controller #(parameter width_p=32)
...
     START: begin
        next_state = NEG0;
        opA_ld_o     = 1'b1;
-       opC_ld_o     = 1'b1;
-       opA_sel_o    = 1'b1;
-       opC_sel_o    = 3'b100;
+       opA_inv_o    = 1'b0;
+       opC_ld_o     = 1'b0;
+       opB_clr_l_o  = 1'b0;
+       latch_msb_divisor_o = 1'b1;
```
taylor-bsg, line 129: latch_msb_divisor_o (ITEM_1)

bsg_misc/bsg_idiv_iterative_controller.v
```diff
@@ -142,6 +154,7 @@ module bsg_idiv_iterative_controller #(parameter width_p=32)
 
     SHIFT: begin
        next_state   = CALC;
+       latch_msb_divisor_o = 1'b1;
```
taylor-bsg, line 157: latch_msb_divisor_o (ITEM_1)

taylor-bsg (changes requested):
Nice initial work! See feedback. After this I think we should talk about testbenches. My suggestion is to dump every possible input/output pairs of the original for a 4-bit divider, review the results, and then use that file as a testbench input for your revised version. The testbenches are located in testing/bsg_misc/bsg_idiv/...

bsg_misc/bsg_idiv_iterative_controller.v
barnold3: BUG

@taylor-bsg . This is a bug. I suppose the testbench is not robust enough. Latch_msb_dividend is meant to be 1 in the SHIFT state not the latch_msb_divisor. The case in which this bug would crop up is when the second MSB is a 1 and extended to the MSB/sign bit in the first cycle. It should then be fixed in the SHIFT state if unsigned division is being performed.

bsg_misc/bsg_idiv_iterative_controller.v
barnold3: BUG - see above

bsg_misc/bsg_idiv_iterative.v
barnold3: My thinking here was that if we implemented it as you say then the _msb would depend on an unregistered/unlatched value. In other words, if the _msb expression is used in a later cycle it might not be accurate if the divisor/dividend inputs have changed while the module is still iterating on the previous inputs.

bsg_misc/bsg_idiv_iterative.v
barnold3: Not sure I understand how these would be cut. The "a" in your ITEM_1 uses the _msb signal. And each of the _msb signals is high in one state and low in all others (see also BUG).

bsg_misc/bsg_idiv_iterative.v
taylor-bsg: Yes, but I believe that the _msb values are only used in the WAIT stage, and are used to calculate the values which go through muxA/muxC and are flopped in regA/regC. So at that point, all of the side effects of _msb wires have been registered and the _msb wires are no longer needed.

bsg_misc/bsg_idiv_iterative_controller.v
taylor-bsg: Yep I saw it -- but this will all go away with the ITEM_1 adjustment

bsg_misc/bsg_idiv_iterative.v
barnold3: I understand now. I think where I got off making this more complicated than it needed to be was thinking that we needed to wait for signed_div input to be registered. We can just do assign opA_mux_in_0[width_p] = add_out[width_p] & ~signed_div_i. *Face in palm*

barnold3: Thanks for the input and bearing with me. Easy fix.

barnold3:
@taylor-bsg 
1) Implemented a simpler hardware modification
2) Fixed signal names to follow style guidelines
3) Modified testbench to check every possible input combination for a 4-bit instantiation of the divider
4) Verified functionality with new testbench

P.S. Substantive changes to bsg_idiv_iterative.v can be found at lines 56, 57, 68, 74, 90 of the new file (other changes are fixing signal names)

bsg_misc/bsg_idiv_iterative.v
```diff
@@ -49,91 +52,75 @@ module bsg_idiv_iterative #(parameter width_p=32, parameter bitstack_p=0)
...
+   wire divisor_msb  = signed_div_i & divisor_i[width_p-1];
+   wire dividend_msb = signed_div_i & dividend_i[width_p-1];
 
-   wire latch_inputs;
+   wire latch_signed_div_lo;
    bsg_dff_en#(.width_p(1)) req_reg
        (.data_i (signed_div_i)
-       ,.data_o (signed_div_r)
-       ,.en_i   (latch_inputs)
+       ,.data_o (signed_div_r_li)
```
dpetrisko, line 62: So this is a register signal, just signed_div_r is fine

bsg_misc/bsg_idiv_iterative.v
```diff
@@ -49,91 +52,75 @@ module bsg_idiv_iterative #(parameter width_p=32, parameter bitstack_p=0)
...
-   wire [width_p-1:0]  dividend_r;
-   wire [width_p-1:0]  divisor_r;
-   wire divisor_msb  = signed_div_r & divisor_r[width_p-1];
-   wire dividend_msb = signed_div_r & dividend_r[width_p-1];
+   wire         signed_div_r_li;
+   wire divisor_msb  = signed_div_i & divisor_i[width_p-1];
+   wire dividend_msb = signed_div_i & dividend_i[width_p-1];
 
-   wire latch_inputs;
+   wire latch_signed_div_lo;
```
dpetrisko, line 59: prefer logic

bsg_misc/bsg_idiv_iterative.v
taylor-bsg: @barnold3

bsg_misc/bsg_idiv_iterative.v
```diff
@@ -50,90 +53,74 @@ module bsg_idiv_iterative #(parameter width_p=32, parameter bitstack_p=0)
...
            ,.data_o(  opC_mux )
-           ,.sel_one_hot_i(opC_sel)
+           ,.sel_one_hot_i(opC_sel_lo)
      );
 
-   wire opA_ld;
+   wire opA_ld_lo;
    bsg_dff_en#(.width_p(width_p+1)) opA_reg
        (.data_i (opA_mux)
        ,.data_o (opA    )
```
taylor-bsg, line 98: opA_r

bsg_misc/bsg_idiv_iterative.v
```diff
@@ -50,90 +53,74 @@ module bsg_idiv_iterative #(parameter width_p=32, parameter bitstack_p=0)
...
-       ,.en_i   (opB_ld )
+       ,.en_i   (opB_ld_lo )
        ,.clk_i(clk_i)
        );
 
-   wire opC_ld;
+   wire opC_ld_lo;
    bsg_dff_en#(.width_p(width_p+1)) opC_reg
        (.data_i (opC_mux)
        ,.data_o (opC    )
```
taylor-bsg, line 115: opC_r

bsg_misc/bsg_idiv_iterative.v
```diff
@@ -50,90 +53,74 @@ module bsg_idiv_iterative #(parameter width_p=32, parameter bitstack_p=0)
...
+       ,.en_i   (opA_ld_lo )
        ,.clk_i(clk_i)
        );
  
-   wire         opB_ld;
+   wire         opB_ld_lo;
    wire [width_p:0]  opB;
    bsg_dff_en#(.width_p(width_p+1)) opB_reg
        (.data_i (opB_mux)
        ,.data_o (opB    )
```
taylor-bsg, line 107: opB_r

testing/bsg_misc/bsg_idiv_iterative/test_bsg.v
```diff
@@ -5,10 +5,10 @@
 // Code refactored based on Sam Larserk's work
```
taylor-bsg, line 5: The goal here is not to replace the original test, which in some sense reduces the coverage that we have, but to add a new one. My suggestion is that you explicitly write down for the 4-bit case, in a text file (one for signed and one for unsigned), 3 values per line, the dividend, the divisor, and what the correct value of the result should be (based on the C code I pointed to), and if it is an corner case, indicate with a comment that this is a RISC-V corner case and what the answer should be and why, rather than relying on the shaky checking implementation in this testbench, which for example has an if statement that just skips some checks for the signed case. Then you can have a testbench that generates all possible 2^(2W) inputs for width W, runs it through the model and prints the output. In the makefile, you can grep out the comments you wrote in the checking file, and use UNIX diff to compare to the testbench output.

testing/bsg_misc/bsg_idiv_iterative/divide_4.stim
```diff
@@ -0,0 +1,256 @@
+0 0
```
taylor-bsg, line 1: I believe we said this file should have the hand-checked inputs and outputs side by side, and a comment that mentions that fact? This way, we can just review this file brute force by hand and have confidence that there is not a bug in our checking code.

barnold3: So the signed.c and unsigned.c programs takes in an input file (i.e. the file you pointed to. and it contains all the possible dividend and divisor outputs for the divider width you want to test) and then produces the file that you describe. Then, the output of the simulation is checked against said file. So a client need only produce a file with all the inputs they want to test and then running make and then make verify will create a file with the expected output (in the format you described) and then check the differences with the simulation output file (also in the format you described).

barnold3: I can easily write a README to clarify this.

barnold3: The idea behind the way I set it up is that the user does not need to know the expected output of their desired input cases, it can be produced for them. Additionally, I commented the RISC-V standards that define the corner cases in the C code that produces the file with expected output.

taylor-bsg:
Yes, but let’s have a single file with both inputs and expected output as I
described so that somebody can check this just by reading on GitHub and not
rely on the nuances of GCC configuration.

dpetrisko:
@taylor-bsg, you just mean for the 4 case right? We don’t want giant 64-bit stims here I don’t think

taylor-bsg:
Yes for 4

barnold3:
How's that? The 4-bit expected file lives in the repo now, and make verify will just compare the output with that file. A user still has the option to tinker with parameters and produce other such files for different divider widths etc.

testing/bsg_misc/bsg_idiv_iterative/unsigned.c
```diff
@@ -0,0 +1,59 @@
...
+  if (b == 0) { // RISC-V 7.2: Quotient of division by zero has all bits set 
+    if (WIDTH == 64) {
+      return ULLONG_MAX;
+    } else {
+      return (pow(2, WIDTH) - 1);
+    }
+  } else { // Return integer division result when not a RISC-V edge case
+    return a / b;
+  }
+  printf("hey\n");
```
dpetrisko, line 23: :P

dpetrisko (approved):
LGTM!

taylor-bsg:
@tommydcjung FYI this is 20% smaller and probably more energy efficient as well.

tommydcjung:
Can someone rebase this on master?

dpetrisko:
@tommydcjung done

tommydcjung:
https://github.com/bespoke-silicon-group/bsg_manycore/pull/513

taylor-bsg (approved):

taylor-bsg (approved):

taylor-bsg:
Congrats Brian!

PR #395, opened 2021-02-03
Adding deq id to reorder fifo

dpetrisko:
This change lets us associate the dequeued data with some external data storage, which may be written at a different time than the fifo data


tommydcjung (approved):

taylor-bsg (commented):
Maybe a comment that says how this output can be useful? Presumably so that other associated metadata with a request can be tracked in a separate register file?

PR #398, opened 2021-02-10
Improved Backup LRU Choice to Avoid Unfair Cache Replacement with Cache Line Locking

Songchun-Li:
### Motivation
In current bsg_cache design, tree pseudo-LRU algorithm is adopted to keep track of the least recently used (LRU) block in the n-way set-associative cache. And it is possible that the chosen LRU way obtained from the algorithm is locked when there is one or more ways in the cache is locked. To handle this issue, the cache picks the backup replecemant way obtained from priority encode of all unlocked ways. However, an unfair replacement may take place and worsen the performance of the cache in this design.

### Problem description
In order to prove this issue , a new test module, `lru_stats` is added. This module binds with bsg_cache_miss and records its each replecement choice. Correspondingly, 3 test trace scripts, test_lock1, test_lock2 and test_lock_multiset are added. 
1. test_lock1: This test simulate the real time opperating system where we want to lock a specific line in the cache for a certain time response in the run time. In this case, `ALOCK` op is used to lock the block and also update the LRU info. An it is allowed to load/store the locked block after it is on locked. All the accessed cache lines are at the index 0.
2. test_lock2: This test simulate the case that we want to abort a specific way at a specific index due to fabrication defects on the SRAM block. `TAGST` op is used at initialized and the locked way is never accessed. All the accessed cache lines are at the index 0.
3. test_lock_multiset: This is a variation of test_lock2 where all set are accessed rather than a single set.

All these 3 tests have 2 different work modes
- Always missed mode (default): Access 32 blocks with tags from 0 to 31 at a set sequentially, so that there is always a cache miss and LRU updates are handled by the bsg_cache_miss. This is a special case because we want to spot the issue among the current replacement strategy.
- Random access mode: Randomly access 32 blocks with tags from 0 to 31 at a set. LRU can also updated by load/store hit, so it is more like a general senario.

The lru choice counts from test_lock2 in always missed mode is attached. A severely unfair cache block replacement may take place when one of the way is locked, depending on which way is locked. By checking the waveform, this phenomenon results from "LRU trap". The LRU bits are trapped at or intended to fall back to the locked way, and therefore the backup way 0 is picked as replacement choice.
Screen Shot 2021-02-09 at 4 38 31 PM

Eventhough test_lock1 and test_lock2 are simulating different senario with cache line locking, their lru replacement statistics are consist. Raw data of test_lock1, test_lock2 in 2 modes for baseline design is [here](https://docs.google.com/spreadsheets/d/1pRIa84TJgovIjaRpI8qSxMGkcl8IDtBQRPtb27zNouI/edit?usp=sharing) in black.

### Root Cause of LRU Trap
If we focus on the data above, it can be easily noticed that the result can be categorized into 3 types based on the number of ways that was rarely accessed.
Screen Shot 2021-02-09 at 3 15 12 PM

And given the backup way is always 0 from priority encode of unlocked ways, excepted the locked way is 0, when we map these categories onto the tree, we get this diagram.
Screen Shot 2021-02-09 at 2 15 50 PM

If we takes the number of node traversed between nodes as distance, above visualization reveals that more ways are rarely accessed due to LRU trap, as the distance between backup way and the locked way is larger. 

From the encoding prespective of the tree pseudo-LRU algorithm, it is beacuse the backup way does not update the LRU bits effectively. For example, when the locked way is 7,  LRU bits are `1zz z1z1` indicating the LRU way is 7, we take the backup way 0. It will update the LRU bits with `--- 1-11`. However, the LRU bits are still `1zz z1z1` after the update, and indicates the LRU way is 7. Then, a LRU trap forms. 

And the same explaination applys for way 2 and 3, whose LRU bits are `zz0 zz10`  and `zz1 zz10` correspondingly. With the a update of backup way 0, only the LSB is updated, it is very possible that it gets back to the same locked way when the LSB is flipped back to 0 after another LRU update.

**In summary, The reason of the getting into LRU trap is that the current backup replacement choice does not update the LRU bits effectively, because the distance between the backup way and the locked way on the tree is too large and they use very different LRU bits.**

Besides, load/store hits help mitigate the LRU trap issue, so random access mode has a little better performance than always miss mode in both tests.

### Solution
**With the above analysis, a simple but efficient solution is to update the more LRU bits of the locked way as possible to avoid LRU trap**. With this tree encoding, the closest sibiling of the locked way from the same parents seems to be the best choice, because it always flips 2 LRU bits in the case of 8 way set associative.

And this can be implemented in hardware easily, the `neighbor_lru_way` is obtained by flipping the LSB of `lru_way_id`. In addition, the current backup_lru_way from prioity encode is kept as the worst choice if the `neighbor_lru_way` is also locked. 

```
assign neighbor_lru_way_id = {lru_way_id[1+:lg_ways_lp-1], ~lru_way_id[0]};

chosen_way_n = invalid_exist
  ? invalid_way_id
  : (lock_v_i[lru_way_id]
    ? (lock_v_i[neighbor_lru_way_id] ? backup_lru_way_id : neighbor_lru_way_id) 
    : lru_way_id);
```

The data in green is the result obtained with the `neighbor_lru_way` as a improved backup replacement choice. It's still a little unfair because the closeset sibiling is always picked as the backup replacement way, but this solution solve the LRU trap effectively with minimum hardware overhead added.


taylor-bsg:
Quick question — what is the green data you refer to? I see green highlighting in the table above but it seems like in that table there is an issue if any of sets 5..8 are locked?

Songchun-Li:
> Quick question — what is the green data you refer to? I see green highlighting in the table above but it seems like in that table there is an issue if any of sets 5..8 are locked?

There are 2 sheets for test_lock1 and test_lock2 correspondingly. The sheet 3 was used as draft(already deleted now).

Screen Shot 2021-02-09 at 10 53 39 PM

And in each sheet, there are two sections, the left section for random access mode and the right one for always miss mode. In each section, the data in black color is obtained from current design as baseline. And the data in green color is obtined with the proposed solution, which should have better uniformity than the baseline.

I double check the data of the proposed solution(in green color), and did not find any issues. Given there is only one way locked, the neighbor sibiling from the same parent node is always the backup LRU choice, so its expected lru hits should be higher than the others (Ideally about 2 times of the others). And I've updated the permission of the link, so it would be helpful if you can label it out.

taylor-bsg (commented):
@tommydcjung can you give a pointer to your backup LRU module and it’s usage and Songchun can run a comparison versus his approach?

tommydcjung:
https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_misc/bsg_lru_pseudo_tree_backup.v
It's used in
https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_cache/bsg_cache_non_blocking_mhu.v#L242

Songchun-Li:
@taylor-bsg The idea of `neighbor backup lru` is the same as the one in `bsg_lru_pseudo_tree_backup`, trying to take the closest way as backup, but `neighbor backup lru` is a **simplifying version** because it is applicable when there's only 1 locked way in a set. 

And all the existing test only locks one of the 8 ways in a set, so these 2 methods have **equivalent performance** in the tests. (The data of `bsg_lru_pseudo_tree_backup` is added to the raw data table in blue color.)

If there are more than one way are locked, `bsg_lru_pseudo_tree_backup` should have a better performance, especially in the situation of 2 leaf nodes of the same parent are locked. In this case, `neighbor backup lru` is invalid and the backup LRU choice falls back to priority encode of unlock ways. I will add a new test locking both leaf nodes of the same parent to validate this situation soon.

Songchun-Li:
@taylor-bsg , @tommydcjung 
The data of locking two consecutive ways in a set is added to the [test_lock_multiway sheet](https://docs.google.com/spreadsheets/d/1pRIa84TJgovIjaRpI8qSxMGkcl8IDtBQRPtb27zNouI/edit#gid=1136518952), and the result is the same as it is expected.

If the 2 lock ways are not the 2 leaf nodes of the same parent, `neighbor backup lru` and `bsg_lru_pseudo_tree_backup` has the same performance, like when way 1 and 2, or way 3 and 4 are locked.  And `bsg_lru_pseudo_tree_backup` is better when the 2 lock ways are the children of the same parent node, like when way 0 and 1, or way 2 and 3 are locked. 

In addition, If the distance between the lock ways and the priority encode of unlocked ways is larger, the advantage of `bsg_lru_pseudo_tree_backup` is more significant.

tommydcjung:
I think we want to get rid of the priority encoder (backup_lru_pe)? How is it in terms of delay? I have seen this path begin on the critical path.

Songchun-Li:
> the priority encoder (backup_lru_pe)? How is it in terms of delay? I have seen this path being on the critical path.

Is there any guide to run STA analysis on the bsg_cache? Or I should use the sweep.sh in regression_v2 to test it with varying dma delay?

dpetrisko:
@tommydcjung do you have a bsg_design for the cache?

taylor-bsg:
@tommydcjung can you kick off a synthesis run on GF12 with this for the vcache to examine whether there is a significant PPA delta? Songchun could do 45, but that would not answer the real question.

tommydcjung:
For low associativity (way=4), the impact is very minimal. backup_lru module has a shorter path than the priority encoder.

taylor-bsg (commented):
@tommydcjung can you review

bsg_cache/bsg_cache_miss.v
```diff
@@ -186,14 +186,33 @@ module bsg_cache_miss
   // backup LRU
   // A slightly crude implementation. read the design doc for the better possibility.
```
tommydcjung, line 187: Please update comment about revised replacement policy

testing/bsg_cache/regression_v2/sv.include
```diff
@@ -14,6 +14,8 @@ $BASEJUMP_STL_DIR/bsg_misc/bsg_decode.v
 $BASEJUMP_STL_DIR/bsg_misc/bsg_priority_encode.v
 $BASEJUMP_STL_DIR/bsg_misc/bsg_lru_pseudo_tree_encode.v
 $BASEJUMP_STL_DIR/bsg_misc/bsg_lru_pseudo_tree_decode.v
+$BASEJUMP_STL_DIR/bsg_misc/bsg_lru_pseudo_tree_backup.v
```
tommydcjung, line 17: filelist for "testing/bsg_cache/regression" should also be updated?

testing/bsg_cache/lock_test/testbench.v
```diff
@@ -0,0 +1,242 @@
...
+
+  logic [rom_addr_width_lp-1:0] trace_rom_addr;
+  logic [ring_width_lp+4-1:0] trace_rom_data;
+
+  logic tr_v_lo;
+  logic [ring_width_lp-1:0] tr_data_lo;
+  logic tr_yumi_li;
+  logic done;
+
+  bsg_fsb_node_trace_replay #(
```
tommydcjung, line 157: Use this instead: https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_test/bsg_trace_replay.v

testing/bsg_cache/lock_test/basic_checker.v
```diff
@@ -0,0 +1,174 @@
+module basic_checker
```
tommydcjung, line 1: Is this the exact same copy from from regression_v2? maybe we can move this to the common directory?

bsg_cache/bsg_cache_miss.v
tommydcjung: And explain what are some of the gotchas (e.g. there should be at least one way unlocked in a set, etc)

tommydcjung:
which vcs version are you using? can you add the option '-assert svaext'?
```
Error-[SE] Syntax error
  Following verilog source has syntax error :
  	Using deferred assertion requires an option -assert svaext
  "/mnt/users/ssd1/homes/dcjung/bsg/basejump_stl/bsg_misc/bsg_circular_ptr.v",
  72: token is 'final'
                assert final( (ptr_n < slots_p) || (|ptr_n === 'X) || reset_i 
  || (add_i > slots_p))
```

Songchun-Li:
@tommydcjung  Recent commits resolves the issues mentioned above, including:
- Updating the comments for backup LRU way in bsg_cache_miss
- Adding bsg_lru_pseudo_tree_backup to the sv.include files of the other tests
- Replacing bsg_fsb_node_trace_replay with bsg_trace_replay in lock_test testbench
- Moved the 32-bit basic_checker shared in regression_v2 and lock_test to the common directory, and rename it to  basic_checker_32
- Adding assert option of svaext to the Makefiles

testing/bsg_cache/lock_test/lru_stats.v
```diff
@@ -0,0 +1,60 @@
...
+        ctr
+        (.clk_i(clk_i)
+        ,.reset_i(reset_i)
+        ,.clear_i(1'b0)
+        ,.up_i(counter_en_li[i])
+        ,.count_o (hit_counter[i])
+        );      
+    end
+
+  always_comb
```
tommydcjung, line 45: Isn't this meant to be a bind module? If you want to print something at the end of simulation, you can use `final` block?

testing/bsg_cache/lock_test/lru_stats.v
```diff
@@ -0,0 +1,60 @@
...
+      bsg_counter_clear_up
+       #(.max_val_p (max_ctr_lp)
+       ,.init_val_p(0)) 
+        ctr
+        (.clk_i(clk_i)
+        ,.reset_i(reset_i)
+        ,.clear_i(1'b0)
+        ,.up_i(counter_en_li[i])
+        ,.count_o (hit_counter[i])
+        );
```
tommydcjung, line 42: I would just have something like `integer [n:0] hit_counter` and have them update in `always_ff` block. This module does not have to be synthesizable.

testing/bsg_cache/lock_test/lru_stats.v
```diff
@@ -0,0 +1,60 @@
+module lru_stats
+  import bsg_cache_pkg::*;
+  #(parameter ways_p="inv"
+    ,localparam lg_ways_lp=`BSG_SAFE_CLOG2(ways_p)
+    )
+  (
+    input clk_i
+    ,input reset_i
+
+    ,input miss_v_i
```
tommydcjung, line 10: unused

testing/bsg_cache/lock_test/sweep.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
```
tommydcjung, line 1: This is nice and thorough, but it takes too long. Can we possibly combine some of these tests or parallelize them?

tommydcjung: I guess you are trying to collect statistics. This is not a must, but something to consider.

dpetrisko: @Songchun-Li  I suggest using gnu parallel for this, it's pretty much what it's built for.

https://www.gnu.org/software/parallel/man.html
Here's an example of how we use it in BP
https://github.com/black-parrot/black-parrot/blob/master/ci/single_core_testlist.sh#L41

tommydcjung:
I don't know since when we got `vc_hdrs.h` showing up but now I see them everywhere. Can we add that to the `testing/.gitignore`?

tommydcjung:
Could you rebase on master so I can test it with manycore?

Songchun-Li:
> Could you rebase on master so I can test it with manycore?

It is rebased to the master now.

testing/bsg_cache/lock_test/sweep.sh
```diff
@@ -0,0 +1,61 @@
...
+      make test_lock_multiway.basic.run WAY_ON_LOCKED_P=$1
+    elif [ $2 == 1 ]
+      then
+        echo WAY_ON_LOCKED_P = $1  >> test_result.log
+        make test_lock2.basic.run WAY_ON_LOCKED_P=$1
+      else
+        echo WAY_ON_LOCKED_P = $1  >> test_result.log
+        make test_lock1.basic.run WAY_ON_LOCKED_P=$1
+  fi
+  make summary >> test_result.log
```
tommydcjung, line 21: when I run ./sweep.sh, I see these error messages.
```
grep: out/test_lock2/simv.log: No such file or directory
grep: out/test_lock_multiset/simv.log: No such file or directory
grep: out/test_lock_multiway/simv.log: No such file or directory
make: *** [summary] Error 2
```

bsg_cache/bsg_cache_miss.v
```diff
@@ -184,16 +184,44 @@ module bsg_cache_miss
...
+  #(.ways_p(ways_p)
+  ) backup_lru (
+    .disabled_ways_i(lock_v_i)
+    ,.modify_mask_o(modify_mask_lo)
+    ,.modify_data_o(modify_data_lo)
+  );
+
+  bsg_mux_bitwise #(
+    .width_p(ways_p-1)
+  ) mux (
```
tommydcjung, line 213: Can you name this instance to something slightly more descriptive? So that it's easier to look at timing reports in the future.

testing/bsg_cache/lock_test/sweep.sh
Songchun-Li: This is safe to ignore because it collect statistics of one test at a time, but `make summary` searches result of all 4 tests when it is used to grep the data into test_result.log. I will fixed it with a new make target later

bsg_cache/bsg_cache_miss.v
Songchun-Li: It is renamed now

bsg_cache/bsg_cache_miss.v
```diff
@@ -184,16 +184,44 @@ module bsg_cache_miss
...
+    ,.sel_i(modify_mask_lo)
+    ,.data_o(modified_lru_bits)
+  );
+
+  bsg_lru_pseudo_tree_encode #(
+    .ways_p(ways_p)
+  ) backup_lru_encode (
+    .lru_i(modified_lru_bits)
+    ,.way_id_o(backup_lru_way_id)
   );
```
tommydcjung, line 226: I just noticed that you have two bsg_lru_pseudo_tree_encode instantiated. But when lock_v_i is all zero, modify_mask_lo should also be all zero, which means that you only need to encode "modified_lru_bits" in the end. and the logic in line 303 can be simplified. Is my understanding correct?

Songchun-Li: Yep, and lock_v_i does not have to be all zero. The lru way encoded from the "modified_lru_bits" is the same as the lru way designated by the stats_mem_info, as long as the designated way is not locked. So it always ends up to the backup_lru_way_id. And this simplification is added.

Songchun-Li:
Hi @tommydcjung. All the issues mentioned above are resolved now, except parallelizing the sweep.sh, because of some trouble in regulating the data output. However, I rewrote the script and removed unnecessary clean, and it takes much less time than it used to be.

tommydcjung (approved):

PR #399, opened 2021-02-15
Make Handshake Signal of sipo_passthrough_dynamic Helpful

Songchun-Li:
In baseline design, sipo_passthrough_dynamic dose not register the last data packet to minimze the hardware. 

```
assign v_o         = (v_i & is_last_cnt) | is_zero_len | is_waiting;
assign ready_and_o = (ready_and_i | ~is_last_cnt) & ~is_waiting;
```
However, if the module works together with a demanding "ready_and_i" from downstream and a demading "v_i" from upstream, when there is a multi data word and it comes to the last cnt, `ready_and_i` is low and therefore, `v_i` is also low. In this case, the sipo gets into a dead lock.

To solve this issue, we would like to make the both `v_o` and ` ready_and_o` helpful by registering the last data word without jeopardizing the performance. To do so, a state register, `is_full_r` is added. 

```
assign v_o         = (v_i & is_last_cnt) | is_full_r | is_zero_len | is_waiting;
assign ready_and_o = ~is_full_r & ~is_waiting;
```
With this change, `ready_and_o` is always helpful when there is available space for upcoming data word, and the data in `dff_en_bypass` are protected from unintended updates by helpful upstream data.

### DIFF
--- a/bsg_dataflow/bsg_serial_in_parallel_out_passthrough_dynamic.v
+++ b/bsg_dataflow/bsg_serial_in_parallel_out_passthrough_dynamic.v
@@ -61,7 +61,7 @@ module bsg_serial_in_parallel_out_passthrough_dynamic
 
     logic [lg_max_els_lp-1:0] count_r, len_r;
     logic is_zero_cnt, is_last_cnt, is_zero_len, is_waiting;
-    logic en_li, clear_li, up_li;
+    logic en_li, clear_li, up_li, is_full_r;
     logic [max_els_p-1:0] data_en_li;
 
     // Register the length of upcoming transaction
@@ -87,6 +87,17 @@ module bsg_serial_in_parallel_out_passthrough_dynamic
     ,.clear_i   (clear_li   )
     ,.up_i      (up_li      )
     ,.count_o   (count_r    )
+    );
+
+    bsg_dff_reset_set_clear
+   #(.width_p(1)
+   ,.clear_over_set_p(1))
+    full_reg
+    (.clk_i(clk_i)
+    ,.reset_i(reset_i)
+    ,.set_i(is_last_cnt & v_i)
+    ,.clear_i(clear_li)
+    ,.data_o(is_full_r)
     );
 
     // Zero count means idle state, accepting new transaction
@@ -121,7 +132,7 @@ module bsg_serial_in_parallel_out_passthrough_dynamic
     // Dequeue incoming serial data and store in registers. Last data word 
     // of each transaction is not registered to minimize hardware. Accept no 
     // data word when waiting to send previous single word transaction. 
-    assign ready_and_o = (ready_and_i | ~is_last_cnt) & ~is_waiting;
+    assign ready_and_o = (ready_and_i | ~is_full_r) & ~is_waiting;
     assign first_o     = is_zero_cnt;
 
     // Decide when to update data registers
--- a/bsg_dataflow/bsg_serial_in_parallel_out_passthrough_dynamic.v
+++ b/bsg_dataflow/bsg_serial_in_parallel_out_passthrough_dynamic.v
@@ -89,13 +89,14 @@ module bsg_serial_in_parallel_out_passthrough_dynamic
     ,.count_o   (count_r    )
     );
 
+    // Set when all max_els_p registers are filled 
     bsg_dff_reset_set_clear
    #(.width_p(1)
    ,.clear_over_set_p(1))
     full_reg
     (.clk_i(clk_i)
     ,.reset_i(reset_i)
-    ,.set_i(is_last_cnt & v_i)
+    ,.set_i(is_last_cnt & v_i & ready_and_o)
     ,.clear_i(clear_li)
     ,.data_o(is_full_r)
     );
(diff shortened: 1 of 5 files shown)


dpetrisko (approved):
LGTM @gaozihou , can you take a quick look here?

Songchun-Li:
The testbench for bsg_sipo_passthrough_dynamic is built based on the testbench for bsg_sipo_passthrough, and bsg_sipo_passthrough_dynamic passes the test.

gaozihou:
Thanks @Songchun-Li @dpetrisko!

I have a quick question, what does "demanding ready_and_i and v_i" refer to? By definition of ready-and interface, v_i/ready_and_o, v_o/ready_and_i must be independent, it cannot "wait" for each other.. For example, if ready_and_i waits for v_o before assertion, then it should be called yumi_i instread..

There are three types of bsg noc interfaces: ready_and_link, ready_then_link and then_ready_link. The demanding ready_i is actually then_ready interface, the demanding v_i is actually ready_then interface. (https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_noc/bsg_noc_links.vh)

Also @dpetrisko, this change breaks the functionality of streamer_out, because it changes the behavior of dynamic sipo... It asserts ready_and_o before ready_and_i is asserted for last element (If you still remember our long discussion with Mark... Any change to the implementation detail will break that module)

dpetrisko:
Songchun means attaching a ready-then-valid on input side and ready-valid-and on the other. 

I’ll have to read the rest of your comment more closely... And refresh on the conversation with Mark :)

gaozihou:
> Songchun means attaching a ready-then-valid on input side and ready-valid-and on the other.
> 
> Ordinarily it's fine to attach ready-then-valid to a ready-valid-and, because the producer is demanding and the consumer is helpful. However in this specific module because we're passing through the handshakes it doesn't work.

Thanks Dan! I guess when we have ready-then-valid on input side and ready-valid-and on output side, ready_and_i will be asserted early (before v_o), then ready_and_o is asserted early, ready-then v_i is asserted late, and v_o is asserted in the end, still works fine?

dpetrisko:
I think @Songchun-Li can explain it better. It's more of high level deadlock problem than a combinational loop

Songchun-Li:
When the input side is attached to ready_then_link, and the output side is attached to then_ready_link, when a multi data words arraives,  the waveform looks like this. Everything works fine before the last data word comes in. 
Screen Shot 2021-02-16 at 4 52 35 PM
However, at the last count, is_last_cnt is asserted, then in current design, ready_and_i is passed through towards ready_and_o. Meahwhile, v_i is passed through towards v_o. 
```
assign ready_and_o = (ready_and_i | ~is_last_cnt) & ~is_waiting;    // equivalent to ready_and_i
assign v_o         = (v_i & is_last_cnt) | is_zero_len | is_waiting;// equivalent to v_i
```
Since ready_and_i is asserted depending on v_o, and v_i is asserted depending on ready_and_o in the given link protocol, here forms a deadlock between the handshake on the 2 sides, and the last data word is never acknowledged.

Therefore, the work in this PR is kind of isolating the ready_and_i from ready_and_o by adding the state register, is_full_r and the register for the last data word. The sipo is ready  as long as there is no pending single data word (is_waiting) or no pending multi data word (is_full_r). And its output is also valid if the data is not fetched immediately.
Screen Shot 2021-02-16 at 8 07 09 PM
```
assign v_o         = (v_i & is_last_cnt) | is_full_r | is_zero_len | is_waiting;   //  unfetched data is waiting with is_full_r
assign ready_and_o = ~is_full_r & ~is_waiting;      // ready_and_o is also asserted at last count, but not in the next cycle if data is not fetched.
```

With the last data word is also registered, the sipo_passthrough_dynamic also works when ready_and_i is asserted after the last data word arrives. Therefore, I think it should solve the issue in the streamer_in mentioned above.
Screen Shot 2021-02-16 at 8 07 18 PM

@dpetrisko Do u think this example sufficient enough?
@gaozihou Let me know if there is any further question.

gaozihou:
Thanks for your detailed explanation @Songchun-Li ! Yeah I agree that deadlock appears when input side is attached to ready_then_link, and output side is attached to then_ready_link, as shown in the waveform above.

I'd say attaching then_ready_link to the output side is violating the ready_and link interfacing rules.. We deliberately named it "ready_and_i" so that people would not accidentally attach yumi_i signal to this port... (The scenario @dpetrisko mentioned would work because it is ready-valid-and on the output side)

We have ready_and_i and ready_and_o interfacing for all 4 passthrough modules. They were designed for ready_and link only... In my opinion we should not expect these modules can still work when attaching input to ready_then_link and output side to then_ready_link. (Let's say we remove this module and attach input side and output side by wires, they will deadlock immediately... right?)

I understand that we currently have the need to attach ready_then_link and then_ready_link together. Normally I will not code in this way... If I have to I would insert a normal SPIO in between to isolate the handshaking signals. Or we can change port to "yumi_i" and breaks the consistency of our "passthrough series modules". Attaching a yumi signal to ready_and_i port is the last thing I would do...

drichmond:
I have no context here- but how are you inserting these waveforms into github?

dpetrisko:
@drichmond https://wavedrom.com/editor.html

It's super cool

dpetrisko:
Long discussion offline -- this doesn't fix the fundamental issue of the passthrough handshakes needing to be compatible

PR #404, opened 2021-02-25
Explicitly size amo_support_p in bsg_cache

dpetrisko:
Currently using a logical expression to derive the amo support level causes the parameter to infer that it is a 1-bit variable. This patch guarantees that it remains 32-bits no matter what is passed in


tommydcjung (approved):

dpetrisko:
good to merge?

PR #405, opened 2021-03-03
Add DRAMSIM3 makefile

dpetrisko:
So users don't have to figure it out for themselves


mrutt92 (approved):
I prefer to torture our users... but if you insist.

drichmond:
For maximum effectiveness you should obfuscate the makefile

dpetrisko:
good to go?

dpetrisko:
The only other place I could see is toplevel,  but I think it makes more sense in bsg_test

mrutt92:
you could name it libdramsim3.mk and folks can use the `-f` flag.

dpetrisko:
>you could name it libdramsim3.mk and folks can use the -f flag

Done

taylor-bsg (approved):

PR #410, opened 2021-03-05
Move cache_dma<->wormhole adapters to basejump

dpetrisko:
I grabbed the cache_dma<->wormhole components and stripped out the manycore-specific parameterizations.  I think there are more optimizations here (having the address in the header flit to save a cycle, etc.), but this should be a good start to combine both BP and manycore use cases.

Biggest change is to add a port wh_dma_id_i to bsg_wormhole_to_cache_dma.v, which means we can separate out the pod-specific hashing logic. This will require a small module to scan the flit in parallel to the module, but should have no PPA impact.

There's also an unrelated fix for bsg_cache_to_dram_ctrl.v I'm sneaking in here. This changes it to use an enum for the port assignment to match the definition bsg_dmc, which was updated.


bsg_cache/bsg_wormhole_to_cache_dma.v
```diff
@@ -0,0 +1,385 @@
+/**
+ *    bsg_wormhole_to_cache_dma.v
+ *
+ *    This module converts a bsg_cache wormhole link to an array of cache dma interface
+ *    this converts vcache wh link to an array of cache dma interface to that it can be interfaced to
+ *    bsg_cache_to_test_dram.v
+ *
+ *    Intended to be used for simulation only.
```
taylor-bsg, line 8: @tommydcjung I think there's nothing inherently here that is simulation only? Except that with extreme fanouts it may have trouble meeting cycle time?

bsg_cache/bsg_wormhole_to_cache_dma.v
```diff
@@ -0,0 +1,385 @@
+/**
+ *    bsg_wormhole_to_cache_dma.v
```
taylor-bsg, line 2: bsg_wormhole_to_cache_dma_endpoint_fanout?

taylor-bsg:
@tommydcjung We are standardizing on cache_dma (and associated wormhole tunnelling) as our DRAM interface across all of BSG. Dan has helpfully factored this code into basejump stl; can you update bigblade code to use this (and also offer any suggestions on the interface). Thanks!

dpetrisko:
@tommydcjung what's the use-case of having data_len != wh_len? Since the module does not support data_width != wh_flit_width

tommydcjung:
wh_len is how many "bit" is the length field in the wormhole header, data_len is how many "flit" does it take to send a block.

dpetrisko:
data_len_lp = 2**wh_len_width_p-1? The wormhole format has a payload of 1 flit for address and then N flits for data.

In that way, these parameters are unnecessary:
```
    , parameter data_width_p = "inv" // cache data width (bit)
    , parameter dma_data_width_p="inv"  // cache dma data width (bit)
    , parameter block_size_in_words_p = "inv" // cache block size (word)
```

tommydcjung:
I don't think it hurts to have those parameter, it makes the interface more clear and helps understand what the calculation is.

dpetrisko:
The problem is that we need to add assertions for all of them to just to ensure that the calculations come out correct. In general, for reusable modules it's better to pass exactly what you need so as to reduce the verification space. And since this module is strongly coupled to the wormhole header / payload format, the parameterization is fairly restrictive

tommydcjung:
I think the assertion is pretty simple. is the wh_len wide enough? Seems like a relatively minor issue. More concerned about whether this is copied and pasted functionally correct, and it supports the existing usage with grace. I think the answer is no to both.

dpetrisko:
If you have a system which is speaking to our new unified DRAM interface , it’s going to use this wormhole format. However, it is not necessarily a bsg_cache. So the parameterization makes no sense in that case. 

Bugs are caught by code review, thank you for catching mine.

tommydcjung:
Looking at the name of the module, it seems obvious that it's supposed to attach to the cache dma interface? having those parameters will help you validate wh_len is wide enough, which helps so that incompatible hardware are not attached together.

You are welcome! that's what I've been asked to do. Though it seems premature to make changes on manycore side yet, until some of the incompatibility issues are resolved, and more importantly this PR currently lacks verification.

tommydcjung:
> I grabbed the cache_dma<->wormhole components and stripped out the manycore-specific parameterizations. I think there are more optimizations here (having the address in the header flit to save a cycle, etc.), but this should be a good start to combine both BP and manycore use cases.

Address and other header info might not fit in a single flit?

dpetrisko:
> Address and other header info might not fit in a single flit?

Yep, what you can do is put a passthrough piso/sipo on the link, and look at the output of that rather than the header itself. It's zero overhead if the flit is large enough, and very little if it's multiple flits, but importantly it lets you change the flit width more flexibly.

dpetrisko:
>Looking at the name of the module, it seems obvious that it's supposed to attach to the cache dma interface? having those parameters will help you validate wh_len is wide enough, which helps so that incompatible hardware are not attached together.

It is attaching to the cache_dma interface.  This does not mean it's attaching to a cache.  In BP we have:
- passthrough store buffer
- accelerator
- coherent DMA engine

which all use the bsg_cache_dma interface, but are decidedly not caches.

> more importantly this PR currently lacks verification.

Yeah I'll add a directed testbench for this, since it's going to be an important interface in the future

bsg_cache/bsg_wormhole_to_cache_dma.v
```diff
@@ -0,0 +1,385 @@
...
+    , parameter dma_data_width_p = wh_flit_width_p
+  )
+  (
+    input clk_i
+    , input reset_i
+
+
+    // wormhole link
+    , input [wh_link_sif_width_lp-1:0]        wh_link_sif_i
+    , input [lg_num_dma_lp-1:0]               wh_dma_id_i
```
tommydcjung, line 43: In addition to this, do the snooping of the header here, and send it out by the ports: wh_cid_o, wh_src_cord_o, etc?
This way we don't have to the same snooping outside this module, and the combination hash unit can have interface
```
hash_unit #( 
 // parameters...
) hash0 (
  .wh_cid_i()
  ,.wh_src_cord_i()
  ,.wh_dma_o()
);
```

dpetrisko: Hm, that just creates a round trip for no reason.  Bit slicing is zero hardware overhead

```
hash_unit #( 
 // parameters...
) hash0 (
  .wh_link_sif_i()
  ,.wh_dma_id_o()
);
```

dpetrisko:
>Oh, well I guess data_len is wh_len + 1 because of the address flit. But it's still constant

I'm wrong on this. Because of the extra address flit, the length of the wormhole message is not a power of two, so it doesn't necessarily match the data payload. Funnily enough, this is why my tests worked, since it rounded up to the nearest power of two which was enough to make the data payload larger

bsg_cache/bsg_wormhole_to_cache_dma.v
dpetrisko: oh, I see. We need to sync them because of the in_fifo

bsg_test/bsg_nonsynth_axi_mem.v
```diff
@@ -123,7 +124,7 @@ module bsg_nonsynth_axi_mem
   // uninitialized data
   //
   logic [axi_data_width_p-1:0] uninit_data;
-  assign uninit_data = {(axi_data_width_p/32){32'hdead_beef}};
+  assign uninit_data = {(axi_data_width_p/32){init_data_p}};
```
dpetrisko, line 127: Unrelated, but hopefully uncontroversial change. Allowing user to set their own uninit string for AXI mem

bsg_cache/bsg_cache_dma_to_wormhole.v
```diff
@@ -0,0 +1,273 @@
...
+
+    , output logic [dma_data_width_lp-1:0] dma_data_o
+    , output logic dma_data_v_o
+    , input dma_data_ready_i
+
+    , input [dma_data_width_lp-1:0] dma_data_i
+    , input dma_data_v_i
+    , output logic dma_data_yumi_o
+
+    // Outgoing wormhole link is valid->ready, while incoming is standard ready-valid-and
```
dpetrisko, line 45: This is a somewhat breaking change. If this path is critical, will need to add an external FIFO. Could add bsg_cache_dma_to_wormhole_buffered, but I think it's unnecessary

bsg_cache/bsg_cache_to_dram_ctrl.v
```diff
@@ -47,7 +48,7 @@ module bsg_cache_to_dram_ctrl
     // dmc side
     , output logic app_en_o
     , input app_rdy_i
-    , output logic [2:0] app_cmd_o
+    , output app_cmd_e app_cmd_o
```
dpetrisko, line 51: Unrelated change for dram_ctrl, use enum instead of logic

bsg_cache/bsg_wormhole_to_cache_dma_fanout.v
```diff
@@ -0,0 +1,368 @@
...
+      recv_cache_id_r <= '0;
+    end
+    else begin
+      recv_state_r <= recv_state_n;
+      recv_cache_id_r <= recv_cache_id_n;
+    end
+  end
+
+  //synopsys translate_off
+  if (wh_flit_width_p != dma_data_width_p)
```
dpetrisko, line 360: These are all the assertions I could think of

taylor-bsg:
Adding @gaozihou as reviewer as well

bsg_cache/bsg_cache_dma_to_wormhole.v
```diff
@@ -0,0 +1,273 @@
...
+      RECV_READY: begin
+        return_fifo_yumi_li = return_fifo_v_lo;
+        recv_state_n = return_fifo_yumi_li
+          ? RECV_DATA
+          : RECV_READY;
+      end
+
+      RECV_DATA: begin
+        return_fifo_yumi_li = return_fifo_v_lo & dma_data_ready_i;
+        dma_data_v_o = return_fifo_yumi_li;
```
gaozihou, line 235: I guess dma_data_v_o = return_fifo_v_lo might be simpler? 
Seems that bsg_cache_dma.v does not require ready-then-valid handshaking

bsg_cache/bsg_cache_dma_to_wormhole.v
```diff
@@ -0,0 +1,273 @@
...
+
+  `declare_bsg_cache_dma_pkt_s(dma_addr_width_p);
+  `declare_bsg_then_ready_link_sif_s(wh_flit_width_p, wh_then_ready_link_sif_s);
+  `declare_bsg_ready_and_link_sif_s(wh_flit_width_p, wh_ready_and_link_sif_s);
+  wh_then_ready_link_sif_s wh_link_sif_in;
+  wh_ready_and_link_sif_s wh_link_sif_out;
+  bsg_cache_dma_pkt_s dma_pkt;
+  assign wh_link_sif_in = wh_link_sif_i;
+  assign wh_link_sif_o = wh_link_sif_out;
+  assign dma_pkt = dma_pkt_i;
```
gaozihou, line 65: Currently we have 2-cycle bubble between dma_pkt and dma_data_i (as shown in image). Adding a two-fifo on dma_pkt_i would eliminate 1-cycle bubble. (Trade-off between buffering and throughput?)

bsg_cache/bsg_cache_dma_to_wormhole.v
```diff
@@ -0,0 +1,273 @@
...
+    , input dma_data_ready_i
+
+    , input [dma_data_width_lp-1:0] dma_data_i
+    , input dma_data_v_i
+    , output logic dma_data_yumi_o
+
+    // Outgoing wormhole link is valid->ready, while incoming is standard ready-valid-and
+    // While one can functionally connect a valid->ready producer to a ready-valid-and consumer,
+    //   it may be desirable for timing or congestion to buffer input flits
+    , input  [wh_then_ready_link_sif_width_lp-1:0] wh_link_sif_i
```
gaozihou, line 48: I guess wh_link_sif_i functionally supports both then-ready-link and ready-and-link (in this module ready_i is always used together with wd_link_sif_o.v==1'b1).
 
Maybe we could add a comment and allow people attach both kinds to this port directly? Like people may want to attach this directly to a wh router.

bsg_cache/bsg_wormhole_to_cache_dma.v
dpetrisko: Updated comment

bsg_cache/bsg_cache_dma_to_wormhole.v
dpetrisko: Fixed!

dpetrisko:
Okay, think this is ready for final review

gaozihou (approved):
LGTM!

dpetrisko:
good to go?

dpetrisko:
good to go?

taylor-bsg:
@tommydcjung any feedback?

dpetrisko:
For context, there needs to be a corresponding manycore PR for this; I'm using it as part of my chip because both BP and manycore are using this DMA interface

tommydcjung (approved):

PR #411, opened 2021-03-08
[bsg_cache_to_test_dram] replace round robin module

tommydcjung:
For num_cache_p > 16


taylor-bsg (commented):
Seems like you are implementing an alternative implementation of bsg_round_robin_n_to_1? Perhaps we should outsource this code into that module and then offer it as an parameter (i.e. use_scan_p)

taylor-bsg (approved):

PR #417, opened 2021-03-14
Minor changes to DRAM modules

dpetrisko:
Should be uncontroversial fixes. Ran into these when bring up BP testbenches


drichmond:
The nonsynthesizable file looks fine, but I don't know enough about what's going on in the cache to DRAM controller to approve

dpetrisko:
It’s just using the enum rather than logic. Makes it match the actual dmc. https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_dmc/bsg_dmc.v#L29

drichmond (approved):

dpetrisko:
good to go then?

PR #418, opened 2021-03-15
Support for 0/1 element CAM

dpetrisko:
Supercedes / extends #415

### DIFF
--- a/bsg_misc/bsg_defines.v
+++ b/bsg_misc/bsg_defines.v
@@ -32,7 +32,7 @@
 `define BSG_SAFE_CLOG2(x) ( ((x)==1) ? 1 : $clog2((x)))
 `define BSG_IS_POW2(x) ( (1 << $clog2(x)) == (x))
 `define BSG_WIDTH(x) ($clog2(x+1))
-`define BSG_SAFE_MINUS(x, y) (((x)-(y)) < 0) ? 0 : ((x)-(y))
+`define BSG_SAFE_MINUS(x, y) (((x)<(y))) ? 0 : ((x)-(y))
 
 // calculate ceil(x/y) 
 `define BSG_CDIV(x,y) (((x)+(y)-1)/(y))
--- a/bsg_mem/bsg_cam_1r1w.v
+++ b/bsg_mem/bsg_cam_1r1w.v
@@ -43,14 +43,16 @@ module bsg_cam_1r1w
    , output logic                    r_v_o
   );
 
+  localparam safe_els_lp = `BSG_MAX(els_p,1);
+
   // The tag storage for the CAM
-  logic [els_p-1:0] tag_r_match_lo;
-  logic [els_p-1:0] tag_empty_lo;
-  logic [els_p-1:0] repl_way_lo;
-  wire [els_p-1:0] tag_w_v_li = repl_way_lo | {els_p{w_nuke_i}};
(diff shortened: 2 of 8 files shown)


dpetrisko:
For context, this is needed to parameterize BlackParrot without super-page support (default configuration)

bsg_misc/bsg_defines.v
```diff
@@ -32,7 +32,7 @@
 `define BSG_SAFE_CLOG2(x) ( ((x)==1) ? 1 : $clog2((x)))
 `define BSG_IS_POW2(x) ( (1 << $clog2(x)) == (x))
 `define BSG_WIDTH(x) ($clog2(x+1))
-`define BSG_SAFE_MINUS(x, y) (((x)-(y)) < 0) ? 0 : ((x)-(y))
+`define BSG_SAFE_MINUS(x, y) (((x) < (y))) ? 0 : ((x)-(y))
```
dpetrisko, line 35: @taylor-bsg I was encountering problems here with unsigned parameters.

`BSG_SAFE_MINUS(0, 1) -> 0-1=2**31-1 > 0 -> -1

I believe this version works with both signed and unsigned

dpetrisko:
>Too much reformatting

Could you explain what the reasoning is here? It's come up in many PRs I've seen.  Using "hide whitespace changes" makes it trivial to review such PRs. And after merge, it's easier to read code which maintains alignment than not

taylor-bsg:
A lot of git tooling does not have the “ignore white space” feature, including git blame, etc

dpetrisko:
git blame -w

dpetrisko:
> It’s just cleaner code to factor the expression

I guess I’m just trying to grok the difference in cases. With safe minus, we can see where the pattern of zero width signals is used, as you say. With a separate parameter, it’s a smaller change but a little unintuitive to have els_p and els_lp

taylor-bsg:
That’s why it’s safe_els_lp and not els_lp?

blame -w; nice if you are using the CLI; does not work in mobile / requires expert users / does not work for gamification aspects of GitHub?

dpetrisko:
I think anyone using git blame is an “expert user”. Personally, I’ve never used it because it’s absolutely useless in nearly all cases. https://tekin.co.uk/2020/11/patterns-for-searching-git-revision-histories

taylor-bsg:
024C1AB0-7F92-4F3A-AE03-EEBC261E86FD

It’s one of the only buttons to click on when you view a file...

dpetrisko:
>It’s one of the only buttons to click on when you view a file...

That doesn't stop git blame from being useless.

Consider that I had to spend a significant amount of time manually undoing whitespace changes, to create objectively worse code, that every developer has to deal with now.  For the use case that somebody who really wants to know what the single last change to clk_i was, on mobile, but doesn't want to look through the history, can't use the command line options and doesn't want to click through options in the github gui.  It's just not a good policy for a project.  It discourages contribution.

That's my piece.  Code is updated to remove whitespace changes

taylor-bsg (approved):
Thanks! IMHO the article you linked to overstates the case for uselessness of git blame. In the relatively rare case where you don’t want to see the latest modifier of the line, you can click on the prior version link and hit blame again. I think a lot of it comes down to 1) GUI versus command line users of GitHub and 2) whether you think of white space as code. But reasonable minds may disagree!

PR #422, opened 2021-03-25
Fixed bug in DPI Clock Generator

### LINKED ISSUE #421: Test bsg_nonsynth_dpi_clock_gen with multiple clocks
https://github.com/bespoke-silicon-group/basejump_stl/blob/61849cd850ed9993344e774af13bda15c5c67a6c/testing/bsg_test/bsg_nonsynth_dpi/top.v#L22

The current testbench attempts to do so, but actually multiply drives the ns_clk net. We're seeing potential issues with multiple clock gens in this BP issue: https://github.com/black-parrot/black-parrot/issues/798

drichmond:
* Swapped the order of pop and push - using a reference when poping will automatically update the reference.

Sort of fixes #421


dpetrisko (approved):

dpetrisko:
Fixes the problem in BP 798

PR #425, opened 2021-04-06
Cache Line Allocate Operation

Songchun-Li:
### Motivation 
In the current design of bsg_cache, when a cache line is missed, it always fetches the cache line from the further memory before load/store operation. In fact, if we are writing an entire cache line back, the fetched data is meaningless. Therefore, we add a new operation here , called cache line allocation. This operation only allocates a space in cache for the upcoming cache line writeback operation, without fetching any data from further memory, so that we can avoid an unnecessary cache fill.

Detailed Documentation: [link](https://docs.google.com/document/d/1EjAXMawQAMUan8xmuw9heMdjKiY_h2iVGt9pmCASFak/edit?usp=sharing)
Diagrams: [link](https://docs.google.com/presentation/d/1B8E6hULchOqSsvB-s49IAdESZNICWy0U1wmS7AUUy-k/edit?usp=sharing)


Songchun-Li:
Hi @tommydcjung! This is the first draft of the adding cache line allocate op to bsg_cache. **A major design problem to resolve is how to implement zeroing out the data of the allocated cache line in the data_mem.** For now, I have 2 ideas and implement the first one which is easier to implement and simpler for future maintaince, in this draft. You can find the description in the above doc link. Any comments or ideas over or beyond this 2 methods would be very helpful!

As for testing, there are now 2 tests focus on the correctness of the cache line allocate op, including _test_aalloc1_ and _test_aalloc_zero_. Detailed description can be found in the above doc and the test comments. Additional tests and binding modules for estimating the performance improvement from the nre feature are in progress. Suggestions on adding more tests are welcome!

tommydcjung:
I think you did the right thing by making the dma engine do the zeroing out. Currently, the dma engine, incoming load, and the write buffer are the three things accessing data_mem, and they need to be arbitrated, based on pipeline states. You don't want to add another thing that needs to be arbitrated.

tommydcjung:
@taylor-bsg maybe we should have separate opcodes for AALLOC that zeroes out (AALLOCZ) and one that doesn't?

taylor-bsg:
That sounds like a good idea to me. But there should be a parameter to
enable it, support_insecure_alloc, and an initial begin statement (with %m)
that warns if it is enabled.

dpetrisko:
Yeah great idea Tommy. You could even go further and keep a (nonsynth) set of shadow “secure bits” which will warn if the insecure data is used. In fact, you could have a synth version...

taylor-bsg:
Let's not go there .. at that point I would just want "read valid" bits for
each byte on the cache line.

M

dpetrisko:
Haha agreed, just thinking out loud

Songchun-Li:
The alloca_zero_p is replaced with AALLOCZ op, but I am not sure about the "read valid" bits and the parameter, support_insecure_alloc. It sounds like a new feature that is relevant to the allocation. Is there any clarification about these terms?

dpetrisko:
Hi Songchun, those would be for a future PR

dpetrisko:
There should be an assertion that you do not use AALLOC without use_insecure_alloc_p=1. It should print an error in this case

bsg_cache/bsg_cache.v
```diff
@@ -317,7 +317,7 @@ module bsg_cache
     ,.v_o(tag_hit_found)
   );
 
-  wire ld_st_miss = ~tag_hit_found & (decode_v_r.ld_op | decode_v_r.st_op);
+  wire ld_st_miss = ~tag_hit_found & (decode_v_r.ld_op | decode_v_r.st_op | decode_v_r.aalloc_op | decode_v_r.aallocz_op);
```
tommydcjung, line 320: Can you put those in a separate wire called "aalloc_miss"?

Songchun-Li:
Hi @tommydcjung, The testing part for cache allocate is finished, and it consists of 2 parts, correctness and performance.
For correctness, there are 3 tests _test_aalloc1_, _test_aalloc1_ and _test_aallocz2_ which cover most of the cases and are added to the _basic_tests_. 

And for performance improvement, there are 3 tests, _test_noaalloc_store_random_, _test_aalloc_store_random_ and _tes_aallocz_random_. And a new binding module, _store_miss_stats_ is added to track the number of store/alloc miss and the number of cycle in the store/alloc miss state, and finally to calculate the average miss cycle for each store/alloc miss. To  run these tests , simply run
`./perf_measure.sh `
in the _regression_v2_, and the result will be logged in _measure_result.log_.

Let me know if you have any good idea on the tests or on the deisgn itself!

PR #426, opened 2021-04-07
[WIP - DEFERRED] Periodic FIFO

dpetrisko:
This FIFO helper makes it easy to cross into a 1x/2x/Nx clock domain, although it's most likely useful only for N in the single digits.  It basically strobes with the ready/valid signals of the producer to the consumer so that it doesn't double count handshakes.  The necessary constraints are described in the file.

I made a test design demonstrating it here:
https://github.com/dpetrisko/bsg_fifo_divide/blob/main/tcl/constraints.tcl

The main motivation for this is to enable a very large L2 even with an aggressive cycle time for the main core, but without paying the cost of many slow cycles of CDC within the tile.

I also considered having the strobe inside this module and outputting a clock so that the constraints are all within this module. Fine with that implementation as well


bsg_dataflow/bsg_fifo_1r1w_periodic.v
```diff
@@ -0,0 +1,55 @@
+
+module bsg_fifo_1r1w_periodic
+ #(parameter `BSG_INV_PARAM(a_period_p)
```
stdavids, line 3: This isn't really the period right? It is more like "1 or clock A / clock B" -- essentially a ratio term.

bsg_dataflow/bsg_fifo_1r1w_periodic.v
```diff
@@ -0,0 +1,55 @@
+
+module bsg_fifo_1r1w_periodic
```
stdavids, line 2: This is not really a fifo right? There is no data storage. If it is intended to be used with a fifo, then I'd drop the 1r1w and called it a fifo periodic controller or something more along those lines.

bsg_dataflow/bsg_fifo_1r1w_periodic.v
```diff
@@ -0,0 +1,55 @@
...
+  // VCS 2016.06 incorrect orders assignments if these were 'assign' statements, so keep them
+  //   in the always_comb...
+  always_comb
+    begin
+      b_v_o         = accept_input & a_v_i;
+      a_ready_and_o = accept_input & b_ready_and_i;
+    end
+
+  if ((a_period_p != 1) && (b_period_p != 1))
+    $error("Only 1:N or N:1 division ratios are currently supported");
```
stdavids, line 50: Rather than having this check, maybe we can solve this by rethinking the parameters. One though is to remove the a_period_p and b_period_p parameters and instead have the parameters be "clk_a_faster_than_b_p = <0 or 1>" and "clk_factor_p = <int>"? These parameters feel more intuitive to me personally.

stdavids (commented):
As usually, before this is done some header comments with functional and parameter descriptions would be useful.

bsg_dataflow/bsg_fifo_1r1w_periodic.v
dpetrisko: I’m still kind of iterating here. I originally had a twofer here to break up the path, but it seems that the tools are able to handle sending between nx domains without the explicit buffer. As it currently stands, maybe bsg_flow_control_periodic?

dpetrisko:
Other, more timing friendly ways to do this

PR #429, opened 2021-04-19
Update bsg_cache.v

dpetrisko:
Tag hits should always be one_hot, so this is a little more efficient. I'm not missing some obscure use case where we want multiple matching entries am I?


dpetrisko:
closing for lack of activity

PR #431, opened 2021-04-29
Changing the special DONT_TOUCH instance names to BSG_DONT_TOUCH

stdavids:
Changing the special DONT_TOUCH instance names to BSG_DONT_TOUCH to be more consistent with other special variables in our cad flows. This also makes it more clear that this is a BSG feature and not something that EDA tools support by default.

The change of NO_CLOCK_GATE to BSG_NO_CLOCK_GATE is just a documentation error. The flow and RTL already uses BSG_NO_CLOCK_GATE.

### DIFF
--- a/hard/gf_14/README.md
+++ b/hard/gf_14/README.md
@@ -6,12 +6,12 @@ name during the chip implementation flow.
 
 ## Special Attributes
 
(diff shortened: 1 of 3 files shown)


gaozihou (approved):
LGTM!

taylor-bsg (approved):

PR #433, opened 2021-05-05
DPI Fixes and additions

drichmond:
* Lower valid after posedge to avoid negedge assertions
* Add new errorno value

### DIFF
--- a/bsg_test/bsg_nonsynth_dpi_to_fifo.v
+++ b/bsg_test/bsg_nonsynth_dpi_to_fifo.v
@@ -180,7 +180,6 @@ module bsg_nonsynth_dpi_to_fifo
    end
 
 
-   // TODO: Check that tx isn't called multiple times in a cycle
    function bit bsg_dpi_fifo_tx(input bit [width_p-1:0] data_bi);
 
       if(init_r === 0) begin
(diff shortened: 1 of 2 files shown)


bsg_test/bsg_nonsynth_dpi_to_fifo.v
```diff
@@ -257,17 +256,17 @@ module bsg_nonsynth_dpi_to_fifo
...
 
    // Save the last ready_i and data_o values for protocol checking
-   // and reset tx_r to 0 to
+   // and reset tx_r to 0 to.
    always @(posedge clk_i) begin
       ready_i_r <= ready_i;
       data_o_r <= data_o;
 
+      // reset v_o to 0 so that negedge clk assertions aren't triggered.
+      v_o <= v_o_n;
```
dpetrisko, line 265: Just so I understand, v_o_n is pulled down at posedge and latched here. Then v_o_n is optionally pulled up at negedge and latched there

dpetrisko: Is that right?

drichmond: * v_o_n is set _after_ the posedge
* v_o is set to v_o_n at negedge
* v_o_n is pulled down at negedge
* v_o is set to v_o_n at posedge (to 0)

dpetrisko (approved):

PR #435, opened 2021-05-07
channel tunnel support harden small fifo

gaozihou:
Pseudo large fifo is ideal when channel utilization is less than 50%. Hardened small fifo can be useful when attached to higher bandwidth io channel.


stdavids (approved):
Looks good, small thing to consider: maybe change the name from "use_harden_small_fifo_p" => "harden_small_fifo_p". It is a bit more consistent with the normal harden_p parameter and grammatically this is a bit cleaner (ie. using the word "harden" as a verb).

gaozihou:
Fixed. Thanks Scott!

PR #437, opened 2021-05-19
Fixing the x-pessimism issue in bsg_mesh_router_decoder_dor

tommydcjung:


tommydcjung:
some optimization in synthesis causes req_o to be X even when v_i = 0.

gaozihou (approved):
LGTM. As long as bsg_mesh_router_decoder_dor is not ungrouped this should work.

taylor-bsg (commented):
LGTM

PR #438, opened 2021-05-19
Use BSG_SAFE_MINUS for bit widths on UltraScale memories

WasabiFan:
This applies the equivalent change to https://github.com/bespoke-silicon-group/basejump_stl/pull/352 in the Xilinx-specific memory.


dpetrisko:
Cool!

Can you use the same style as https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_mem/bsg_mem_1rw_sync_synth.v?

- no new width variables
- SAFE MINUS in port list
- generate if for zero case

WasabiFan:
Updated. How does this look?

I haven't validated this new change yet as I'm having trouble coaxing Vivado into running the appropriate synthesis job, but it appears correct and I'll use this next time I go through the full flow.

dpetrisko:
LGTM

dpetrisko (approved):

WasabiFan:
This has been confirmed working in hardware on the Arty A7.

dpetrisko:
@taylor-bsg @drichmond good to merge this?

drichmond:
Sure, but I see conflicts.

drichmond:
Hold on, the Arty A7 isn't a ultrascale plus part. Do you have the ability to test it on an ultrascale Plus part?

WasabiFan:
I can't comment with any authority on the taxonomy of Xilinx chips, but I _can_ say that Vivado had the same issue expressed [here](https://github.com/black-parrot/black-parrot/blob/master/bp_common/syn/tcl/vivado_synth.tcl#L28) in BP for the Artix part in the Arty A7. Using the UltraScale+ variant of this memory fixed the synthesis issue, and this patch fixed further warnings, sim errors and (seemingly) tool slowdowns.

There's nothing specific in this file to the UltraScale+ family, as far as I can tell, aside from the fact it was the family in use at the time someone wrote it. If I'm understanding correctly it might be more appropriately scoped to e.g. all semi-recent Xilinx FPGAs. But I'm just guessing there. And that doesn't prove I haven't broke something for other devices.

I don't have another board to test with, unfortunately; the BlackParrot synthesis check in CI may be sufficient, alternately I'd appreciate if someone could try a vu37p with this submodule version.

drichmond:
I'm fine with a warning disappearing, and simulations.

Yes, you're right about the Xilinx families. It was the family in use when I (or more likely Leonard) wrote it.

dpetrisko:
Yeah we (BP) use this memory prophylactically on all Xilinx FPGAs. Perhaps it makes more sense to rename the porting dir to Xilinx? I like that better than the alternative of having a duplicate arty dir. Unless @drichmond, do you think it’s likely other hard modules will vary between the families in the future?

taylor-bsg:
The use of BSG_SAFE_MINUS looks wrong and does not match the original?

dpetrisko:
Aha yes, @WasabiFan there’s a trailing -1. Can you verify that when you were testing this the swap actually occurred? This should have resulted in [-1:0], which should be a width mismatch

taylor-bsg:
Regarding Xilinx versus Ultrascale plus, if we instantiate specific IP blocks like memories or LUTs then they are incompatible. One way is to have Xilinx/generic and Xilinx/Ultrascale+ etc, and then we can have a generic default and be able to override it. I think differences in individual versions of Vivado would be handled within the file.

WasabiFan:
Good catch @taylor-bsg -- my bad.

I've pushed a fix for this.

I ran VCS simulations of BP using stock basejump STL before this MR, the version I had in this MR as of this morning, and the version now that I've removed the erroneous `-1`s. The first reports "Port connection width mismatch" errors as I saw originally. The second reports those (although with different bad bounds) _plus_ some "Select index out of bounds" and a handful of others, internal to the module, due to the bad `-1`. The final reports no warnings nor errors whatsoever.

It seems likely I originally tested the version of this MR in the first commit rather than the version after reworking style. I've double-checked with latest BlackParrot dev branch after merging in latest master of this repo to confirm that I haven't introduced any new errors. Sorry about that!

taylor-bsg (approved):

PR #447, opened 2021-06-10
 bsg_fsb_node_trace_replay.v print wall time

tommydcjung:
bsg_fsb_node_trace_replay.v used in bsg_tag_trace_replay

### DIFF
--- a/bsg_fsb/bsg_fsb_node_trace_replay.v
+++ b/bsg_fsb/bsg_fsb_node_trace_replay.v
@@ -176,7 +176,16 @@ module bsg_fsb_node_trace_replay
         if (instr_completed & ~reset_i & ~done_r)
           begin
              case(op)
-               eSend: $display("### bsg_fsb_node_trace_replay SEND %d'b%b (%m)", ring_width_p,data_o);
+               eSend: begin
+                reg [63:0] t;
+                int fd;
+                t = 0;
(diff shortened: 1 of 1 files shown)


drichmond:
@taylor-bsg I would be interested in seeing your version. I was considering making a general comment earlier:

This is becoming a common pattern in BSG. Perhaps we should have a module that prints current timestamps based on an event. However, one issue is that output file could be stomped on by different modules -- particularly in multi-threaded environments.

I suspect your version fixes my second thought

mrutt92:
I wrote pretty much an exact copy of this code in hb_bigblade. Echoing @drichmond point, we might want to have a function or macro to do this?

tommydcjung:
I think we might as well make a dpi version that is easy to include, so that it doesn't write to some file and parse it back.

taylor-bsg:
Hi, as mentioned I have solution for this that avoids the file stuff and DPI, we just read directly /proc/uptime. I can commit it later tonight.

taylor-bsg:
I committed the code

taylor-bsg:
https://github.com/bespoke-silicon-group/basejump_stl/commit/57553c53cda0d825ee3f576cc879e84e26726330

taylor-bsg:
Already merged.

PR #449, opened 2021-06-14
Verilatable testbenches

adithyasunil26:
Modifying existing testbenches to work with verilator. These testbenches can be used as a target with FuseSoC.


testing/bsg_misc/bsg_dff/test_bsg.v
```diff
@@ -1,5 +1,3 @@
-`define WIDTH_P 3
```
dpetrisko, line 1: make sure not to change the existing semantics of the testbenches. The current makefiles rely on WIDTH_P being a define, not a param

dpetrisko:
Thanks, this looks great to me! You can continue on this PR with the remaining bsg_misc modules

adithyasunil26:
Thanks for the review! I will continue committing the remaining testbenches to this PR.

testing/bsg_misc/bsg_dff/test_bsg.v
```diff
@@ -16,22 +14,28 @@
 
 ***************************************************************************/
 
-module test_bsg;
+module test_bsg 
+#(parameter width_p=3,
+  parameter sim_clk_period=10,
```
taylor-bsg, line 21: parameters should end in _p

testing/bsg_misc/bsg_adder_one_hot/test_bsg.v
```diff
@@ -0,0 +1,49 @@
+module test_bsg
+#(parameter width_p=4, 
+  parameter output_width_p=2*width_p-1,
+  parameter sim_clk_period=10,
```
taylor-bsg, line 4: parameters should end in _p

taylor-bsg:
nice start! @adithyasunil26

testing/bsg_misc/bsg_dff/test_bsg.v
adithyasunil26: Thank you for pointing that out. It must have slipped my mind while adding the `sim_clk_period` parameter. 
Will correct this asap.

adithyasunil26:
Thank you! @taylor-bsg

dpetrisko (approved):
LGTM!

dpetrisko:
Let's merge this before #459

taylor-bsg:
quick question; any reason we are removing asserts? I thought that Verilator supports them?

dpetrisko:
oh good catch, I didn't notice that.  @adithyasunil26 were you running into problems with asserts and verilator? They should be supported

adithyasunil26:
@dpetrisko @taylor-bsg Verilator wasn't giving any syntax errors or something like that with assert statements and the code seems like it is running but they weren't working as expected. I tried deliberately making assert statements fail and verilator did not give the $error in the else instead it seemed to continue running like the assert didn't exist.

dpetrisko:
Ah, you have to add --assert to the Verilator command line

adithyasunil26:
@dpetrisko ohhh okay, will do that and revert the assert lines in this PR

adithyasunil26:
@dpetrisko It's done.

dpetrisko:
Great! Should be good to go then

taylor-bsg:
Perhaps we can modify bsg_nonsynth_clock_gen so it automatically includes the dpi version for verilator? this way we do not have to have all of these ifs all over the testbench code? 

@drichmond

dpetrisko:
Hi @adithyasunil26 can you update based on #463? Thanks!

adithyasunil26:
Hey @dpetrisko! I'm on it.

adithyasunil26:
@dpetrisko I have removed the ifdefs in the testbenches. Let me know if I missed something. Thanks!

taylor-bsg (approved):
LGTM!

PR #452, opened 2021-06-23
fixup to makefile

mrutt92:


dpetrisko (approved):

mrutt92:
I don't have perms to fix this, @taylor-bsg

taylor-bsg:
reason for going from C++-14 to C++-11?

mrutt92:
C++14 is not necessary for compiling DRAMSim3 and it's not as widely supported across systems/compilers as C++11.

PR #457, opened 2021-07-11
Adding inv parameters

dpetrisko:
This was a scripted change to add inv parameters to the vast majority of remaining basejump modules. For reference, the command was:

```
find basejump_stl/ -type f -name "*.v" -exec sed -i '/"inv"/{s/parameter /`BSG_INV_PARAM(/g;s/=.*"inv"/)/g;}' {} \;
```


taylor-bsg:
Neat.

Seems like BSG abstract module was added to some of these but not all?

dpetrisko:
Yeah I couldn’t figure out a good way to script it, fine exercise to the reader

dpetrisko:
FYI latest cadence Xcelium (VCS equivalent) doesn’t support default parameters without values

taylor-bsg:
I am curious if that is also the case for their synthesis tools...  and
whether = "inv" works

M

dpetrisko:
Puzzle completed:

```
for f in *; do \
base=$(basename "$f" '.v'); \ if ! grep -q "BSG_ABSTRACT_MODULE" $f ; then \
sed -i $"/endmodule/a\\\nBSG_ABSTRACT_MODULE($base)" $f; \
fi \
done
```

All modules should have BSG_INV_PARAM and BSG_ABSTRACT_MODULE now

dpetrisko:
There are a few minor problems here
- Generators with bsg_abstract modules
- multiple inv parameters per line
- -1 parameters

dpetrisko:
@taylor-bsg I believe all problems are solved. Additionally, experiments with cadence shows that Genus does support bare parameters but Xcelium does not. Both Genus and Xcelium support BSG_ABSTRACT_MODULE style top-levels. Therefore, I tweaked the bsg_defines macro slightly:

```
diff --git a/bsg_misc/bsg_defines.v b/bsg_misc/bsg_defines.v
index b3f40e79..40908055 100644
--- a/bsg_misc/bsg_defines.v
+++ b/bsg_misc/bsg_defines.v
@@ -27,8 +27,11 @@
 // it should be sufficient to omit the "inv" but we include this for tool portability
 // if later we find that all tools are compatible, we can remove the use of this from BaseJump STL

-//`define BSG_INV_PARAM(param) param = "inv"
+`ifdef XCELIUM // Bare default parameters are incompatible as of 20.09.012
+`define BSG_INV_PARAM(param) param = "inv"
+`else // VIVADO, DC, VERILATOR, GENUS
 `define BSG_INV_PARAM(param) param
+`endif
```

This should be good to go now!

bsg_cache/bsg_cache_non_blocking_decode.v
```diff
@@ -61,3 +61,5 @@ module bsg_cache_non_blocking_decode
   assign decode_o.mgmt_op = ~(decode_o.ld_op | decode_o.st_op | decode_o.block_ld_op);
 
 endmodule
+
```
taylor-bsg, line 64: This module is not abstract

bsg_cache/bsg_cache_decode.v
```diff
@@ -92,3 +92,5 @@ module bsg_cache_decode
   end
 
 endmodule
+
```
taylor-bsg, line 95: not abstract

bsg_chip/bsg_rocket/bsg_chip_rocket.v
```diff
@@ -128,3 +128,5 @@ module bsg_chip_rocket #(
       );
 
 endmodule
+
```
taylor-bsg, line 131: not abstract

bsg_chip/bsg_rocket/bsg_nonsynth_chipset_rocket_fsb.v
```diff
@@ -259,3 +259,5 @@ module bsg_nonsynth_chipset_rocket
       );
 
 endmodule // bsg_nonsynth_chipset_rocket
+
+`BSG_ABSTRACT_MODULE(bsg_nonsynth_chipset_rocket_fsb)
```
taylor-bsg, line 263: not abstract

bsg_chip/bsg_rocket/bsg_rocket_core_fsb.v
```diff
@@ -157,3 +157,5 @@ module bsg_rocket_core_fsb
       );
 
 endmodule
+
+`BSG_ABSTRACT_MODULE(bsg_rocket_core_fsb)
```
taylor-bsg, line 161: not abstract

bsg_clk_gen/bsg_clk_gen_osc.v
```diff
@@ -69,3 +69,5 @@ module bsg_clk_gen_osc
 
 
 endmodule
+
+`BSG_ABSTRACT_MODULE(bsg_clk_gen_osc)
```
taylor-bsg, line 73: not abstract

bsg_clk_gen/bsg_dly_line.v
```diff
@@ -69,3 +69,5 @@ module bsg_dly_line
 
 
 endmodule
+
```
taylor-bsg, line 72: not abstract

taylor-bsg (commented):
A lot of modules got tagged with abstract that are not abstract

dpetrisko:
What's the downside of tagging all modules as abstract?

taylor-bsg:
The purpose of the tag is to add semantic information to the module; this
is adding semantic noise =)

m

dpetrisko:
Here I was thinking the purpose was to not kill Vivado... :P

Fixed!

taylor-bsg (commented):
Nice!

- There are a bunch of testing/bsg_cache modules that did not get the abstract module!
- There are a bunch of files with whitespace only change; can you remove those from the commit list?
- This is a minor aside, not necessary to fix; but I noticed you inserted the parameter keyword; this is now unnecessary in SystemVerilog, is that right?  Maybe we want to standardize going forward on not using it as part of the coding guidelines, since it affects code metrics versus other languages like Py* and *hisel

dpetrisko:
Feedback addressed! The easiest way to squash whitespace only changes it ended up clobbering the commit history of this branch, sorry about that

taylor-bsg (approved):

PR #460, opened 2021-07-15
remove reset from bsg_tag_client

taylor-bsg:
This will of course require some patches to various modules, but it should be straight-forward delete of the default parameter and the reset input signal; I believe no current hardware makes us of the reset signal; it was essentially a fail-safe (what happens if tag client does not work!) that was too complicated to use in practice.


bsg_tag/bsg_tag_client.v
```diff
@@ -55,14 +49,6 @@ module bsg_tag_client
```
tommydcjung, line 49: you can remove default_p="inv" now?

dpetrisko (approved):
LGTM!

PR #461, opened 2021-07-15
make tag master decentralized the standard tag

taylor-bsg:
refactor redundant code in repo

Co-authored-by: gaozihou <gaozihou1@gmail.com>

### DIFF
--- a/bsg_tag/bsg_tag_master.v
+++ b/bsg_tag/bsg_tag_master.v
@@ -24,16 +24,27 @@
 module bsg_tag_master
   import bsg_tag_pkg::bsg_tag_s;
 
-   // els_p is the number of clients to attach
+   // els_p is the maximum number of clients supported in system
+   // local_els_p is the number of clients to attach
    // lg_width_p is the number of bits used to describe the payload size
-   
-   #(els_p="inv", lg_width_p="inv", debug_level_lp=2)
+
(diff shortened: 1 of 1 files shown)


taylor-bsg (commented):
@gaozihou I see a few things that are different -- the enable signal is not being forwarded, we are not flopping the output signals to the clientsm what are the implications of these?

gaozihou:
Comparisons between bsg_tag_master and bsg_tag_master_decentralized:
1. Only one bsg_tag_master per chip vs multiple bsg_tag_master_decentralized per chip
2. bsg_tag_master is far away from tag clients vs bsg_tag_master_decentralized is pretty close to tag clients

In decentralized tag master en_i port is removed and enable signal to tag client is 1'b1. There is an assumption that data_i signal is actually (tag_data_i & tag_en_i), which means only one AND gate is needed for all decentralized tag masters, and we don't have to route tag_en_i signal all over the chip.

Since bsg_tag_master_decentralized is close to tag clients, output registers are removed to reduce latency and resource utilization.

So I guess the implication is that bsg_tag_master_decentralized cannot be used as bsg_tag_master directly..

dpetrisko:
Would it make sense to have bsg_tag_master wrap bsg_tag_master_decentralized with a constant offset '0?

PR #464, opened 2021-07-19
Adding pickle macros

dpetrisko:
This updates the pickled hardened memories to use the macro scheme we've adopted in our other hardened swaps. I created a new directory called pickle/ since there's nothing 40nm specific here. Options are:
1) have pickle/ and pickle_40/ coexist
2) have these changes in pickle_40/
3) remove pickle_40 and use pickle/

My preference is 3 for clarity, but I'm not passionate about it

PR #467, opened 2021-07-20
[RFC] Generic sram wrappers

dpetrisko:
The hardened memory .vh files are the differentiation between processes. The hardened wrappers themselves should be the same. This PR reduces the work of updating each tech node when we add/subtract parameters from bsg_mem modules. Rare, but it still happens frequently enough where this is useful.

The bsg methodology for hardened RAMs is to copy these generic wrappers out and modify the substitution values, so any hardcoded or unsupported parameters per project can be set there. If a tech node has a weird default that we really want to enforce, we could wrapper the generic value with a warning about setting that parameter.

There are a few rough edges, like having 'mux' in pickled designs where that's meaningless.

Depends on #464


dpetrisko:
Superceded by #475

PR #468, opened 2021-07-20
Variable pipeline depth mul_add for easy DSP absorption on FPGAs

mysoreanoop:
Not sure how to handle the ZYNQ variable -- any other better way?

### DIFF
--- a/bsg_misc/bsg_mul_add.v
+++ b/bsg_misc/bsg_mul_add.v
@@ -19,7 +19,7 @@ module bsg_mul_add_unsigned #(
     ,input [width_a_p-1 : 0] a_i
     ,input [width_b_p-1 : 0] b_i
     ,input [width_c_p-1 : 0] c_i
-    ,output [width_o_p : 0] o
+    ,output [width_o_p-1 : 0] o
     );
 
     localparam pre_pipeline_lp = 0;
--- a/bsg_misc/bsg_mul_add.v
+++ b/bsg_misc/bsg_mul_add.v
@@ -1,5 +1,11 @@
-`include "bsg_defines.v"
+// A special module for use with HardFloat's mulAddRecFN for 
+// easy absorption of pipeline stages by the DSPs that get
+// synthesised in some FPGAs. This helps in retiming the 
+// paths in FPGA implementations as only immediate registers 
+// are absorbed, and global retiming does not seem to do this.
 
+`include "bsg_defines.v"
+  
 module bsg_mul_add #(
     parameter width_a_p = 4
     ,parameter width_b_p = 4
--- a/bsg_misc/bsg_mul_add.v
+++ b/bsg_misc/bsg_mul_add.v
@@ -1,0 +1,39 @@
+`include "bsg_defines.v"
+
+module bsg_mul_add #(
+    parameter width_a_p = 4
+    ,parameter width_b_p = 4
+    ,parameter width_o_p = width_a_p + width_b_p + 1
+    ,parameter pipeline_p = 3
+  ) (
+    input clk_i
+    ,input [width_a_p-1 : 0] a_i
+    ,input [width_b_p-1 : 0] b_i
+    ,input [width_a_p + width_b_p - 1 : 0] c_i
+    ,output [width_o_p : 0] o
+    );
+
+    `ifdef ZYNQ
+      initial assert (pipeline_p > 2) else $error ("pipeline stages may not be enough")
+    `endif
+
+    localparam pre_lp = pipeline_p > 2? 1 : 0;
+    localparam post_lp = pipeline_p > 2? pipeline_p - 1 : pipeline_p;
+    
+    wire [width_a_p-1:0] a_r;
+    wire [width_b_p-1:0] b_r;
+    wire [width_a_p + width_b_p -1 : 0] c_r;
+    bsg_dff_chain #($bits(a_r + b_r + c_r), pre_lp)
+        pre_mul_add (
+            .clk_i(clk_i)
+            ,.data_i({a_i, b_i, c_i})
+            ,.data_o({a_r, b_r, c_r})
+        );
+    wire [width_a_p + width_b_p : 0] o_r = a_r * b_r + c_r;
+    bsg_dff_chain#($bits(o_r), post_lp)
+        post_mul_add (
+            .clk_i(clk_i)
+            ,.data_i(o_r)
+            ,.data_o(o)
+        );
+endmodule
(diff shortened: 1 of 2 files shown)


bsg_misc/bsg_mul_add.v
```diff
@@ -0,0 +1,45 @@
+// A special module for use with HardFloat's mulAddRecFN for 
+// easy absorption of pipeline stages by the DSPs that get
+// synthesised in some FPGAs. This helps in retiming the 
+// paths in FPGA implementations as only immediate registers 
+// are absorbed, and global retiming does not seem to do this.
+
+`include "bsg_defines.v"
+  
+module bsg_mul_add #(
```
taylor-bsg, line 9: bsg_mul_add_unsigned

bsg_misc/bsg_mul_add.v
```diff
@@ -0,0 +1,45 @@
...
+    ,input [width_b_p-1 : 0] b_i
+    ,input [width_a_p + width_b_p - 1 : 0] c_i
+    ,output [width_o_p : 0] o
+    );
+
+    `ifdef ZYNQ
+      initial assert (pipeline_p > 2) else $error ("pipeline stages may not be enough")
+    `endif
+
+    localparam pre_lp = pipeline_p > 2? 1 : 0;
```
taylor-bsg, line 26: pull these as parameters to the module, post_pipeline_p, pre_pipeline_p

bsg_misc/bsg_mul_add.v
```diff
@@ -0,0 +1,45 @@
...
+// synthesised in some FPGAs. This helps in retiming the 
+// paths in FPGA implementations as only immediate registers 
+// are absorbed, and global retiming does not seem to do this.
+
+`include "bsg_defines.v"
+  
+module bsg_mul_add #(
+    parameter width_a_p = 4
+    ,parameter width_b_p = 4
+    ,parameter width_o_p = width_a_p + width_b_p + 1
```
taylor-bsg, line 12: width_c_p should be a parameter

taylor-bsg: tighter bound on default value for width_o_p, maybe something like `BSG_SAFE_CLOG2( ((1 << width_a_p) - 1) * ((1 << width_b_p) - 1) + (1 << width_c_p)-1+1)

bsg_misc/bsg_mul_add.v
```diff
@@ -0,0 +1,45 @@
...
+    ,parameter pipeline_p = 3
+  ) (
+    input clk_i
+    ,input [width_a_p-1 : 0] a_i
+    ,input [width_b_p-1 : 0] b_i
+    ,input [width_a_p + width_b_p - 1 : 0] c_i
+    ,output [width_o_p : 0] o
+    );
+
+    `ifdef ZYNQ
```
taylor-bsg, line 22: can delete this; then clone this file (after it has converged) into 

hard/ultrascale_plus/bsg_misc/

and add the zynq specific stuff.

and we would use a modified file list

bsg_misc/bsg_mul_add.v
taylor-bsg: make the default all post-pipelining; then the zynq specific file can modify those parameters.

bsg_misc/bsg_mul_add.v
```diff
@@ -0,0 +1,45 @@
...
+      initial assert (pipeline_p > 2) else $error ("pipeline stages may not be enough")
+    `endif
+
+    localparam pre_lp = pipeline_p > 2? 1 : 0;
+    localparam post_lp = pipeline_p > 2? pipeline_p - 1 : pipeline_p;
+    
+    wire [width_a_p-1:0] a_r;
+    wire [width_b_p-1:0] b_r;
+    wire [width_a_p + width_b_p -1 : 0] c_r;
+    bsg_dff_chain #($bits(a_r + b_r + c_r), pre_lp)
```
taylor-bsg, line 32: add a empty line between wires and the DFF chain.

bsg_misc/bsg_mul_add.v
```diff
@@ -0,0 +1,45 @@
...
+        pre_mul_add (
+            .clk_i(clk_i)
+            ,.data_i({a_i, b_i, c_i})
+            ,.data_o({a_r, b_r, c_r})
+        );
+    wire [width_a_p + width_b_p : 0] o_r = a_r * b_r + c_r;
+    bsg_dff_chain#($bits(o_r), post_lp)
```
taylor-bsg, line 39: space between bsg_dff_chain and "#"

taylor-bsg (commented):
nice first iteration

bsg_misc/bsg_mul_add.v
mysoreanoop: So, should we can keep the pipeline_p as the only parameter and device specific implementations can distribute it accordingly? Cause having both pre_pipeline_p and post_pipeline_p obviates the Zynq specific changes, right?

dpetrisko: Yep

bsg_misc/bsg_mul_add.v
```diff
@@ -0,0 +1,47 @@
+// 21 Jul 2021
+// A special module for use with HardFloat's mulAddRecFN for
```
taylor-bsg, line 2: change comment to be more generic:

// multiply add often takes more than one cycle of time.
// this module provides a mechanism for indicating the degree
// of pipelining required. the implementation still relies upon
// retiming support from the cad tools. the basejump_stl
// /hard mechanisms allows for platform-specific implementations
// to be swapped in.

taylor-bsg (approved):
good! just fix the comment and good to go.

dpetrisko:
Shouldn’t pipeline_p be an array parameter? This way, the first stage is always 0

taylor-bsg:
Where the pipeline registers go should be abstracted by the implementation, to allow portability.

dpetrisko:
What I’m saying is that the hard/ version of this will have to duplicate all this code, whereas if it were an array parameter, the hard version could simply override the parameter. Could do something like this for a default to get best of both worlds:

// set to total latency 
parameter pipeline_p = 3
// Should be overridden by portability layer
parameter pipeline_dist_p = ‘{pipeline_p,0}

taylor-bsg:
Yeah, but then we need three files, an _synth file, and the two others.
Which is okay, I just thought we would wait until we have more than one
child derived module. :-)

dpetrisko:
Works for me!

taylor-bsg (approved):

PR #469, opened 2021-07-22
[bsg_barrier] add vcs testbench; minor fix; add schematic

tommydcjung:


taylor-bsg:
added schematic link to master; makefile for test appears to be newer in master.

PR #472, opened 2021-07-23
Fifo 1r1w small hardened bug fix

gaozihou:
This PR fixes the bug that data_o turns X when fifo is full.
Comments and README are updated.
Testbench is updated to cover all possible combination of clock speeds to catch the bug.

To run regression:
$ cd basejump_stl/testing/bsg_dataflow/bsg_fifo_1r1w_small_hardened
$ git clone https://bitbucket.org/taylor-bsg/bsg_cadenv.git ../../../../bsg_cadenv
$ make -j125
$ make check

All 125 tests should pass. With buggy fifo_1r1w_fifo_small there should be 20 failed tests.

### DIFF
--- a/bsg_dataflow/bsg_fifo_1r1w_small_hardened.v
+++ b/bsg_dataflow/bsg_fifo_1r1w_small_hardened.v
@@ -63,8 +63,7 @@ module bsg_fifo_1r1w_small_hardened #(parameter `BSG_INV_PARAM(width_p)
    logic [ptr_width_lp-1:0] rptr_n;
    // avoid reading and writing same address in mem_1r1w_sync
    logic [width_p-1:0] data_o_mem, data_o_reg;
-   logic read_write_same_addr_n;
-   logic write_bypass_r, write_bypass_n;
+   logic read_write_same_addr_r, read_write_same_addr_n;
   
    bsg_fifo_tracker #(.els_p(els_p)
                           ) fts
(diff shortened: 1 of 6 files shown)


tommydcjung:
https://github.com/bespoke-silicon-group/basejump_stl/pull/476 I added a randomized testbench. Merge it if found useful.
My feedback is that r_v_i is on almost all the time, which isn't a functional bug, but it would be wasting energy.

dpetrisko:
Randomized testbench is a good idea. I had concerns that there is still a lack of coverage on edge cases, due to the clock periods being constant, but the randomizer most likely addresses it.  For instance, I don't think the fifo being full while yumi is low was tested. Coverpoints should expose any gaps, and reducing the size of the fifo to 4 or 8 could help undirected tests reach coverage faster

tommydcjung:
Yeah I wasn't saying that the current random testbench is sufficient to cover all cases. some tweaks would be required to get full coverage, or even some directed tests.

tommydcjung (approved):

tommydcjung:
On the other hand, the full coverage doesn't necessarily mean that the hardware is bug-free. The testbench still needs to be designed in such way that full coverage implies full correctness.

dpetrisko:
>On the other hand, the full coverage doesn't necessarily mean that the hardware is bug-free. The testbench still needs to be designed in such way that full coverage implies full correctness.

Yep for sure. To be clear, I was saying the testbench without randomness couldn't address these issues, while the randomized one probably does, or could be made to

PR #473, opened 2021-07-24
[WIP] 1RW bit-masked RAM using 1R1W RAM

mysoreanoop:
PR for design discussion. The testing infrastructure needs work; I'm on it.


hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,114 @@
...
+
+  always_ff @(posedge clk_i) begin
+     if(r_en)
+        r_data_r <= ram[addr_i];
+       
+     if(w_en) 
+        ram[addr_r] <= r_data_r & ~w_mask_r | w_data_r & w_mask_r;
+  end
+
+  `ifndef SYNTHESIS
```
dpetrisko, line 74: take out this behavior, we don't want to mismatch with synthesis -- common cause of chip-killing bugs

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,114 @@
...
+  , input                        v_i
+  , input  [ addr_width_lp-1:0]  addr_i
+  , input                        w_i
+  , input  [width_lp:0]          data_i
+  , input  [width_lp:0]          w_mask_i
+
+  , output [width_lp:0]          data_o
+);
+  
+  initial assert(int'(addr_i) < els_p) else $warning("Accessing uninitialized address!");
```
dpetrisko, line 47: put in always_ff @(negedge), this will only trigger once

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,114 @@
...
+
+  logic [addr_width_lp-1:0] addr_r;
+  logic [width_lp:0]        r_data_r;
+  logic [width_lp:0]        w_data_r;
+  logic [width_lp:0]        w_mask_r;
+  logic                     w_en_r;
+  logic                     w_r;
+
+  //AM: may be skipped if it's guaranteed that the first operation to an uninitialized address is always a full write
+  logic                     not_first_r;
```
dpetrisko, line 57: It's up to the user to initialize the RAM

mysoreanoop: Will need it for the first case; updated comment.

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,114 @@
...
+      w_en_r   <= 1'b0;
+      w_r      <= 1'b0;
+      w_data_r <= '0;
+      r_data_r <= '0;
+      w_mask_r <= '0;
+      addr_r   <= '0;
+    end
+
+    else
+      if(v_i) begin: v
```
dpetrisko, line 95: don't label non-generate blocks

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,107 @@
...
+  always_ff @(posedge clk_i) begin
+     if(r_en)
+        r_data_r <= ram[addr_i];
+       
+     if(w_en) 
+        ram[addr_r] <= r_data_r & ~w_mask_r | w_data_r & w_mask_r;
+  end
+
+  always_ff @(posedge clk_i) begin
+    if(!reset_i) begin
```
dpetrisko, line 73: Think about which of these actually need reset. I believe it's very few

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,107 @@
...
+      if(v_i) begin: v
+        w_r     <= w_i;
+        addr_r  <= addr_i;
+        w_mask_r <= ((addr_r == addr_i) & w_i == 1'b0) ? '0 : w_mask_i;
+        w_data_r <= ((addr_r == addr_i) & w_i == 1'b0) ? '0 : data_i;
+
+        if(!not_first_r)
+          not_first_r <= 1'b1;
+        else begin
+          w_en_r <= (addr_r == addr_i) ?
```
dpetrisko, line 93: factor out common subexpression (addr_r == addr_i)

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,107 @@
...
+        w_data_r <= ((addr_r == addr_i) & w_i == 1'b0) ? '0 : data_i;
+
+        if(!not_first_r)
+          not_first_r <= 1'b1;
+        else begin
+          w_en_r <= (addr_r == addr_i) ? 
+            (!w_i && !w_r) ? w_en_r : 1'b1 :
+            w_i;
+          if(addr_r == addr_i)
+            r_data_r <= w_r ? (r_data_r & ~w_mask_r | w_data_r & w_mask_r) : r_data_r;
```
dpetrisko, line 97: This seems incorrect. You're driving r_data_r from multiple process blocks. Probably should always just be the read data from the ram. Can use bsg_dff_en_bypass to simplify the logic

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,107 @@
...
+  logic                     w_r;
+ 
+  logic                     not_first_r; //for first r_en
+
+  // Infers an SDP BRAM
+  (* ram_style = "block" *) logic [width_lp:0] ram [els_p-1:0];
+
+  wire masked   = (w_mask_i != '1) & w_i;
+  wire w_en     = v_i & (addr_r != addr_i) & w_en_r;
+  wire r_en     = v_i & (masked | !w_i) & ((addr_r != addr_i) | !not_first_r);
```
dpetrisko, line 62: What would happen if addr_r is reset to 0, and it spuriously matches? We have garbage in the data register. But then we would overwrite it with the new write data. So what's the advantage of not_first_r?

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,107 @@
...
+        w_r     <= w_i;
+        addr_r  <= addr_i;
+        w_mask_r <= ((addr_r == addr_i) & w_i == 1'b0) ? '0 : w_mask_i;
+        w_data_r <= ((addr_r == addr_i) & w_i == 1'b0) ? '0 : data_i;
+
+        if(!not_first_r)
+          not_first_r <= 1'b1;
+        else begin
+          w_en_r <= (addr_r == addr_i) ? 
+            (!w_i && !w_r) ? w_en_r : 1'b1 :
```
dpetrisko, line 94: Look to simplify these expressions

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,107 @@
...
+  logic [width_lp:0]        w_mask_r;
+  logic                     w_en_r;
+  logic                     w_r;
+ 
+  logic                     not_first_r; //for first r_en
+
+  // Infers an SDP BRAM
+  (* ram_style = "block" *) logic [width_lp:0] ram [els_p-1:0];
+
+  wire masked   = (w_mask_i != '1) & w_i;
```
dpetrisko, line 60: Possible power optimization: We don't need to read the RAM on byte masks. Tradeoff depends on width of the RAM and activity factor whether the static power of the mask detection logic outweighs the dynamic power of the reads

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,107 @@
...
+     if(w_en) 
+        ram[addr_r] <= r_data_r & ~w_mask_r | w_data_r & w_mask_r;
+  end
+
+  always_ff @(posedge clk_i) begin
+    if(!reset_i) begin
+      not_first_r <= 1'b0;
+      w_en_r   <= 1'b0;
+      w_r      <= 1'b0;
+      w_data_r <= '0;
```
dpetrisko, line 77: I believe w_data_r and r_data_r can be combined into wbuf_r with care. Basically, by combining the read data of the ram with the write data directly to do the masked write, rather than first reading to rdata

mysoreanoop: So, we'll have r_data as a wire (with parameterised latching, separately), w_data_r for latching the data_i, and w_buf_r for masked write data (we'll be needing this in case of consecutive reads to different addresses), right?

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
mysoreanoop: true, no advantage; will remove it

taylor-bsg:
See coding guidelines here:

https://docs.google.com/document/d/1xA5XUzBtz_D6aSyIBQUwFk_kSUdckrfxa2uzGjMgmCU/edit

All signals should be high-true, including reset

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,94 @@
...
+);
+  
+  logic [addr_width_lp-1:0] addr_r;
+  logic [width_lp:0]        r_data_r;
+  logic [width_lp:0]        w_data_r;
+  logic [width_lp:0]        w_mask_r;
+  logic                     w_en_r;
+  logic                     w_r;
+
+  // Infers an SDP BRAM
```
taylor-bsg, line 55: We should try the module without this; to give vivado the flexibility to choose based on resource utilization.
Moreover, this should be in the mainline bsg_mem directory, since this could be potentially useful for other backends that don't support masking. Then you can instantiate this module from the zynq specific /hard implementation that is just a wrapper for this.

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
taylor-bsg: There is a separate module, bsg_mem_1rw_sync_mask_write_byte to handle this case if the byte granularity is known statically.
Let's keep it simple until we know that is case we will use.

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
mysoreanoop: Updated

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,94 @@
...
+  wire masked   = (w_mask_i != '1) & w_i;
+  wire w_en     = v_i & !same_addr & w_en_r;
+  wire r_en     = v_i & (masked | !w_i) & !same_addr;
+
+  always_ff @(posedge clk_i) begin
+     if(r_en)
+        r_data_r <= ram[addr_i];
+       
+     if(w_en) 
+        ram[addr_r] <= r_data_r & ~w_mask_r | w_data_r & w_mask_r;
```
taylor-bsg: parens to make operator ordering intent more clear. ( & )  | ( & )

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,94 @@
...
+        w_mask_r <= (same_addr & !w_i) ? '0 : w_mask_i;
+        w_data_r <= (same_addr & !w_i) ? '0 : data_i;
+        w_en_r   <= same_addr ? ((!w_i && !w_r) ? w_en_r : 1'b1) : w_i;
+        if(same_addr)
+          r_data_r <= w_r ? (r_data_r & ~w_mask_r | w_data_r & w_mask_r) : r_data_r;
+      end
+  end
+
+  always @(negedge clk_i)
+    assert(int'(addr_i) < els_p) else $warning("Accessing uninitialized address!");
```
taylor-bsg: always label the module source for your warnings with %m; see other parts of basejump STL

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
mysoreanoop: Updated

dpetrisko: > handle this case if the byte granularity is known statically.

yeah I’m talking about the dynamic case

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,116 @@
...
+    assert(int'(addr_i) < els_p) else $warning("%m Accessing uninitialized address!");
+
+  logic [addr_width_lp-1:0]      addr_r;
+  logic [width_lp:0]             w_data_r;
+  logic [width_lp:0]             w_mask_r;
+  logic                          w_r;
+  logic [width_lp:0]             r_data_r;
+  logic                          w_en_r;
+
+  logic [width_lp:0] ram [els_p-1:0];
```
taylor-bsg, line 58: this module should instantiate bsg_mem_1r1w_sync

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,116 @@
...
+  generate
+    integer ram_index;
+    initial
+      for (ram_index = 0; ram_index < els_p; ram_index = ram_index + 1)
+        ram[ram_index] = {(width_p){1'b0}};
+  endgenerate
+
+
+  wire same_addr = addr_r == addr_i;
+  wire masked    = w_i & (w_mask_i != '1);
```
taylor-bsg, line 69: let's get rid of this optimization; I think it is very unlikely

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,116 @@
+// 23 July, 2021
+//
+// Essentially, a 1RW synchronous-read RAM supporting bit-wise masked writes.
+// Some FPGAs do not inherently support synthesising 1RW RAMs with bit-wise masked writes.
```
taylor-bsg, line 4: // Some ASIC SRAM generators do not support bit-wise masked writes
// and many FPGAs do not inherently support synthesizing 1RW RAMs with bit-wise masked writes.

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
dpetrisko: I believe the cases are:
- read match, writeback wbuf and return wbuf
- read no match, writeback wbuf and return wbuf
- write match, write merge to wbuf
- write no match, writeback wbuf and write to wbuf

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
taylor-bsg: In addition to all of the benefits outlined in the style guide about why we abstract basic RAM instantiations, I think this will make the synthesis intent of the code a lot more clear as well to the tools.

taylor-bsg: @mysoreanoop

mysoreanoop:
Since the previous version was driving the r_data_r with two different signals, instantiating sync read RAM was a little ugly, so got rid of some of the logic. We could use asynchronous read RAM and retain the previous one, but it seems this variation is more elegant.

hard/ultrascale_plus/bsg_misc/bsg_mul_add_unsigned.v
```diff
@@ -24,7 +24,7 @@ module bsg_mul_add_unsigned #(
     ,output [width_o_p-1 : 0] o
     );
```
mysoreanoop, line 26: A typo

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -48,67 +41,70 @@ module bsg_mem_1rw_sync_mask_write_bit_from_1r1w #(
...
+      );
 
   always_ff @(posedge clk_i) begin
-    if(reset_i) begin
-      w_en_r   <= 1'b0;
-      w_r      <= 1'b0;
-      r_data_r <= '0;
-    end
-    else begin
+    if(verbose_p==1)
```
dpetrisko, line 73: put in its own always_ff block and surround with 

//synopsys translate_off
//synopsys translate_on

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,112 @@
...
+  , input  [ addr_width_lp-1:0]  addr_i
+  , input                        v_i
+  , input  [width_lp:0]          w_mask_i
+  , input                        w_i
+
+  , output [width_lp:0]          data_o
+);
+ 
+  always @(negedge clk_i)
+    assert(int'(addr_i) < els_p) else $warning("%m Accessing uninitialized address!");
```
dpetrisko, line 42: put in its own always_ff block and surround with

//synopsys translate_off
//synopsys translate_on

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,112 @@
...
+              , w_en_li, r_en_li, w_data_li, addr_r, haz, buf_r);
+
+    if(~reset_i) begin
+      haz_r <= haz;
+      w_r   <= v_i & w_i;
+      if(v_i) begin
+        addr_r <= addr_i;
+        if(w_i) begin
+          w_mask_r <= w_mask_i;
+          w_data_r <= data_i;
```
dpetrisko, line 84: I still don't believe you need a separate w_data_r buffer. See these slides: https://docs.google.com/presentation/d/1at_ug3g-5TBDDznidGXTgL3DeKY9ShWEUqt0bMsgMIM/edit#slide=id.p

mysoreanoop: Updated

mysoreanoop:
I've made provision for latch_last_read_p in the instantiated 1r1w RAM. 
We could instantiate the latched reads RAM in the current design to save some accesses; it works optimistically by delaying the write until a next valid request arives, so that if the address accessed matches, we could skip reading the RAM; could have some significant benefits when RAM is big, or if the consequent data accesses repeat the address frequently (which may be a possibility with tag RAMs, right?)

This version is the other kind which doesn't use the latched reads RAM, instead writes into RAM reliably in the 1st cycle if there's no hazard.

bsg_mem/bsg_mem_1r1w_sync.v
```diff
@@ -81,7 +83,7 @@ module bsg_mem_1r1w_sync #(parameter `BSG_INV_PARAM(width_p)
    always_ff @(posedge clk_lo)
      if (w_v_i)
        begin
-          assert ((reset_i === 'X) || (reset_i === 1'b1) || (w_addr_i < els_p))
+          assert ((reset_i === 'X) || (reset_i === 1'b1) || (int'(w_addr_i) < els_p))
```
taylor-bsg, line 86: curious about the reason for this cast?

mysoreanoop: Failed to notice it, didn't mean to keep it; basically verilator complains that passed els_p is integer of 32 bits, but w_addr_i is of different width. Can be solved with verilator's `-Wno-WIDTH`
Will revert it in the next commit.

taylor-bsg: @dpetrisko

bsg_mem/bsg_mem_1r1w_sync_synth.v
```diff
@@ -73,12 +74,37 @@ module bsg_mem_1r1w_sync_synth #(parameter `BSG_INV_PARAM(width_p)
      else
        r_addr_r <= 'X;
 
-   assign r_data_o = mem[r_addr_r];
+   wire [width_p-1:0] data_out = mem[r_addr_r];
```
taylor-bsg, line 77: mem_out might be a better name for this, to avoid confusion with data_o

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,139 @@
+// 23 July, 2021
+//
+// Some ASIC SRAM generators do not support bit-wise masked writes,
+// and many FPGAs do not inherently support synthesizing 1RW RAMs with bit-wise masked writes.
+// In such cases, where, also, a 1R1W RAM is on the offer,
```
taylor-bsg, line 5: more succint.

// We can implement these with 1R1W RAMs, if available.

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,139 @@
...
+`include "bsg_defines.v"
+
+module bsg_mem_1rw_sync_mask_write_bit_from_1r1w #(
+  parameter    `BSG_INV_PARAM(width_p)
+  , parameter  `BSG_INV_PARAM(els_p)
+  , parameter  latch_last_read_p     = 0
+  , parameter  enable_clock_gating_p = 0
+  , parameter  verbose_p             = 0
+  , localparam addr_width_lp         = `BSG_SAFE_CLOG2(els_p)
+  , localparam width_lp              = `BSG_SAFE_MINUS(width_p,1)
```
taylor-bsg, line 40: having two variables with the same name but just a different suffix but the same value begs for confusion, although I understand you are using this idiom in a very specific context. I suggest width_m1_lp (m1 = minus 1)

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,139 @@
...
+      );
+    end
+  else
+    begin: no_llr
+      assign data_o = data_out;
+    end
+
+endmodule
+
+//`BSG_ABSTRACT_MODULE(bsg_mem_1rw_sync_mask_write_bit_from_1r1w)
```
taylor-bsg, line 139: I believe this module IS abstract

bsg_mem/bsg_mem_1r1w_sync.v
dpetrisko: Yeah it’s not guaranteed that els will be an int unless explicitly declared as one (although for all practical purposes it will be). Slightly safer to cast els_p to addr width. 

Curious about the goal of this assertion though. Since address width is derived from els, we can’t really overflow. Is it for the case of NPOT els?

dpetrisko:
Do BRAMs support this latch last read behavior in 1R1W? I believe we don’t have it in our RAMs because SRAMs are inconsistent in their behavior.

For example, what happens if you read foo, and then write foo? Does the output value correspond to the new value or the old?

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,139 @@
...
+  , input                        w_i
+
+  , output [width_lp:0]          data_o
+);
+
+  // synopsys translate_off
+  always @(negedge clk_i)
+    assert(int'(addr_i) < els_p) else $warning("%m Accessing uninitialized address!");
+  // synopsys translate_on
+
```
taylor-bsg, line 58: For this section, the code is mostly there; but the readability could be better:

a. There are a lot of forward references for complex signals (i.e. not just pipelined versions of inputs, but expressions), and so to understand it, you have to jump around a lot. Generally, we want to structure the code so that there are as few forward references as possible. 

b. I like the enables you put in for clock gating; we don't usually do this in the group although maybe we should start thinking about it. Stylistically, I think it's good to indicate when it's there for clock gating reasons versus functional reasons -- perhaps we will create a new line of bsg_dff's for this purpose! 

c. One last detail; I think if you use the abstraction of a bypassing memory on a read-after-write hazard, then it simplifies the code a lot; and you don't need to have these multiple variants of a match. 

d. comments to highlight your intent will also help the reader a lot

Here is an incomplete sketch that tries to illustrate some of these concepts; you can cross compare: 

// we should bypass if the second stage is writing
// the data we are about to read
  
  wire match_n  = v_and_w_r & (w_addr_r == r_addr_li);
  wire bypass_n = v_i & match_n;
  wire bypass_r;
  
  wire [width_p-1:0] mem_data_lo, bypass_data_r;
  wire [addr_width_lp-1:0] addr_r;
  
  // pipeline inputs that are valid
  // for either reads or writes
  
  bsg_dff_en #(.width_p(...))
  (.clk_i(clk_i)
   ,.en_i(v_i) // clock gating
   ,.data_i({bypass_n, addr_i, v_and_w_n})
   ,.data_o({bypass_r, addr_r, v_and_w_r})
  );
  
  // pipeline inputs that are valid for writes
  bsg_dff_reset_en #(.width_p(...))
  (.clk_i(clk_i)
   ,.en_i(v_and_w_n) // clock gating
   ,.data_i({mask_i, data_i})
   ,.data_o({mask_r, data_i_r})
  );
 
  bsg_dff_en #(.width_p(...))
  (.clk_i(clk_i)
   ,.en_i(w_v_li & bypass_n) // clock gating
   ,.data_i(w_data_li)
   ,.data_o(bypass_data_r)
  );
  
 // only read if the next pipeline stage is not writing the same address    
 bsg_mem_1r1w_sync
    #(.width_p   (width_p)
      ,.els_p    (els_p)
      ,.latch_last_read_p(0)
      ) ram
      (.clk_i    (clk_i)
      ,.reset_i  (reset_i)
      ,.w_v_i    (v_and_w_r)
      ,.w_addr_i (w_addr_r)
      ,.w_data_i (w_data_li)
       // don't read if we are going to bypass
      ,.r_v_i    (v_i & ~(match_n))
      ,.r_addr_i (addr_i)
       ,.r_data_o (mem_data_lo)
      );
  
  // bypass in data if we are reading
  assign data_o = bypass_r ? bypass_data_r : mem_data_lo;
  
  wire [width_p-1:0] w_data_li = (data_o & ~mask_r) | (data_i_r & mask_r);


endmodule

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
mysoreanoop: Will move in the final commit

bsg_mem/bsg_mem_1r1w_sync.v
mysoreanoop: For NPOT els_p, yes. 
els_p is usually one more than $bits(w_addr_i); we'll then have to cast both to $bits(w_addr_i) + 1.

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
mysoreanoop: I'm somehow unable to get Verilator to recognize the BSG_INV_PARAM (like els_p) with this uncommented, not sure why. Should we do something in Verilator to get it to recognize BSG_INV_PARAM?

mysoreanoop:
> Do BRAMs support this latch last read behavior in 1R1W? I believe we don’t have it in our RAMs because SRAMs are inconsistent in their behavior.
> 
> For example, what happens if you read foo, and then write foo? Does the output value correspond to the new value or the old?

Should be old value, right? Unless you read post write post read.

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
mysoreanoop: I've addressed most of the points, I feel. Do we really want so many DFF variants here?

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
dpetrisko: You need a recent version of Verilator. I added support to it ~a month ago for this

dpetrisko:
> Should be old value, right? Unless you read post write post read.

this is why I’m asking. It also could be X. Either way, since it’s not used in the PR I would remove it

testing/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w/vsim/Makefile
```diff
@@ -0,0 +1,104 @@
...
+# 
+# This makefile is a skeleton that simulates a module many times
+# trying out variations of a product of different command line parameters.
+#
+# MBT 5/12/2015
+#
+# Edited to test bsg_mem_1rw_sync_mask_write_bit_from_1r1w
+
+TOP = ../../../..
+REPLACE = bsg_mem_1rw_sync_mask_write_bit_from_1r1w
```
dpetrisko, line 11: I would just use `define

testing/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w/vivado/test_bsg.v
```diff
@@ -0,0 +1,116 @@
...
+  //For cleaner testing iteration, uncomment this and initialize DUT RAM as well
+  generate
+    integer ram_index;
+    initial
+      for (ram_index = 0; ram_index < `ELS_P; ram_index = ram_index + 1)
+        ram[ram_index] = {(`WIDTH_P){1'b0}};
+  endgenerate
+
+  bsg_mem_1rw_sync_mask_write_bit_from_1r1w 
+  #(`WIDTH_P
```
dpetrisko, line 67: use named parameters

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,171 @@
...
+    );
+
+  wire [addr_width_lp-1:0]  addr_r;
+  wire                      haz = (addr_r == addr_i);
+  wire                      bypass_n = haz & v_and_w_r;
+  wire                      bypass_r;
+
+  bsg_dff_en
+    #(.width_p(addr_width_lp+1))
+    common_pipe
```
dpetrisko, line 77: nit, 'reg' rather than pipe

dpetrisko: throughout

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,170 @@
...
+
+  wire  [width_m1_lp:0]     mem_data_lo;
+
+  // If WAW, write bypass_data_r which incorporates both updates
+  // else write updated mem_data_lo
+  wire  [width_m1_lp:0] w_data_li = (bypass_r & v_and_w_r)
+          ? bypass_data_r : (mem_data_lo & ~w_mask_r) | (bypass_data_r & w_mask_r);
+
+  // Read if either RAR/WAR hazard or not a hazard
+  wire r_en_li = v_i & (haz & ~v_and_w_r | ~haz);
```
taylor-bsg, line 114: v_i & ~bypass_n

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
taylor-bsg: The sketch I wrote above uses 6 | or & operators, but yours uses 17 -- why?

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,170 @@
...
+
+  // Write previous if not a write hazard currently
+  wire w_en_li = v_and_w_r & ~(v_i & w_i & haz);
```
taylor-bsg, line 117: the code is simpler if you always write, if there is a write to do; by definition the write is the latest version of the data, and you can bypass the value over for either RAW or WAW hazards..

taylor-bsg:
@dpetrisko 

> Do BRAMs support this latch last read behavior in 1R1W? I believe we don’t have it in our RAMs because SRAMs are inconsistent in their behavior.
> 
> For example, what happens if you read foo, and then write foo? Does the output value correspond to the new value or the old?

I looked, and the latching of read outputs appears to be pretty ubiquitous across ARM and TSMC rams. Apparently as well in Virtex Ultrascale+ and 7-series block rams:

p.13 https://www.xilinx.com/support/documentation/user_guides/ug573-ultrascale-memory-resources.pdf
"EN[A|B] When inactive, no data is written to the block RAM and the output bus remains
in its previous state."

dpetrisko:
> All output ports are latched or registered (optional). The state of the output port does not change until the port executes another read or write operation. The default
block RAM output is register mode.

pg 8. This seems to imply that doing a read, then a write on the same port should X the output register

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
mysoreanoop: 1. r_en / w_en optimizations -- 5 gates extra in my case. I got rid of this optimization, it's not really helping much; we could keep this as a parameter based optimization if you think it helps.
2. Your case uses a separate data_i_r and bypass_data_r (saves 3 extra gates and 3 muxes). PS this was the version I had 2 commits ago. But as Dan smartly pointed out, even in case of a critical WAW, we can do without this data_i_r (previously called w_data_r) like so: [sans_data_i_r](https://wavedrom.com/editor.html?%7Bsignal%3A%20%5B%0A%20%20%7Bname%3A%20%27clk%27%20%20%20%09%2C%20wave%3A%20%27p....%27%7D%2C%0A%20%20%7Bname%3A%20%27addr%27%20%20%09%2C%20wave%3A%20%27x2.x.%27%20%2C%20data%3A%20%220%22%7D%2C%0A%20%20%7Bname%3A%20%27w_i%27%20%20%20%09%2C%20wave%3A%20%27x1.x.%27%7D%2C%0A%20%20%7Bname%3A%20%27data_i%27%20%20%09%2C%20wave%3A%20%27x48x.%27%20%2C%20data%3A%20%5B%221%22%2C%20%222%22%5D%7D%2C%0A%20%20%7Bname%3A%20%27bypass_n%27%09%2C%20wave%3A%20%27x.1x.%27%7D%2C%0A%20%20%7Bname%3A%20%27bypass_r%27%09%2C%20wave%3A%20%27x..1x%27%7D%2C%0A%20%20%7Bname%3A%20%27buf_r%27%20%20%20%20%2C%20wave%3A%20%27x.42x%27%20%2C%20data%3A%20%5B%221%22%2C%20%224%22%5D%7D%2C%0A%20%20%7Bname%3A%20%27w_data%27%20%09%2C%20wave%3A%20%27x.52x%27%20%2C%20data%3A%20%5B%223%22%2C%20%224%22%5D%7D%2C%0A%0A%0A%5D%7D%0A). The PDF linked by Dan is more informative. Having a dedicated data_i_r helps simplify the logic, like so: [with_data_i_r](https://wavedrom.com/editor.html?%7Bsignal%3A%20%5B%0A%20%20%7Bname%3A%20%27clk%27%20%20%20%09%2C%20wave%3A%20%27p....%27%7D%2C%0A%20%20%7Bname%3A%20%27addr%27%20%20%09%2C%20wave%3A%20%27x2.x.%27%20%2C%20data%3A%20%220%22%7D%2C%0A%20%20%7Bname%3A%20%27w_i%27%20%20%20%09%2C%20wave%3A%20%27x1.x.%27%7D%2C%0A%20%20%7Bname%3A%20%27data_i%27%20%20%09%2C%20wave%3A%20%27x48x.%27%20%2C%20data%3A%20%5B%221%22%2C%20%222%22%5D%7D%2C%0A%20%20%7Bname%3A%20%27bypass_n%27%09%2C%20wave%3A%20%27x.1x.%27%7D%2C%0A%20%20%7Bname%3A%20%27bypass_r%27%09%2C%20wave%3A%20%27x..1x%27%7D%2C%0A%20%20%7Bname%3A%20%27buf_r%27%20%20%20%20%2C%20wave%3A%20%27x.42x%27%20%2C%20data%3A%20%5B%221%22%2C%20%224%22%5D%7D%2C%0A%20%20%7Bname%3A%20%27w_data%27%20%09%2C%20wave%3A%20%27x.52x%27%20%2C%20data%3A%20%5B%223%22%2C%20%224%22%5D%7D%2C%0A%0A%0A%5D%7D%0A). Basically, it's a tradeoff between indulging registers or logic, from what I understand.

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
mysoreanoop: Addressed in another comment

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
mysoreanoop: Addressed in another comment

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -123,39 +118,40 @@ module bsg_mem_1rw_sync_mask_write_bit_from_1r1w #(
     (.clk_i    (clk_i)
     ,.reset_i  (reset_i)
     ,.w_v_i    (w_en_li)
-    ,.w_addr_i (addr_r)
+    ,.w_addr_i (w_addr_r)
     ,.w_data_i (w_data_li)
     ,.r_v_i    (r_en_li)
     ,.r_addr_i (addr_i)
     ,.r_data_o (mem_data_lo)
     );
```
mysoreanoop, line 127: moved closer to definition

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,166 @@
...
+  // If WAW, write bypass_data_r which incorporates both updates
+  // else write updated mem_data_lo
+  wire  [width_m1_lp:0] w_data_li = bypass_r
+          ? bypass_data_r : (mem_data_lo & ~w_mask_r) | (bypass_data_r & w_mask_r);
+
+  always_ff @(posedge clk_i)
+    if(~reset_i)
+      bypass_data_r <= (v_and_w_r & bypass_n)
+                 ? (w_i
+                   ? (w_data_li & ~w_mask_i) | (data_i & w_mask_i)
```
taylor-bsg, line 105: add sidecomment: // WAW

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,166 @@
...
+                   : w_data_li)
```
taylor-bsg, line 106: add sidecomment: // RAW

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,166 @@
...
+                 : data_i;
```
taylor-bsg, line 107: add sidecomment: // no hazard

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,166 @@
...
+
+  logic [width_m1_lp:0]     bypass_data_r;
+
+  // If WAW, write bypass_data_r which incorporates both updates
+  // else write updated mem_data_lo
+  wire  [width_m1_lp:0] w_data_li = bypass_r
+          ? bypass_data_r : (mem_data_lo & ~w_mask_r) | (bypass_data_r & w_mask_r);
+
+  always_ff @(posedge clk_i)
+    if(~reset_i)
```
taylor-bsg, line 102: not sure what ~reset_i is accomplishing here?

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,166 @@
...
+      bypass_data_r <= (v_and_w_r & bypass_n)
```
taylor-bsg, line 103: v_and_w_r is redundant

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,166 @@
...
+    #(.width_p(width_p))
+    write_regs
+    (.clk_i  (clk_i)
+    ,.en_i   (v_i & w_i)
+    ,.data_i (w_mask_i)
+    ,.data_o (w_mask_r)
+    );
+
+  wire  [width_m1_lp:0]     mem_data_lo;
+
```
taylor-bsg, line 93: // if there are no back-to-back RAW or WAW dependence, then this contains the
// data to write. If there is a back-to-back dependence, then this contains the latest
// running value of this memory address.

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,166 @@
...
+  , output [width_m1_lp:0]       data_o
+);
+
+  // synopsys translate_off
+  always @(negedge clk_i)
+    assert(int'(addr_i) < els_p) else $warning("%m Accessing uninitialized address!");
+  // synopsys translate_on
+
+  wire                      v_and_w_r;
```
taylor-bsg, line 58: wire v_and_w_n  = v_i & w_i;

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,166 @@
...
+  bsg_dff
+    #(.width_p(1))
+    common_next_regs
+    (.clk_i  (clk_i)
+    ,.data_i (v_i & w_i)
+    ,.data_o (v_and_w_r)
+    );
+
+  wire [addr_width_lp-1:0]  w_addr_r;
+  wire                      bypass_n = v_and_w_r & (w_addr_r == addr_i);
```
taylor-bsg, line 69: wire addr_match = (w_addr_r == addr_i);

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
```diff
@@ -0,0 +1,166 @@
...
+      bypass_data_r <= (v_and_w_r & bypass_n)
+                 ? (w_i
+                   ? (w_data_li & ~w_mask_i) | (data_i & w_mask_i)
+                   : w_data_li)
+                 : data_i;
+
+  // Read if either RAR/WAR hazard or not a hazard
+  // Write previous if not a write hazard currently
+  wire r_en_li = v_i & ~bypass_n;
+  wire w_en_li = v_and_w_r;
```
taylor-bsg, line 111: // to make it consistent with the above 

wire w_en_li = v_and_w_r & v_and_w_i & addr_match;

mysoreanoop: And r_en_li from `r_en_li = v_i & ~bypass_n;`
to `r_en_li = v_i & (addr_match & ~v_and_w_r | ~addr_match);`?

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
taylor-bsg: Okay I see.  In ASIC, flip-flop internal power is about 2X the MUX (applies only when not clock gated), and the data input is about 1.25X), and area is about 2X so I think the benefit is pretty marginal if we are adding multiple multiplexers to eliminate one flop.

taylor-bsg: I am okay with either implementation, as long as the more complex one has the comments I suggested. But maybe something to thing about for future codes. The real bottleneck is verification, so if a simpler implementation means somebody has to stare at the code less, that is the major win.

dpetrisko: 2x seems low for overheads. If the enable flop is implemented with a mux, adding a mux to eliminate the flop is basically straight savings.

Agreed on verification, but this is a module that’s very easy to get to 100% coverage

taylor-bsg: The enable on the flop is solely for clock gating, doesn't turn into muxes.

taylor-bsg:
> > All output ports are latched or registered (optional). The state of the output port does not change until the port executes another read or write operation. The default
> > block RAM output is register mode.
> 
> pg 8. This seems to imply that doing a read, then a write on the same port should X the output register

pg 16. "In NO_CHANGE mode, the output latches remain unchanged during a write operation. As
shown in Figure 1-4, data output remains the last read data and is unaffected by a write
operation on the same port. T"

dpetrisko:
> pg 16. "In NO_CHANGE mode, the output latches remain unchanged during a write operation. As
shown in Figure 1-4, data output remains the last read data and is unaffected by a write
operation on the same port. T"

works for me, long as that is the case for tsmc / arm rams as well. I still say we should make a new PR, because it involves updating the hardened rams with the new parameter, etc

taylor-bsg:
The idea with latch last read is that it can always be emulated by adding a
flop in the hardened implementations (much as Anoop is doing here); and
anecdotally it does appear that this feature is widely supported, so it
seems reasonable to expose it as part of the ram interface.

M

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.v
mysoreanoop: > I am okay with either implementation, as long as the more complex one has the comments I suggested. But maybe something to thing about for future codes. The real bottleneck is verification, so if a simpler implementation means somebody has to stare at the code less, that is the major win.

Understood, I'll keep that in mind. I've retained the version you reviewed.

mysoreanoop:
I can move this to /bsg_mem/ if there are no more corrections

taylor-bsg:
Looks great Anoop! Yes, let's move to /bsg_mem/. Congrats on this excellent module.

mysoreanoop:
Thank you. Please let me know if the testbenches need any changes.

taylor-bsg (approved):

taylor-bsg:
@mysoreanoop merged; thank you Anoop!

PR #475, opened 2021-07-27
Adding hardened sram generator script

dpetrisko:
This is a first pass at a more user-friendly SRAM flow. It generates the hardened RAM based on a memgen.json file with a similar one to our current CAD flows. There are several open questions about what fields to require, default and exclude from said json file, as well as how to standardize on naming and parameters, but I think this is the first step.

Design doc here: https://docs.google.com/document/d/1aqRVsSbKJlbX_TA-xT7cd4KW68SlMt1lBHkePMOUktk/edit#


stdavids (commented):
The generators are all very similar, feels like some refactoring would eliminate quite a bit of code. Something like splitting all of the template strings into different files and then having a single generator.

Do we need the pre-generated .v files in common/bsg_mem?

dpetrisko:
>Do we need the pre-generated .v files in common/bsg_mem?

I don't think so. I left them so that existing tapeout configurations which rely on them could continue to work. However, with the philosophy that tapeouts are "frozen in time" with a specific basejump version, it's probably unnecessary

hard/common/bsg_mem/bsg_mem_1r1w_sync.v
```diff
@@ -0,0 +1,59 @@
+
+  `include "bsg_mem_1r1w_sync_macros.vh"
```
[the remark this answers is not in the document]
dpetrisko: That's an interesting point, maybe @taylor-bsg has an opinion. The macros are extremely coupled to the memory generator specifics. Having them in basejump makes it easier to reason about what's going on, but there's nothing saying that you couldn't have another gf14 memory generator semantic (especially for naming).

In this case, we'd have a header files as an input in addition to the memgen.json. Should be straightforward to implement

dpetrisko:
>The generators are all very similar, feels like some refactoring would eliminate quite a bit of code. Something like splitting all of the template strings into different files and then having a single generator.

That's something I feel should push to another PR, for when someone feels like some code golf. I foresee a more urgent PR which unifies the bsg_mem parameter/input space (write_mask vs w_mask, width_p vs data_width_p). But I don't want to foist too many breaking changes here

dpetrisko:
@taylor-bsg @stdavids This should be good to go, need to merge https://github.com/bespoke-silicon-group/bsg_14/pull/133 at the same time, though. I tested out the full flow on BP and it's working fine

stdavids:
Looks good, it would be nice if the readme went a bit more in depth about the role of the macros.vh files and maybe spoke to the new organization of the hardened memories flow, like where the ram templates are and how the combo of common/ and <process>/ files come together.

stdavids (approved):

dpetrisko:
Updated the README!

dpetrisko:
@taylor-bsg maybe take a look at the README, as someone with less context?

stdavids (approved):
Like the readme improvements!

taylor-bsg:
@tommydcjung can you review?

tommydcjung:
```
      if (latch_last_read_p && !True)
        $error("BSG ERROR: latch_last_read_p is set but unsupported");
```

is "True" supposed to be a SV keyword? I'm not sure what's going on there.

dpetrisko:
> if (latch_last_read_p && !True)
        $error("BSG ERROR: latch_last_read_p is set but unsupported");

Oh this looks like leaky python. I've been running this through synthesis, so the //synopsys translate_off was masking it.  Will fix

dpetrisko:
Fixed!

hard/common/bsg_mem/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+            `bsg_mem_1rw_sync_macro(512, 64, 2) else
+
+      will result in a 512x64 (mux 2) RAM taking the place of any instantiated 512x64 RAMs in the chip. The fallthrough default case is to synthesize the RAM.
+
+- The actual hardened SRAM macro. These should be named as per the scheme in the bsg\_mem\_\*.vh macro headers and their .lib / .v views should be accessible to the CAD tools as needed.
+
+Because the SRAMs generated depends on each project's requirements, it is most convenient if this wrapper is generated as well. We provide a generic generator script in this directory which can be used to generate a set of these SRAM wrappers which is process agnostic (The process information is included based on the .vh macro headers). The input to this generator script (bsg\_mem\_generator.py) is a json file. We have a sample memgen.json in this directory. This same memgen.json can be used for feeding the SRAM generator itself, which is useful for keeping front-end and back-end in sync. An example line from the memgen.json is:
+
+        {"ports": "1rw" , "width":  128, "depth": 1024, "mux": 2, "type": "1rf", mask: 1, "adbanks": 2, "awbanks": 2},
```
tommydcjung, line 17: Isn't there a parameter for # of banks? I think you need to add that as well

dpetrisko: Yeah so that parameter is relevant for the bsg_14 flow, but we're not differentiating the SRAM wrapper for that. i.e. banks=1,2,4 all map to gf14_1rw_d128_w64_m2_sync_2rf. So it's basically just ignored by this generator

dpetrisko:
Other comments? Good to go?

taylor-bsg:
@tommydcjung @stdavids we good on this PR?

tommydcjung:
how does it handle special cases like 30-bit icache SRAM or explicitly instantiating delay cells at the data pins?

dpetrisko:
>how does it handle special cases like 30-bit icache SRAM or explicitly instantiating delay cells at the data pins?

The old flow (v/hard) is still supported, so you would copy the output of the generator in soft_ip, modify it, and point your filelist to that instead

tommydcjung:
Nice, that's what I needed to know

tommydcjung (approved):

stdavids:
Yep, still looks good to me

dpetrisko:
Thanks for the reviews!

PR #478, opened 2021-07-29
Update bsg_nonsynth_dpi_gpio.v

dpetrisko:
typo


drichmond (approved):
Yeah, whoops. Is a format string that provides _just_ the module name?

PR #480, opened 2021-08-03
Separating out addr and tag in bsg_cache_to_axi

dpetrisko:
Currently, we embed the hash function from multiple caches to AXI channels inside the module. This is not flexible enough to handle multiple schemes, which might correspond to different topologies, striping schemes or different ratios of caches to channels. This PR separates out the raw AXI addresses and cache tags, so that a user can implement their own hash function after the AXI conversion.

Usage seen here: https://github.com/black-parrot/black-parrot/commit/be10baeb7565b4576262b73ed6f1439e42567bc3

### DIFF
--- a/bsg_cache/bsg_cache_to_axi_rx.v
+++ b/bsg_cache/bsg_cache_to_axi_rx.v
@@ -95,13 +95,13 @@ module bsg_cache_to_axi_rx
   // axi read address channel
   //
   assign axi_arid_o = {axi_id_width_p{1'b0}};
-  assign axi_araddr_addr_o = addr_i;
+  assign axi_araddr_addr_o = axi_addr_width_p'(addr_i);
   assign axi_araddr_cache_id_o = cache_id_i;
   assign axi_arlen_o = (8)'(axi_burst_len_p-1); // burst length
   assign axi_arsize_o = (3)'(`BSG_SAFE_CLOG2(axi_data_width_p>>3));
   assign axi_arburst_o = 2'b01;   // incr
   assign axi_arcache_o = 4'b0000; // non-bufferable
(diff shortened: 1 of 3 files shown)


tommydcjung (approved):

taylor-bsg:
Looks good but maybe we can use something besides tag, since it's not a tag. cache_id?

taylor-bsg:
@drichmond take note

dpetrisko:
yeah cache_id works, updated

bsg_cache/bsg_cache_to_axi_rx.v
```diff
@@ -93,7 +95,8 @@ module bsg_cache_to_axi_rx
   // axi read address channel
   //
   assign axi_arid_o = {axi_id_width_p{1'b0}};
-  assign axi_araddr_o = axi_addr_i;
+  assign axi_araddr_addr_o = addr_i;
```
tommydcjung, line 98: potential width mismatch

taylor-bsg: good catch

dpetrisko: I added an explicit width cast, which handles the AXI address > cache address case. Should we have an assertion for axi_addr_width_p >= addr_width_p? Auto-truncation is scary behavior

taylor-bsg: I think it should just return the original bits in the original length, and leave it up to the instantiator to cast it to the right size.

dpetrisko: Okay, the address now passes through transparently

taylor-bsg (approved):

PR #483, opened 2021-08-04
Update tag clients

dpetrisko:


dpetrisko:
These are residual tag clients from #466 and #460

taylor-bsg (approved):

PR #485, opened 2021-08-07
Adding BSG_INV_PARAM exception for yosys

adithyasunil26:
Yosys does not support bare default parameters so an exception will have to be added for it as well.

### DIFF
--- a/bsg_misc/bsg_defines.v
+++ b/bsg_misc/bsg_defines.v
@@ -34,6 +34,8 @@
 
 `ifdef XCELIUM // Bare default parameters are incompatible as of 20.09.012
 `define BSG_INV_PARAM(param) param = "inv"
+`elsif YOSYS // Bare default parameters are incompatible as of 0.9
+`define BSG_INV_PARAM(param) param = "inv"
 `else // VIVADO, DC, VERILATOR, GENUS
 `define BSG_INV_PARAM(param) param
 `endif


dpetrisko (approved):

taylor-bsg:
Adithya, did you get any of the tests to run through Yosys? I wasn't sure of the extent to which Yosys supports SystemVerilog.

adithyasunil26:
@taylor-bsg I have not tried running tests yet. I attempted to synthesize with Yosys through FuseSoC and this change in the `bsg_defines` is necessary to make that possible.

Yosys has limited support for SystemVerilog i.e., things like `assert`, `always_ff`, `logic`, etc are supported. Although all BaseJump modules may not be synthesizable with Yosys. One of the things I have identified to cause problems with Yosys are multidimensional arrays. I had a discussion with @dpetrisko  yesterday and we came to the verdict that adding Yosys targets in the FuseSoC cores will help us identify what can be synthesized in Yosys so it is something worth trying. Would be great to hear your thoughts on this.

dpetrisko:
One other option would be to pickle the entirety of basejump with bsg_sv2v and the yosys target could run off that verilog-2005 file. I know @taylor-bsg has looked at pickling basejump as a single library file in the past

dpetrisko:
good to merge? While yosys doesn't support most basejump code, this will allow it to compile anything that it happens to like

PR #490, opened 2021-08-30
[bsg_mem_multiport_latch] Add write data clk gate param

tommydcjung:


bsg_mem/bsg_mem_multiport_latch.v
```diff
@@ -71,10 +73,18 @@ module bsg_mem_multiport_latch
 
 
   // write data latch 
+  logic w_data_clk;
```
taylor-bsg, line 76: use the clock gate module

PR #492, opened 2021-09-06
[WIP] Tsmc28

dpetrisko:
Preliminary PR for hardened TSMC 28 work. Interesting, the memory generator comes with 2RW SRAMs. I can make a synth version if we think there are generally useful, or else keep them in the hardened directory. It also has a hardened ROM generator, which seems difficult to generalize in a portable way, since we would need to pass the name as a parameter and then instantiate the correct backend at synthesis time. Curious if you have any suggestions there


dpetrisko:
Broken into #505 #504 #503 #475

PR #493, opened 2021-09-06
Update bsg_defines.v for GENUS

dpetrisko:


dpetrisko:
Hold off on this, it appears that variable is not always defined

dpetrisko:
NVM, CDS_TOOL_DEFINE is guaranteed by all CADENCE tools, GTG

PR #495, opened 2021-09-16
Guarding initial from synthesis

dpetrisko:
initials / assertions should not be present in synthesis


taylor-bsg (approved):

taylor-bsg (commented):
Usually a space between // and synopsys

dpetrisko:
>Usually a space between // and synopsys

fixed

PR #500, opened 2021-09-29
Changing info in bsg_mesh_router_buffered to assertion

dpetrisko:
Assertions are more flexible than displays i.e. they can be disabled and enabled during times of reset, as well as searched for in waveforms.

This is needed to prevent spurious reset errors in Verilator on the Manycore

### DIFF
--- a/bsg_noc/bsg_mesh_router_buffered.v
+++ b/bsg_noc/bsg_mesh_router_buffered.v
@@ -72,8 +72,8 @@ module bsg_mesh_router_buffered
 
       // synopsys translate_off
       always @(negedge clk_i)
-        if (link_o_cast[i].v)
-          $display("## warning %m: stubbed port %x received word %x",i,link_i_cast[i].data);
+        assert (~link_o_cast[i].v)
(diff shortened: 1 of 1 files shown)


bsg_noc/bsg_mesh_router_buffered.v
```diff
@@ -72,8 +72,8 @@ module bsg_mesh_router_buffered
 
       // synopsys translate_off
       always @(negedge clk_i)
-        if (link_o_cast[i].v)
-          $display("## warning %m: stubbed port %x received word %x",i,link_i_cast[i].data);
+        assert (~link_o_cast[i].v)
+          $warning("## stubbed port %x received word %x",i,link_i_cast[i].data);
```
tommydcjung, line 76: missing 'else'? also it should assert if (~reset_i)

dpetrisko: ah yeah that's what I get for using the web editor

taylor-bsg (approved):

PR #501, opened 2021-10-05
Fixing latch for verilator

dpetrisko:
Verilator 4.213 does not recognize assignments to outputs as a latch. The proposed patch should have no inference change on synthesis

### DIFF
--- a/bsg_misc/bsg_dlatch.v
+++ b/bsg_misc/bsg_dlatch.v
@@ -11,11 +11,14 @@ module bsg_dlatch #(parameter `BSG_INV_PARAM(width_p)
   if (i_know_this_is_a_bad_idea_p == 0)
     $fatal( 1, "Error: you must admit this is a bad idea before you are allowed to use the bsg_dlatch module!" );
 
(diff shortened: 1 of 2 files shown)


taylor-bsg:
Is this something we can file a bug report against verilator for?

dpetrisko:
Yeah trying to make a minimal testcase. Seems to be one of those things that disappears when reduced

bsg_misc/bsg_dlatch.v
```diff
@@ -11,12 +11,15 @@ module bsg_dlatch #(parameter `BSG_INV_PARAM(width_p)
...
 
+  logic [width_p-1:0] data_r;
   always_latch
     begin
       if (clk_i)
-        data_o <= data_i;
+        data_r <= data_i;
     end
 
+  assign data_o = data_r;
```
taylor-bsg, line 21: // Verilator Version XXX is unhappy if data_o is merged with data_r

taylor-bsg (approved):

PR #504, opened 2021-10-07
Adding clock gen for TSMC28

dpetrisko:
Adds the clock gen. Part of decomposition of #492. This hasn't been fully validated, but has gone through APR with SDP groups

### DIFF
--- a/hard/tsmc_28/bsg_clk_gen/README.verification
+++ b/hard/tsmc_28/bsg_clk_gen/README.verification
@@ -1,0 +1,24 @@
+This clock generator is intended to be able to change the frequency as the chip operates without glitches or short pulses.
+
+* Verification using SDF and gate-level simulation
+
+After the system has gone through place-and-route, it's important to check the timing corners in the circuit. The default tester tests the worst-case transition, which
+occurs when transitioning from the fastest frequency to the closest frequency.
+
+In gate-level simulation, this can be view by looking at clk_gen_osc_inst/adt/sel_r_reg_0, and finding the part in the test where the clock goes from fastest to slowest.
+Measure the distance between the D input and the CK signal. Typically D will fall shortly before CK rises, indicating a setup time condition that needs to be verified.
+This time can be compared against the negedge setup time for the same gate in the .SDF file.
+
+The other potential race is in the mux4i of the adt. The select lines should transition in a way that the output does not glitch. Currently the select lines transition
+from 3 (D)->2 (C)->0(A) after the D is already zero, C is hardwired to 0, and then A has just transitioned to 0.
+
+In practice, if the chip exhibits this race in the real world, it can be avoid by transitioning to an intermediate frequency before going to the slowest frequency.
+
+* Verification using Spice
+
+True transistor-level extraction-based spice is the best way to verify the circuit, although it takes a long time.  The simulation-based model in the spice directory
+can be used to eyeball frequency ranges.
+
+
+
+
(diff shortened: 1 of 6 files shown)


hard/tsmc_28/bsg_clk_gen/bsg_rp_clk_gen_atomic_delay_tuner.v
```diff
@@ -0,0 +1,99 @@
...
+   // synopsys rp_fill (0 0 RX)
+
+   // this gate picks input 01 when async reset is low, initializing the oscillator
+   IND2D2BWP7T40P140 NB (.A1(sel_r[0]), .B1(async_reset_neg_i), .ZN(sel_r[1]));
+   TIELBWP7T40P140   ZB (.ZN(zero_bit));
+
+   DFCND4BWP7T40P140 sel_r_reg_0 (.D(mux_lo[0]), .CP(o)      ,.CDN(async_reset_neg_i), .Q(sel_r[0]), .QN());
+   //LHCND4BWP7T40P140 sel_r_latch_0 (.D(mux_lo[0]), .E(o)      ,.CDN(async_reset_neg_i), .Q(sel_r[0]), .QN());
+
+   // 40nm: non-inverting mux 32.5ps + load S->Z
```
taylor-bsg, line 78: delete timing info (which is also not relevant)

hard/tsmc_28/bsg_clk_gen/README.verification
```diff
@@ -0,0 +1,24 @@
+This clock generator is intended to be able to change the frequency as the chip operates without glitches or short pulses.
```
taylor-bsg, line 1: Seems like these two files are replicated from 40 and 14; perhaps we should refactor them up to bsg_clk_gen; and rename the conflicting oscillator file with _nonsynth?

hard/tsmc_28/bsg_clk_gen/bsg_rp_clk_gen_atomic_delay_tuner.v
```diff
@@ -0,0 +1,99 @@
+// bsg_rp_clk_gen_coarse_delay_element
```
taylor-bsg, line 1: maybe add a readme file that says what the measured (non-downsampled range) range is for the oscillator

hard/tsmc_28/bsg_clk_gen/bsg_rp_clk_gen_atomic_delay_tuner.v
```diff
@@ -0,0 +1,99 @@
...
+//
+// o       contains controllably delayed signal
+//
+// module bsg_clk_gen_coarse_delay_element #(parameter `BSG_INV_PARAM(start_tap_p))
+//
+// DWP 10/6/2021: Ported to TSMC28 by reusing cells and appending 30P140 for the specific cell lib
```
taylor-bsg, line 7: probably want to say whether you ported from -- 14 or 40

hard/tsmc_28/bsg_clk_gen/README.verification
dpetrisko: Interesting idea. The bsg_clk_gen/bsg_clk_gen_osc.v is used in bsg_clk_gen/bsg_clk_gen.v which is used by both synth and nonsynth. So the portability layer is bsg_clk_gen_osc itself.

Maybe we can move it to hard/common/bsg_clk_gen/ ? For hardened modules which are "generic" between proc nodes?

dpetrisko: or are you suggesting bsg_nonsynth_clk_gen_osc.v would contain "bsg_clk_gen_osc" (not named nonsynth)?

dpetrisko:
Closing, we taped out v3 clock gen

PR #507, opened 2021-10-07
Removing defines from bsg_cache_non_blocking_pkg and bsg_dmc_pkg

dpetrisko:
In order for some tools to handle basejump code, each file needs to be able to parse as its own compilation unit. This means we can't rely on definitions from packages being available at import time.  This PR separates out bsg_cache_non_blocking_pkg and bsg_dmc_pkg defines into their own include files. Users should both import the package and include the defines

### DIFF
--- a/bsg_cache/bsg_cache_non_blocking.v
+++ b/bsg_cache/bsg_cache_non_blocking.v
@@ -10,7 +10,7 @@
  */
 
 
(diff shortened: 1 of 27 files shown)


taylor-bsg:
@tommydcjung good to go?

tommydcjung:
I think it's good, but you can try running the regression

PR #523, opened 2021-11-16
Adding fixed and wrap behavior to bsg_cache_to_axi

dpetrisko:
This PR adds FIXED and WRAP burst types to bsg_cache_to_axi, and bsg_nonsynth_axi_mem to test. There's a very basic testbench for demonstrating this wraparound feature


bsg_cache/bsg_cache_to_axi_rx.v
```diff
@@ -98,7 +99,7 @@ module bsg_cache_to_axi_rx
   assign axi_araddr_cache_id_o = cache_id_i;
   assign axi_arlen_o = (8)'(axi_burst_len_p-1); // burst length
   assign axi_arsize_o = (3)'(`BSG_SAFE_CLOG2(axi_data_width_p>>3));
-  assign axi_arburst_o = 2'b01;   // incr
+  assign axi_arburst_o = 2'(axi_burst_type_p);   // fixed, incr or wrap
```
taylor-bsg, line 102: add package and enum? do we have burst type defined anywhere?

dpetrisko: we have an axi package in BP, but it probably makes sense to move it to basejump

drichmond: We have (had?) some AXI packages in replicant too. I haven't touched them in forever, just a note.

dpetrisko: AXI is pretty slow-changing

dpetrisko: @taylor-bsg Updated with bare-bones axi package

dpetrisko:
good to go?

bsg_cache/bsg_cache_to_axi_tx.v
```diff
@@ -113,9 +115,9 @@ module bsg_cache_to_axi_tx
   assign axi_awaddr_cache_id_o = cache_id_i;
   assign axi_awlen_o = (8)'(axi_burst_len_p-1); // burst len
   assign axi_awsize_o = (3)'(`BSG_SAFE_CLOG2(axi_data_width_p>>3));
-  assign axi_awburst_o = 2'b01;   // incr
-  assign axi_awcache_o = 4'b0000; // non-bufferable
-  assign axi_awprot_o = 2'b00;    // unprivileged
+  assign axi_awburst_o = 2'b10;   // fixed, incr or wrap
```
farzamgl, line 118: Should this be fixed at wrap or set by the parameter?

bsg_cache/bsg_cache_to_axi_tx.v
```diff
@@ -139,11 +141,21 @@ module bsg_cache_to_axi_tx
...
+    ,.data_i(mask_i)
+    ,.data_o(mask_r)
+  );
+
   bsg_expand_bitmask #(
     .in_width_p(mask_width_p)
     ,.expand_p(byte_mask_width_lp/mask_width_p)
   ) expand (
-    .i(mask_i)
+    .i(mask_r)
```
farzamgl, line 158: Was this a bug related to the word-tracking change?

testing/bsg_cache/axi_test/testbench.v
```diff
@@ -0,0 +1,264 @@
...
+  //
+
+  parameter addr_width_p = 5;
+  parameter data_width_p = 8;
+  parameter block_width_p = 32;
+
+  parameter ring_width_p = data_width_p;
+  parameter rom_addr_width_p = 32;
+
+  parameter wrap_type_p = e_axi_burst_wrap;
```
farzamgl, line 21: Can we test the other modes with this testbench as well?

bsg_cache/bsg_cache_to_axi_tx.v
dpetrisko: I think for all bsg_cache applications, wrap is more appropriate. The current implementation of bsg_cache uses aligned bursts (for which wrap == incr) but other units using the same format (such as write buffers) may use wrap to leverage critical word first

testing/bsg_cache/axi_test/testbench.v
dpetrisko: Similar to above answer

bsg_cache/bsg_cache_to_axi_tx.v
dpetrisko: Yep, before the word tracking change, the wstrb was always hardcoded to '1 rather than coming from the dma_pkt

bsg_cache/bsg_cache_to_axi_tx.v
dpetrisko: Never mind I misunderstood your question. You're totally right, it should use the parameter. updated!

testing/bsg_cache/axi_test/testbench.v
dpetrisko: copying since I resolved the other question:

```
I think for all bsg_cache applications, wrap is more appropriate. The current implementation of bsg_cache uses aligned bursts (for which wrap == incr) but other units using the same format (such as write buffers) may use wrap to leverage critical word first
```

farzamgl (approved):

PR #524, opened 2021-11-18
Multiport write banked bypassing FF

tommydcjung:


bsg_mem/bsg_mem_multiport_write_banked_bypassing_ff.v
```diff
@@ -0,0 +1,132 @@
...
+    // mem_r sorted into banks
+    assign banked_mem[i%num_banks_p][i/num_banks_p] = mem_r[i];
+  end
+
+
+  // Read ports
+  logic [num_rs_p-1:0][num_banks_p-1:0][width_p-1:0] bank_r_data;
+  logic [num_rs_p-1:0][num_banks_p-1:0][width_p-1:0] bank_r_data_bypass;
+
+  for (genvar i = 0; i < num_rs_p; i++) begin: rs
```
taylor-bsg, line 92: perhaps we can:

1. create a submodule bsg_compare_bipartite, that given two arrays of A and B with potentially different numbers of items, returns 2D matrix of all-to-all compare between  A and B
2. create a module bsg_bypass_network that runs bsg_compare_bipartite on reads and writes, nor's the bits for each read and appends to bit 0; then uses it as control signals for  bsg_crossbar_o_by_i, where position 0 is the original read from the regfile; and the other positions are the writes

PR #525, opened 2021-11-22
fix clz output width

tommydcjung:
Also, fix when input bits are all zeros.


taylor-bsg:
What's the backstory on this? Needed for hardfloat correctness?

tommydcjung:
Yep. it's needed for matching with the hardfloat countLeadingZeros module.
https://github.com/bsg-external/HardFloat/issues/10

taylor-bsg (approved):

PR #531, opened 2021-12-07
Adding credit channels to channel tunnel

dpetrisko:
This PR adds the ability to have credit channels to the channel tunnel. This is necessary in the case of tunneling a large number of channels over a narrow channel with a large number of credits.

Ignore the testbench changes, will update Makefile.sim to use VCS before merging


dpetrisko:
Discussed offline to defer. Interesting improvement would be to serialize the split credit channel to reduce the CT arbitration logic

PR #534, opened 2021-12-11
Sdr tag pkg

dpetrisko:
This PR adds a tag package for SDR links. This helps standardize the chip interface for these modules, as these offsets can be included in other projects, etc.

### DIFF
--- a/bsg_link/bsg_link_pkg.v
+++ b/bsg_link/bsg_link_pkg.v
@@ -1,0 +1,18 @@
+
+package bsg_link_pkg;
+
+  import bsg_tag_pkg::*;
+
+  typedef struct packed
+  {
+    bsg_tag_s sdr_disable;
+    bsg_tag_s uplink_reset;
+    bsg_tag_s downlink_reset;
+    bsg_tag_s downstream_reset;
+    bsg_tag_s token_reset;
+  }  bsg_link_sdr_tag_lines_s;
+  localparam bsg_link_sdr_tag_local_els_gp =
+	$bits(bsg_link_sdr_tag_lines_s) / $bits(bsg_tag_s);
+
+endpackage
+
--- a/bsg_link/bsg_link_pkg.v
+++ b/bsg_link/bsg_link_pkg.v
@@ -11,7 +11,7 @@ package bsg_link_pkg;
     bsg_tag_s token_reset;
   }  bsg_link_sdr_client_tag_lines_s;
   localparam bsg_link_sdr_client_tag_local_els_gp =
-	$bits(bsg_link_sdr_tag_lines_s) / $bits(bsg_tag_s);
+	$bits(bsg_link_sdr_client_tag_lines_s) / $bits(bsg_tag_s);
 
   typedef struct packed
   {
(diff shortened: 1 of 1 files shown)


bsg_link/bsg_link_pkg.v
```diff
@@ -0,0 +1,28 @@
...
+
+  typedef struct packed
+  {
+    bsg_tag_s uplink_reset;
+    bsg_tag_s downlink_reset;
+    bsg_tag_s downstream_reset;
+    bsg_tag_s token_reset;
+  }  bsg_link_sdr_client_tag_lines_s;
+  localparam bsg_link_sdr_client_tag_local_els_gp =
+	$bits(bsg_link_sdr_tag_lines_s) / $bits(bsg_tag_s);
```
gaozihou, line 14: typo, should be
```
$bits(bsg_link_sdr_client_tag_lines_s) / $bits(bsg_tag_s);
```

bsg_link/bsg_link_pkg.v
```diff
@@ -0,0 +1,28 @@
...
+  typedef struct packed
+  {
+    bsg_tag_s sdr_disable;
+    bsg_tag_s uplink_reset;
+    bsg_tag_s downlink_reset;
+    bsg_tag_s downstream_reset;
+    bsg_tag_s token_reset;
+  }  bsg_link_sdr_master_tag_lines_s;
+  localparam bsg_link_sdr_master_tag_local_els_gp =
+	$bits(bsg_link_sdr_tag_lines_s) / $bits(bsg_tag_s);
```
gaozihou, line 25: Could we merge bsg_link_sdr_master_tag_lines_s with bsg_link_sdr_client_tag_lines_s?

In bsg_link_sdr we don't have master/client relationships, both sides are equal. Maybe we could have sdr_disable signal on both sides and revert to the bsg_link_sdr_tag_lines_s name?

bsg_link/bsg_link_pkg.v
```diff
@@ -0,0 +1,28 @@
+
```
gaozihou, line 1: Maybe we could add a short comment about the usage of the structs? Like to properly de-assert reset signals of bsg_link_sdr, one option is to instantiate bsg_tag_master and bsg_tag_client modules, and this package can tidy up all the tag lines...

bsg_link/bsg_link_pkg.v
dpetrisko: We could, my concern was that the manycore pod did not use the sdr_disable line so I was trying to keep them separate to minimize. I think it makes sense to combine them though. If the disable tag line is not used, the synthesizer should easily optimize away the extra line

bsg_link/bsg_link_pkg.v
dpetrisko: Great idea! I think there's a comment block about the reset sequence in the tag generation modules, I will grab that and modify

bsg_link/bsg_link_pkg.v
gaozihou: I see, yeah I am more concerned about the naming (master client), maybe we could have something like bsg_link_sdr_tag_lines_s and bsg_link_sdr_with_disable_tag_lines_s?

dpetrisko:
Okay adding comment block and renamed!

gaozihou (approved):
LGTM!

dpetrisko:
@taylor-bsg updated comment

PR #536, opened 2021-12-13
Adding bsg_crossbar_control_locking_o_by_i.v

dpetrisko:
This PR adds a locking crossbar, which is useful for crossbars which have multiple beat streams and no tag info on their data payloads

### DIFF
--- a/bsg_misc/bsg_crossbar_control_locking_o_by_i.v
+++ b/bsg_misc/bsg_crossbar_control_locking_o_by_i.v
@@ -61,7 +61,7 @@ module bsg_crossbar_control_locking_o_by_i
   
   for (genvar i = 0 ; i < o_els_p; i++) begin: rr
 
-    logic [i_els_p-1:0] not_req_mask_r, req_mask_r;
+    logic [i_els_p-1:0] not_req_mask_r, req_mask_r, reqs_li;
 
     bsg_dff_reset_en #( .width_p(i_els_p) )
       req_words_reg
@@ -74,7 +74,9 @@ module bsg_crossbar_control_locking_o_by_i
 
     assign req_mask_r = ~not_req_mask_r;
 
-    assign valid_o[i] = |o_select_t[i];
+    assign reqs_li = o_select_t[i] & req_mask_r;
+
+    assign valid_o[i] = |reqs_li;
     assign rr_yumi_li[i] = valid_o[i] & ready_and_i[i];
 
     bsg_arb_round_robin #(
@@ -83,7 +85,7 @@ module bsg_crossbar_control_locking_o_by_i
       .clk_i(clk_i)
       ,.reset_i(reset_i)
     
-      ,.reqs_i(o_select_t[i] & req_mask_r)
+      ,.reqs_i(reqs_li)
       ,.grants_o(grants_oi_one_hot_o[i])
       ,.yumi_i(rr_yumi_li[i])
     );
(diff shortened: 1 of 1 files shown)


taylor-bsg:
@gaozihou Do we have a module (or code inside a module) that already does this kind of thing?

bsg_misc/bsg_crossbar_control_locking_o_by_i.v
```diff
@@ -0,0 +1,113 @@
...
+  (
+    input clk_i
+    , input reset_i
+
+    , input [i_els_p-1:0] valid_i
+    , input [i_els_p-1:0][lg_o_els_lp-1:0] sel_io_i
+    , output [i_els_p-1:0] yumi_o
+
+    // crossbar outputs (ready & valid interface)
+    , input [o_els_p-1:0] unlock_i
```
taylor-bsg, line 27: typically there is a notion in wormhole routers of having a "tail flit", which has a tail bit set. Probably you could rename it. Except... why is unlock_i o_els_p and not i_els_p?

dpetrisko: Yeah the protocol we’re using this for has a “last” bit. 

I think it’s dealer’s choice whether to lock inputs vs outputs since each input can only have 1 simultaneous output and vice versa. I think the implementation is a little simpler on outputs, but not a big deal either way

dpetrisko: https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_misc/bsg_locking_arb_fixed.v

It’s named unlock to be consistent with this

bsg_misc/bsg_crossbar_control_locking_o_by_i.v
```diff
@@ -0,0 +1,115 @@
...
```
gaozihou, line 27: For `unlock_i[o_els_p-1:0]` vs `unlock_i_in[i_els_p-1:0]`, may I ask if this module is mainly used for bsg wormhole protocol? If yes, I guess unlock signal is naturally on the input side, since we usually have a counter on the input side to determine header flit and count down from there.. If we unlock on the output side, then `unlock_i[i] = | (grants_oi_one_hot_o[i] & unlock_i_in)`

If we are targeting other protocols, which have a tail bit, then unlocking on the output side is definitely simpler.

dpetrisko: Good point, yes this is for a protocol with a tail bit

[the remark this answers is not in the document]
dpetrisko: added comment!

gaozihou: Thanks Dan! Makes sense.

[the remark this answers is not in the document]
gaozihou: Thanks!

gaozihou (approved):
LGTM! I guess this module is an alternative to the existing [bsg_wormhole_router_output_control.v](https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_noc/bsg_wormhole_router_output_control.v). This new module is targeting protocols that have a tail bit, and requires hold-on-valid.

PR #539, opened 2022-01-07
Adding labels to genblks in bsg_cache

dpetrisko:
Verilator seems to have a problem if you instantiate multiple bsg_caches in for loop, due to inlining. I'm working on raising a bug report, but the faster workaround is to add labels for all genblks in bsg_cache.  Besides, it's part of the style guidelines so no harm no foul.

### DIFF
--- a/bsg_cache/bsg_cache.v
+++ b/bsg_cache/bsg_cache.v
@@ -120,7 +120,7 @@ module bsg_cache
   if (burst_len_lp == 1) begin
     assign ld_data_mem_addr = addr_index;
   end
-  else if (burst_len_lp == block_size_in_words_p) begin
+  else if (burst_len_lp == block_size_in_words_p) begin : fi1 // Required for Verilator 4.213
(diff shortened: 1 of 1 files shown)


taylor-bsg:
It's not great. Usually the missing labels are desireable because they provide handles for the backend flow to work on. But since this is RTL and not structural verilog, no such benefit is conferred, instead we have to check to make sure none of the labels repeat.

I would say, you can commit it, but maybe uncommit when verilator is fixed..

M

dpetrisko:
Works for me, I don't like them either. good to merge then?

dpetrisko:
Minimized the changes

dpetrisko:
Fixed in 4.217, closing

dpetrisko:
>Usually the missing labels are desireable because they provide handles for the backend flow to work on

For a note though, both Verilator and VCS appear to flag unlabelled genblks as lint warnings, regardless if there is structure or assigns inside

dpetrisko:
I spoke too soon on this. The latest Verilator is incompatible with multicore BlackParrot, so we are prevented from upgrading until that is resolved. Please merge this in the meantime

PR #380, opened 2021-01-11
Sim dmc tag

hema0730:
Updating bsg_dmc test infrastructure with simulation for configuration process by bsg_tag.

PR #540, opened 2022-01-09
DMC initial programming with BSG Tag

akashnt4:
Rebased and updated PR #380 .

### DIFF
--- a/bsg_clk_gen/bsg_dly_line.v
+++ b/bsg_clk_gen/bsg_dly_line.v
@@ -30,29 +30,32 @@ module bsg_dly_line
    bsg_clk_gen_osc_tag_payload_s fb_tag_r;
    wire  fb_we_r;
 
-   // note: oscillator has to be already working in order
+   // note: delay line has to be already working in order
    // for configuration state to pass through here
 
-   bsg_tag_client #(.width_p($bits(bsg_clk_gen_osc_tag_payload_s))
-                    ,.harden_p(1)
-                    ) btc
-     (.bsg_tag_i     (bsg_tag_i)
-      ,.recv_clk_i   (clk_o)
-      ,.recv_new_r_o (fb_we_r)  // default is already in OSC flops
-      ,.recv_data_r_o(fb_tag_r)
-      );
+   bsg_tag_client_unsync
+     #(.width_p($bits(bsg_clk_gen_osc_tag_payload_s))
+       ,.harden_p(1)
+       ) btc
+       (.bsg_tag_i(bsg_tag_i)
+        ,.data_async_r_o(fb_tag_r)
+        );
+
+   bsg_tag_client_unsync
+     #(.width_p(1)
+       ,.harden_p(1)
+       ) btc_trigger
+       (.bsg_tag_i(bsg_tag_trigger_i)
+        ,.data_async_r_o(fb_we_r)
+        );
 
    wire [1:0] cdt = fb_tag_r.cdt;
    wire [1:0] fdt = fb_tag_r.fdt;
    wire [num_adgs_p-1:0] adg_ctrl = fb_tag_r.adg;
 
    logic [4+num_adgs_p-1:0] ctrl_rrr;
-   always @(clk_o or async_reset_i)
-     if (async_reset_i)
-       ctrl_rrr <= '0;
-     else
-       if (fb_we_r)
-         ctrl_rrr <= {adg_ctrl, cdt, fdt};
+   always @(posedge fb_we_r)
+     ctrl_rrr <= {adg_ctrl, cdt, fdt};
 
    always
      begin
--- a/testing/bsg_dmc/testbench.v
+++ b/testing/bsg_dmc/testbench.v
@@ -90,6 +90,9 @@ module testbench();
   wire  [(dq_data_width_p>>3)-1:0] ddr_dqs_n;
   wire       [dq_data_width_p-1:0] ddr_dq;
 
+  // All tag lines from the btm
+  bsg_tag_s [22:0] tag_lines_lo;
+
   traffic_generator #
     (.num_adgs_p         ( clk_gen_num_adgs_p  )
     ,.ui_addr_width_p    ( ui_addr_width_p     )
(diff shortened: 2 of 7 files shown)


testing/bsg_dmc/traffic_generator.v
```diff
@@ -0,0 +1,323 @@
...
+        btc
+          (.bsg_tag_i     ( dmc_cfg_tag_lines_lo[idx] )
+          ,.recv_clk_i    ( dfi_clk_1x_i )
+          ,.recv_new_r_o  ( dmc_cfg_tag_new_data_lo[idx] )
+          ,.recv_data_r_o ( dmc_cfg_tag_data_lo[idx] )
+          );
+    end
+  endgenerate
+
+  assign dmc_p_o.trefi        = {dmc_cfg_tag_data_lo[1], dmc_cfg_tag_data_lo[0]};
```
dpetrisko, line 190: I would put this tag client in the testbench next to the dmc. After all, it'll be on the chip in the long run

testing/bsg_dmc/traffic_generator.v
```diff
@@ -0,0 +1,323 @@
...
+`endif
+
+`ifndef DFI_CLK_PERIOD
+  `define DFI_CLK_PERIOD 5000.0
+`endif
+
+`ifndef TAG_CLK_PERIOD
+  `define TAG_CLK_PERIOD 10000.0
+`endif
+
```
dpetrisko, line 15: `include "bsg_defines.v"

testing/bsg_dmc/traffic_generator.v
```diff
@@ -0,0 +1,323 @@
...
+module traffic_generator
+  import bsg_tag_pkg::*;
+  import bsg_dmc_pkg::*;
+ #(parameter  num_adgs_p         = 1
+  ,parameter  ui_addr_width_p    = "inv"
```
dpetrisko, line 20: `BSG_INV_PARAM(ui_addr_width_p)

dpetrisko: etc

testing/bsg_dmc/testbench.v
```diff
@@ -25,10 +20,10 @@ module testbench
 
   integer j,k;
 
-  wire bsg_tag_s        dmc_reset_tag_lines_lo;
-  wire bsg_tag_s  [3:0] dmc_dly_tag_lines_lo;
-  wire bsg_tag_s  [3:0] dmc_dly_trigger_tag_lines_lo;
-  wire bsg_tag_s        dmc_ds_tag_lines_lo;
+  wire bsg_tag_s        dmc_reset_tag_lines;
```
dpetrisko, line 23: _lo is the code convention (local output)

bsg_clk_gen/bsg_dly_line.v
```diff
@@ -30,29 +30,32 @@ module bsg_dly_line
    bsg_clk_gen_osc_tag_payload_s fb_tag_r;
    wire  fb_we_r;
 
-   // note: oscillator has to be already working in order
+   // note: delay line has to be already working in order
```
dpetrisko, line 33: Were you seeing issues with this module?  This is the "soft" version of this module, so it's possible only the hard version has been rigorously tested. Still, seems to be safer with the async reset

testing/bsg_dmc/traffic_generator.v
akashnt4: Done

testing/bsg_dmc/traffic_generator.v
akashnt4: done

bsg_clk_gen/bsg_dly_line.v
akashnt4: No, I didn't see any issues... any particular syndrome to look for?

dpetrisko: Ah, if there weren’t issues then let’s not modify this file

akashnt4: Oh, sorry I had misunderstood your first question. Without this change, the read data is corrupted. I'm checking on the reason behind it.

dpetrisko: Discussed offline, these are basically backports of: https://github.com/bespoke-silicon-group/basejump_stl/blob/master/hard/tsmc_40/bsg_clk_gen/bsg_dly_line.v

dpetrisko (approved):

bsg_clk_gen/bsg_dly_line.v
```diff
@@ -30,29 +30,32 @@ module bsg_dly_line
...
 
    logic [4+num_adgs_p-1:0] ctrl_rrr;
-   always @(clk_o or async_reset_i)
-     if (async_reset_i)
-       ctrl_rrr <= '0;
-     else
-       if (fb_we_r)
-         ctrl_rrr <= {adg_ctrl, cdt, fdt};
+   always @(posedge fb_we_r)
+     ctrl_rrr <= {adg_ctrl, cdt, fdt};
```
gaozihou, line 58: I guess the new logic does not have the same behavior as the old one?

Seems like the hardened module does utilize async_reset_i as asynchronous DFF reset (inside adt, cdt and fdt modules): https://github.com/bespoke-silicon-group/basejump_stl/tree/master/hard/tsmc_40/bsg_clk_gen

dpetrisko: https://github.com/bespoke-silicon-group/basejump_stl/commit/a611e9efa48685c67ee01751d632b8aa1196298b

Looks like this was the commit where these came in. So I guess the question is which version is the legit one. Probably the newer one?

gaozihou: Thanks for the explanation Dan! Looks like new version has Chun's latest update.

My understanding is that the RTL model should match the behavior of the hardened one. In this case, when async_reset_i is cycled 0->1->0, hardened version clears the adt/cdt/fdt registers and falls back to the default delay, while new version RTL model keeps the previously configured delay value.

gaozihou: I guess it is possible that Chun's latest version RTL (in this PR) is legit, and the hardened dly line is out-of-date...

The [hardened dly line](https://github.com/akashnt4/basejump_stl/blob/dmc_tag/hard/tsmc_40/bsg_clk_gen/bsg_dly_line.v) wouldn't get out of reset & get programmed without clk_i toggling, while the DDR DQS input clocks are usually at high-impedance Z..

Maybe we should try to find the latest version hardened dly line?

taylor-bsg:
Hi Paul,

I believe the flops in the TSMC 40 clock gen are async reset flops (you can
verify); maybe double check the assertion about the clk_i being required to
toggle?

M

gaozihou:
Thanks @taylor-bsg !

The flops being used are DFCND and DFNCND in TSMC40. Both flops have "async reset", which can reset the flops with clk_i==1'bX. So yeah I was half wrong, the flops can come out of reset without clk_i toggling.

I think we still have the programmability problem. All the flops are synchronous to clk_i; we_async_i (trigger) signal is attached to 2-stage sync flops. This means clk_i needs to toggle at lease 3 times in order to get the dly line programmed, and this must happen exactly when trigger_r == 1'b1 (programmed by bsg_tag). I guess it is hard for DQS clock to toggle when we are configuring the DMC...

Here is a sketch for proposed fix to the hardened dly line: https://github.com/bespoke-silicon-group/basejump_stl/commit/fa72d107c6a56dc2c577a65e550f0687a884a13a

PR #542, opened 2022-01-11
1r1w enhancements

dpetrisko:
- Banked 1r1w
- 1r1w latch last read behavior
- 1r1w byte mask

### DIFF
--- a/bsg_mem/bsg_mem_1r1w_sync_banked.v
+++ b/bsg_mem/bsg_mem_1r1w_sync_banked.v
@@ -115,7 +115,7 @@ module bsg_mem_1r1w_sync_banked
           .clk_i(clk_i)
           ,.reset_i(reset_i)
 
-          ,.w_v_i(bank_w_v_li)
+          ,.w_v_i(bank_w_v_li[j])
           ,.w_addr_i(bank_w_addr_li)
           ,.w_data_i(w_data_i[i*bank_width_lp+:bank_width_lp])
           ,.r_v_i(bank_r_v_li[j])
--- a/bsg_mem/bsg_mem_1r1w_sync_banked.v
+++ b/bsg_mem/bsg_mem_1r1w_sync_banked.v
@@ -10,6 +10,11 @@
  *  It is also useful for power and delay perspective, since only one depth
  *  bank is activated while reading or writing.
  *
+ * if the desired banking is decided by node-specific (e.g. TSMC 28) SRAM 
+ * generator characteristics then the preferred methodology is to invoke this
+ * from the /hard SRAM macro, rather than to instantiate it in the RTL.
+ * However, in some cases, for example, for research infrastructures exploring
+ * a pareto design space, it may be desirable to directly use this interface.
  *
  *  - width_p : width of the total memory
  *  - els_p : depth of the total memory
--- a/bsg_mem/bsg_mem_1r1w_sync_banked.v
+++ b/bsg_mem/bsg_mem_1r1w_sync_banked.v
@@ -28,6 +28,7 @@ module bsg_mem_1r1w_sync_banked
   #(parameter `BSG_INV_PARAM(width_p)
     , parameter `BSG_INV_PARAM(els_p)
     , parameter read_write_same_addr_p=0
+    , parameter latch_last_read_p=0
 
     , parameter num_width_bank_p=1
     , parameter num_depth_bank_p=1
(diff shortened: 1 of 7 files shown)


bsg_mem/bsg_mem_1r1w_sync_banked.v
```diff
@@ -0,0 +1,168 @@
...
+    ) demux_r_v (
+      .i(depth_bank_r_idx_li)
+      ,.v_i(r_v_i)
+      ,.o(bank_r_v_li)
+    );
+
+    bsg_decode_with_v #(
+      .num_out_p(num_depth_bank_p)
+    ) demux_w_v (
+      .i(w_depth_bank_w_idx_li)
```
tommydcjung, line 101: Is "w_" a typo?

bsg_mem/bsg_mem_1r1w_sync_banked.v
```diff
@@ -0,0 +1,168 @@
...
+      ,.v_i(w_v_i)
+      ,.o(bank_w_v_li)
```
tommydcjung, line 103: Is this going to be used somewhere?

bsg_mem/bsg_mem_1r1w_sync_mask_write_byte_synth.v
```diff
@@ -0,0 +1,69 @@
...
+//
+// reads are synchronous
+//
+// NOTE: Users of BaseJump STL should not instantiate this module directly
+// they should use bsg_mem_1r1w_sync_mask_write_bit.
+
+
+`include "bsg_defines.v"
+
+module bsg_mem_1r1w_sync_mask_write_bit_synth #(parameter `BSG_INV_PARAM(width_p)
```
tommydcjung, line 13: Is this supposed to be bit or byte?

bsg_mem/bsg_mem_1r1w_sync_mask_write_byte_synth.v
```diff
@@ -0,0 +1,69 @@
...
+
+   if (width_p == 0)
+    begin: z
+      wire unused0 = &{clk_i, w_v_i, w_mask_i, w_addr_i, r_v_i, r_addr_i};
+      assign r_data_o = '0;
+    end
+   else
+    begin: nz
+
+  for(i=0; i<write_mask_width_lp; i=i+1)
```
tommydcjung, line 48: missing genvar.

bsg_mem/bsg_mem_1r1w_sync_banked.v
dpetrisko: yep used below

dpetrisko:
feedback addressed!

dpetrisko:
@tommydcjung additional comments or can we merge?

tommydcjung:
Is it tested and ready to go?

bsg_mem/bsg_mem_1r1w_sync_banked.v
```diff
@@ -0,0 +1,168 @@
...
+ *  This module has the same interface/functionality as
+ *  bsg_mem_1r1w_sync.
+ *
+ *  This module can be used for breaking a big SRAM block into
+ *  smaller blocks. This might be useful, if the SRAM generator does not
+ *  support sizes of SRAM that are too wide or too deep.
+ *  It is also useful for power and delay perspective, since only one depth
+ *  bank is activated while reading or writing.
+ *
+ *
```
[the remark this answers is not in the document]
dpetrisko: Agree with power, I'll add it to the other banked implementation as well. 

Delay I would be surprised. The external mux and additional fanout seem like it would dominate over extending the SRAM internal architecture. 

Otherwise, the degenerate case would imply to always use the smallest SRAM and bank it?

taylor-bsg:
If the SRAMs get big enough, then the internal delay of the SRAM array can
start to dominate everything -- the bitlines are slowing down because of RC
delay.
Probably the least likely thing is that you would save energy by splitting
the ram. In terms of power, splitting width is probably not a power saver.

M

dpetrisko:
Agree if you're just talking splitting when a larger is available. However, you can clock/power gate half of a split RAM but not a full width ram. So for some access patterns e.g. sequential it certainly could lower energy

dpetrisko:
Updated comment, good to go now

dpetrisko:
Note to remember to add: https://github.com/bespoke-silicon-group/basejump_stl/pull/572 as ultrascale plus variant

tommydcjung (approved):

dpetrisko:
thanks!

PR #547, opened 2022-02-09
Tsmc28 sync libraries

dpetrisko:


taylor-bsg:
I would suggest the gf_14 as the starting place; and to indicate that prominently in the comments.  For example, it has the BSG_NO_CLK_GATE attribute.

What happens to the rp group macros? Are they just ignore?

dpetrisko:
>What happens to the rp group macros? Are they just ignore?

Yeah they're just ignored

taylor-bsg:
Then I would say, maybe remove all that stuff and just a generate statement
that instantiates the three bit registers... But check out the GF14 for
naming

dpetrisko:
@taylor-bsg Updated to have a single hardened BSS instance with the hardened modules instantiating that instance

hard/tsmc_28/bsg_async/bsg_sync_sync_unit.v
```diff
@@ -0,0 +1,25 @@
+
```
stdavids, line 1: So I am assuming that you had no SYNC cells so you have this 'unit' module to build one from normal DFFs. If this is the case, then what are the timing/physical constraints for these two cells? I assume that they should be right next to each other with a very small max delay on the path between them? These would be good to put in comments :)

hard/tsmc_28/bsg_async/bsg_sync_sync_async_reset_unit.v
```diff
@@ -0,0 +1,28 @@
+
```
stdavids, line 1: See comments for 'bsg_sync_sync_unit.v'

stdavids (commented):
sync_sync and launch_sync_sync look good, but some header comments on the *unit.v files explaining the physical/timing constraints for composing a SYNC2 cells would be good for posterity.

hard/tsmc_28/bsg_async/bsg_sync_sync_unit.v
dpetrisko: Added comments!

stdavids (approved):

PR #548, opened 2022-02-14
delay cell for hardening

tommydcjung:
Delay cell for hardening


taylor-bsg:
what is it for? How much delay should it have? Is there a equivalent hardened cell?

tommydcjung:
There are delay cells in the pdk. This can be used to manually place delay cells to fix hold time. I don't think it needs any delay here.

PR #550, opened 2022-02-17
[DEFERRED] Clock Downsampler with != 50% duty cycle

dpetrisko:
This PR introduces bsg_counter_clock_downsample_unbalanced.v, which can be used for generating a clock with an arbitrary duty cycle. The specific use-case being targeted is generating a 1/3 (67% 33%) clock waveform


taylor-bsg:
I think would just have a register to hold the two init values and then switch between the values on the strobe?

taylor-bsg:
Then you would have one strobe unit.

dpetrisko:
Superceded by #700  (which will decompose into a smaller PR)

PR #551, opened 2022-02-17
Adding misc tsmc28 hard

dpetrisko:
This PR adds miscellaneous TSMC28 hardened modules, specifically those used in bsg_strobe

### DIFF
--- a/hard/tsmc_28/bsg_misc/bsg_buf.v
+++ b/hard/tsmc_28/bsg_misc/bsg_buf.v
@@ -1,0 +1,114 @@
+`define bsg_buf_macro(bits)                                   \
+if (harden_p && (width_p==bits))                              \
+  begin: macro                                                \
+    for (genvar j = 0; j < width_p; j++)                      \
+      begin : b                                               \
+        BUFFD8BWP7T40P b (.I(i[j]), .Z(o[j]));                \
+      end                                                     \
+  end
+
+module bsg_buf #(parameter `BSG_INV_PARAM(width_p)
+                 , parameter harden_p=1
+		 , parameter vertical_p=1
+                 )
+   (input    [width_p-1:0] i
+    , output [width_p-1:0] o
+    );
+
+   `bsg_buf_macro(89) else
+   `bsg_buf_macro(88) else
+   `bsg_buf_macro(87) else
+   `bsg_buf_macro(86) else
+   `bsg_buf_macro(85) else
+   `bsg_buf_macro(84) else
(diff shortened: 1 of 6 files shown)


hard/tsmc_28/bsg_misc/bsg_dff.v
```diff
@@ -0,0 +1,193 @@
...
+`define bsg_dff_macro(bits,strength)                      \
+if (harden_p && (width_p==bits) && (strength_p==strength)) \
+  begin: macro                                                \
+    for (genvar j = 0; j < width_p; j++)                      \
+      begin : d                                               \
+        DFQD``strength``BWP7T40P140 d (.CP(clk_i), .D(data_i[j]), .Q(data_o[j])); \
+      end                                                     \
+  end
+
+module bsg_dff #(width_p=-1, harden_p=1, strength_p=1)
```
tommydcjung, line 13: BSG_INV_PARAM

hard/tsmc_28/bsg_misc/bsg_dff.v
```diff
@@ -0,0 +1,193 @@
...
+    else
+   `bsg_dff_macro(32,8)
+     else
+     begin: notmacro
+        reg [width_p-1:0] data_r;
+
+        assign data_o = data_r;
+
+        always @(posedge clk_i)
+          data_r <= data_i;
```
tommydcjung, line 189: It's kinda weird that some of these have checks on the input params and other don't..

dpetrisko: Yeah good point, I'll clean that up. FWIW, they're copied from tsmc40, so the issues exist there as well

dpetrisko:
@tommydcjung fixed the inv params!

hard/tsmc_28/bsg_misc/bsg_buf.v
```diff
@@ -0,0 +1,117 @@
...
+    , output [width_p-1:0] o
+    );
+
+   `bsg_buf_macro(89) else
+   `bsg_buf_macro(88) else
+   `bsg_buf_macro(87) else
+   `bsg_buf_macro(86) else
+   `bsg_buf_macro(85) else
+   `bsg_buf_macro(84) else
+   `bsg_buf_macro(83) else
```
tommydcjung, line 27: Seems little arbitrary why the width has to be less than 90. (or why does it have to go all the way up to 90) For other modules too. Might wanna think about that for later...

tommydcjung (approved):

PR #552, opened 2022-02-22
expose hold_on_valid_p param in wh router and concentrators

muwyse:
This PR simply exposes the hold_on_valid_p parameter from bsg_round_robin_arb.v in bsg_wormhole_router_output_control, bsg_wormhole_router, and bsg_wormhole_concentrator*.

This parameter is useful in situations where an output channel may not be consumed in a single cycle, for example when a flit must be processed by a PISO.

### DIFF
--- a/bsg_noc/bsg_wormhole_concentrator.v
+++ b/bsg_noc/bsg_wormhole_concentrator.v
@@ -29,6 +29,12 @@ module bsg_wormhole_concentrator
     ,parameter num_in_p            = 1
     ,parameter debug_lp            = 0
     ,parameter link_width_lp       = `bsg_ready_and_link_sif_width(flit_width_p)
+    // Hold on valid sets the arbitration policy such that once an output tag is selected, it
+    // remains selected until it is acked, then the round-robin scheduler continues cycling
+    // from the selected tag. This is consistent with BaseJump STL handshake assumptions.
+    // Notably, this parameter is required to work with bsg_parallel_in_serial_out_passthrough.
+    // This policy has a slight throughput degradation but effectively arbitrates based on age,
+    // so minimizes worst case latency.
     ,parameter hold_on_valid_p     = 0
    )
 
--- a/bsg_noc/bsg_wormhole_concentrator.v
+++ b/bsg_noc/bsg_wormhole_concentrator.v
@@ -26,9 +26,10 @@ module bsg_wormhole_concentrator
     ,parameter `BSG_INV_PARAM(len_width_p)
     ,parameter `BSG_INV_PARAM(cid_width_p)
     ,parameter `BSG_INV_PARAM(cord_width_p)
-   ,parameter num_in_p            = 1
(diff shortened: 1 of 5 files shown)


dpetrisko:
@taylor-bsg notably the passthrough PISO

bsg_noc/bsg_wormhole_router_output_control.v
```diff
@@ -1,7 +1,16 @@
 `include "bsg_defines.v"
 
 module bsg_wormhole_router_output_control
-  #(parameter `BSG_INV_PARAM(input_dirs_p))
+  #(parameter `BSG_INV_PARAM(input_dirs_p)
+    // Hold on valid sets the arbitration policy such that once
+    // a output tag is selected, it remains selected until it is
+    // acked. This is consistent with BaseJump STL handshake
```
gaozihou, line 7: Maybe add ", then the round-robin scheduler continues cycling from the selected tag."

bsg_noc/bsg_wormhole_router.v
```diff
@@ -18,6 +18,9 @@ module bsg_wormhole_router
    ,parameter reverse_order_p       = 0
    ,parameter `BSG_INV_PARAM(len_width_p)
    ,parameter debug_lp              = 0
+   // hold router_output_control input selection until consumed
+   // See bsg_wormhole_router_output_control and bsg_round_robin_arb.v for details
```
gaozihou, line 22: Could we also add detailed explanations here? It might be convenient for the wh router users to decide whether to enable this feature.

bsg_noc/bsg_wormhole_concentrator.v
```diff
@@ -26,9 +26,10 @@ module bsg_wormhole_concentrator
     ,parameter `BSG_INV_PARAM(len_width_p)
     ,parameter `BSG_INV_PARAM(cid_width_p)
     ,parameter `BSG_INV_PARAM(cord_width_p)
-   ,parameter num_in_p            = 1
-   ,parameter debug_lp            = 0
-   ,parameter link_width_lp       = `bsg_ready_and_link_sif_width(flit_width_p)
+    ,parameter num_in_p            = 1
+    ,parameter debug_lp            = 0
+    ,parameter link_width_lp       = `bsg_ready_and_link_sif_width(flit_width_p)
+    ,parameter hold_on_valid_p     = 0
```
gaozihou, line 32: Maybe also add detailed explanation?

gaozihou (commented):
Passed regression, the new behaviors are as expected (it holds the selection till granted).

Would be great if we could add comments on the top-level module (bsg_wormhole_router and bsg_wormhole_concentrator) to add convenience to the users.

muwyse:
@gaozihou expanded the comment and added to requested modules, thanks!

gaozihou (approved):

gaozihou:
Both wormhole_router and wormhole_concentrator works. Corner case (pow2 inputs) also works.

PR #553, opened 2022-02-23
Updates to clock gen osc and delay line

dpetrisko:
This PR adds the capability to tune our nonsynth oscillators (clock gen and delay line), for different process nodes using BSG_OSC_BASE_DELAY and BSG_OSC_GRANULARITY. By setting these parameters, we can get a fairly good approximation of the delays after APR. This is especially useful / necessary for the DMC which relies on relative delays between the oscillators and interface.

It also adds the BSG_NO_TIMESCALE macro which removes timescale from all modules. This is useful for simulating with libraries which have a defined timescale, since there are very messy semantics for overriding timescale in SystemVerilog.

### DIFF
--- a/bsg_clk_gen/bsg_dly_line.v
+++ b/bsg_clk_gen/bsg_dly_line.v
@@ -44,32 +44,29 @@ module bsg_dly_line
    bsg_clk_gen_osc_tag_payload_s fb_tag_r;
    wire  fb_we_r;
 
-   // note: delay line has to be already working in order
+   // note: oscillator has to be already working in order
    // for configuration state to pass through here
 
-   bsg_tag_client_unsync
-     #(.width_p($bits(bsg_clk_gen_osc_tag_payload_s))
-       ,.harden_p(1)
-       ) btc
-       (.bsg_tag_i(bsg_tag_i)
-        ,.data_async_r_o(fb_tag_r)
-        );
-
-   bsg_tag_client_unsync
-     #(.width_p(1)
-       ,.harden_p(1)
-       ) btc_trigger
-       (.bsg_tag_i(bsg_tag_trigger_i)
-        ,.data_async_r_o(fb_we_r)
-        );
+   bsg_tag_client #(.width_p($bits(bsg_clk_gen_osc_tag_payload_s))
+                    ,.harden_p(1)
+                    ) btc
+     (.bsg_tag_i     (bsg_tag_i)
+      ,.recv_clk_i   (clk_o)
+      ,.recv_new_r_o (fb_we_r)  // default is already in OSC flops
+      ,.recv_data_r_o(fb_tag_r)
+      );
 
    wire [1:0] cdt = fb_tag_r.cdt;
    wire [1:0] fdt = fb_tag_r.fdt;
    wire [num_adgs_p-1:0] adg_ctrl = fb_tag_r.adg;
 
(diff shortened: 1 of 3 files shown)


gaozihou:
Thanks @dpetrisko! I was aware of this problem, it prevents us from running RTL simulation on clk_gen (we were running RTL_hard though)

Would it be better or worse to set them as parameters (currently they are localparams + defines)? Which one is more suitable to our flow?

dpetrisko:
Originally I had them as params, but the problem is that it’s not cleanly compatible with the hardened flow (need to wire up the parameters in the source RTL, but then those parameters are ignored in the hardened RTL). And it’s also a “nonsynth parameter”, which is its own weirdness…

gaozihou:
I see! Another option is to add the following to the RTL simulation Makefile, then we don't have to wire up the parameters in the source RTL:
```
VCS_OPTIONS += -pvalue+bsg_osc_base_delay_p=<num>
VCS_OPTIONS += -pvalue+bsg_osc_granularity_p=<num>
```

Btw, current method (define + localparams) passes the clk_gen_pearl test. I don't have a testbench for bsg_dly_line..

dpetrisko:
```
VCS_OPTIONS += -pvalue+bsg_osc_base_delay_p=<num>
VCS_OPTIONS += -pvalue+bsg_osc_granularity_p=<num>
```

Oh that’s cool, didn’t know you could do that with VCS. At least in old versions of Verilator,  the parameters could only be passed to the toplevel from command line, so you had to thread them all the way through. That feature may have been added at some point though

gaozihou:
Oh yeah I forgot the Verilator... Then maybe define+localparam is the only option

bsg_clk_gen/bsg_dly_line.v
```diff
@@ -25,41 +27,56 @@ module bsg_dly_line
...
 
    logic [4+num_adgs_p-1:0] ctrl_rrr;
-   always @(clk_o or async_reset_i)
-     if (async_reset_i)
-       ctrl_rrr <= '0;
-     else
-       if (fb_we_r)
-         ctrl_rrr <= {adg_ctrl, cdt, fdt};
+   always @(posedge fb_we_r)
+     ctrl_rrr <= {adg_ctrl, cdt, fdt};
```
gaozihou, line 72: I guess the new logic does not have the same behavior as the old one?

Seems like the hardened module does utilize async_reset_i as asynchronous DFF reset (inside adt, cdt and fdt modules): https://github.com/bespoke-silicon-group/basejump_stl/tree/master/hard/tsmc_40/bsg_clk_gen

dpetrisko: Oh you know what it is, I was working off this PR: https://github.com/bespoke-silicon-group/basejump_stl/pull/540/files

Let me revert those changes so this PR is just the ‘defines

dpetrisko: Should be good to go now

drichmond:
I think the same change needs to be made in `bsg_nonsynth_clock_gen`

I can push it, but it's dan's PR.

dpetrisko:
@drichmond I’m fine if you add on to this, what’s here should be final

drichmond:
done!

gaozihou (approved):
LGTM

PR #557, opened 2022-03-01
Make passthrough SIPO truly passthrough

dpetrisko:
The passthrough SIPO right now has weird behavior. It builds up incrementally and then comes through all at once. This was done to save hardware but breaks certain usecases which want to examine the intermediate results.

For instance, passing in ABCD would have C and D both available starting at cycle 4.

Instead, if we bypass each entry, the SIPO becomes truly passthrough. Then the output sequence will be ABCD as expected.

### DIFF
--- a/bsg_dataflow/bsg_serial_in_parallel_out_passthrough.v
+++ b/bsg_dataflow/bsg_serial_in_parallel_out_passthrough.v
@@ -57,7 +57,7 @@ module bsg_serial_in_parallel_out_passthrough
   for (genvar i = 0; i < els_p-1; i++)
     begin: rof
       wire my_turn = v_i & count_r[i];
-      bsg_dff_en #(.width_p(width_p)) dff
+      bsg_dff_en_bypass #(.width_p(width_p)) dff
       (.clk_i
        ,.data_i
        ,.en_i   (my_turn)


gaozihou:
Thanks Dan!

I guess you are expecting to use the passthrough PISO in a special use case (utilizing the intermediate results). With only one v_o signal there is no way to tell whether A,B or C is valid or not. In my opinion you'll have to modify the interface with
```
, output logic [els_p-1:0]           v_o
```
The new v_o signals can be driven by the ```my_turn``` signal.

dpetrisko:
Yep, makes sense! And then the normal usecase of only detecting full can be achieved by looking at the high bit of the valids

taylor-bsg:
Is this an N buncher and not a SIPO?

https://github.com/bespoke-silicon-group/basejump_stl/blob/66dc2bab55e888efe5b2626c0c97ea5bae2b04c8/bsg_dataflow/bsg_two_buncher.v#L4

dpetrisko:
Hm, interesting. I guess yes and no, but I'll have to look at the handshaking to understand if there are any fundamental differences

dpetrisko:
The difference is in the output handshake.

In bsg_n_buncher, there are N ready signals so the packets can be popped individually (though not out of sequence).

In bsg_serial_in_parallel_out_passthrough, there is 1 ready signal and the entire packet is acked when received.

I believe bsg_serial_in_parallel_out_passthrough is a specialization of bsg_n_buncher, if we adjust the ready_and_i signal externally

PR #559, opened 2022-03-09
Clk gen v3

dpetrisko:
Putting here as WIP, need to validate with spice + gate level sims. This is a more scalable clock generator with composable blocks to create balanced delays. Also the corresponding delay line. Haven't added RP groups since I'll be focusing on TSMC28 SDP groups. Sizing the cells must be done with more care, as well


dpetrisko:
Partially superseded by #571, GF14 porting will need to be redone and validated

PR #560, opened 2022-03-25
Adding balanced_p parameter to bsg_reduce

dpetrisko:
This is useful for constructing trees of balanced clock ors/ands


taylor-bsg:
@dpetrisko the implementation is intended to allow the synthesizer to auto-balance. but why add the parameter, if it is unused?

dpetrisko:
Used in the hard/ version, otherwise you can get a parameter error when swapping

taylor-bsg:
Maybe add a comment to that effect

dpetrisko:
@taylor-bsg good to go?

stdavids (approved):
Seems good to me, comment added addresses the main concern.

PR #561, opened 2022-03-26
Combining latch implementations

dpetrisko:
We have two identical implementations of latches, this PR combines them. (And removes the noisy warning since we're actually using latches in major projects at this point)


bsg_misc/bsg_latch.v
```diff
@@ -3,14 +3,17 @@
  *
  */
 
+`include "bsg_defines.v"
+
 module bsg_latch
+ #(parameter `BSG_INV_PARAM(width_p)
```
tommydcjung, line 9: Can we set the default to 1?

dpetrisko: Up to @taylor-bsg , following the pattern of bsg_dff not having a default

taylor-bsg:
ported the substance of this, but use bsg_dlatch as the winning name, to mirror bsg_dff  (there are multiple types of latches under the sun), also fixed depending multiported latch register file.

PR #563, opened 2022-04-01
Unharden synth memories

dpetrisko:
Some of our synth memories have parameters for harden_p, some do not. It doesn't really make sense to have in an explicitly synth block, so this PR removes the parameter. None of our code instantiates synth memories directly so no additional updates should be needed. It also updates bsg_mem_generator to reflect this


dpetrisko:
Needed for https://github.com/bespoke-silicon-group/bsg_14/pull/133 if your design has synth memories

tommydcjung:
Did you check that the wrapper modules in basejump don't use harden_p?

dpetrisko:
Good call, a few did. I also added it to bsg_mem_1rw_banked for consistency

tommydcjung (approved):

PR #564, opened 2022-04-06
Adding ultraram inference

dpetrisko:
This PR adds URAM inference for ultrascale parts. There's substantially more URAM in many parts which is poorly utilized because it won't often be inferred without specific coding styles


dpetrisko:
Superceded by #572

PR #572, opened 2022-06-25
[WIP] Ultrascale plus byte mask block ram

gaozihou:
Original code causes inefficient block ram mapping with Xilinx Ultrascale FPGAs, since every byte is mapped to a separate RAMB18E2 (50% utilization in the best case).

This PR hardens the byte masking block ram for Ultrascale Plus FPGAs. With this change, byte make BRAM utilization will reduce at least by half.

Reference: Vivado Design Suite User Guide: Synthesis (UG901), Byte Write Enable (Block RAM), https://docs.xilinx.com/v/u/2019.1-English/ug901-vivado-synthesis

### DIFF
--- a/hard/ultrascale_plus/bsg_mem/bsg_mem_1r1w_sync_mask_write_byte.v
+++ b/hard/ultrascale_plus/bsg_mem/bsg_mem_1r1w_sync_mask_write_byte.v
@@ -12,6 +12,29 @@
 * https://docs.xilinx.com/v/u/2019.1-English/ug901-vivado-synthesis
 *
 * reads are synchronous
+*
+*
+* By default, the tool selects which type of RAM to infer based upon heuristics
+* that give the best results for most designs. Possible RAM mappings are:
+*
+* ┌───────────┬──────────┬──────┬─────┬─────┬─────┬──────────────────────────────┐
+* │ RAM Type  │Primitives│ Size │PortA│PortB│wb_en│           Mapping            │
+* ├───────────┼──────────┼──────┼─────┼─────┼─────┼──────────────────────────────┤
+* │Distributed│  LUTRAM  │ N/A  │ N/A │ N/A │ N/A │Memory with small depth       │
+* ├───────────┼──────────┼──────┼─────┼─────┼─────┼──────────────────────────────┤
+* │           │ RAMB18E2 │ 18Kb │  W  │  R  │  Y  │Width narrower than 2 bytes   │
+* │ Block RAM ├──────────┼──────┼─────┼─────┼─────┼──────────────────────────────┤
+* │           │ RAMB36E2 │ 36Kb │  W  │  R  │  Y  │Width narrower than 4 bytes   │
+* ├───────────┼──────────┼──────┼─────┼─────┼─────┼──────────────────────────────┤
+* │ Ultra RAM │ URAM288  │288Kb │  W  │  R  │  Y  │Width wider than 4 bytes      │
+* └───────────┴──────────┴──────┴─────┴─────┴─────┴──────────────────────────────┘
+*
+* To force the RAM into a specific type, use the RAM_STYLE attribute to tell
+* Vivado synthesis to infer the target primitives:
+*
+* (* ram_style = "x" *) logic [data_size-1:0] mem [2**addr_size-1:0];
+* Where x = [block, distributed, registers, ultra]
+*
 */
 
 `include "bsg_defines.v"
--- a/hard/ultrascale_plus/bsg_mem/bsg_mem_1r1w_sync_mask_write_byte.v
+++ b/hard/ultrascale_plus/bsg_mem/bsg_mem_1r1w_sync_mask_write_byte.v
@@ -5,7 +5,8 @@
 * Write mode: No-change | Read mode: No-change
 * Note:
 * There are 2 basic BRAM library primitives, RAMB18E2 and RAMB36E2 in Vivado
-* Both of them support byte write enable
+* There is 1 URAM library primitive, URAM288 in Vivado
+* All of them support byte write enable
 *
 * Refer to Vivado Design Suite User Guide: Synthesis (UG901), Byte Write Enable (Block RAM)
 * https://docs.xilinx.com/v/u/2019.1-English/ug901-vivado-synthesis
@@ -55,9 +56,9 @@ module bsg_mem_1r1w_sync_mask_write_byte #(parameter `BSG_INV_PARAM(width_p)
   else
   begin: nz
 
-  /* WARNING: This implementation will use BRAM/URAM inference.
+  /* WARNING: Vivado will automatically choose between BRAM and URAM.
    *
-   * We also can support URAM inference
+   * We also can support URAM-only inference
    * (https://github.com/bespoke-silicon-group/basejump_stl/pull/564/files)
    * if we provide a hardened switch file which can choose between
    * BRAM and URAM inference based on depth and width parameterizations.
--- a/hard/ultrascale_plus/bsg_mem/bsg_mem_1r1w_sync_mask_write_byte.v
+++ b/hard/ultrascale_plus/bsg_mem/bsg_mem_1r1w_sync_mask_write_byte.v
@@ -1,0 +1,84 @@
+/*
+* bsg_mem_1r1w_sync_mask_write_byte.v
+*
+* 2-ports byte write mask block ram for xilinx ultrascale or ultrascale plus FPGA
+* Write mode: No-change | Read mode: No-change
+* Note:
+* There are 2 basic BRAM library primitives, RAMB18E2 and RAMB36E2 in Vivado
+* Both of them support byte write enable
+*
+* Refer to Vivado Design Suite User Guide: Synthesis (UG901), Byte Write Enable (Block RAM)
+* https://docs.xilinx.com/v/u/2019.1-English/ug901-vivado-synthesis
+*
+* reads are synchronous
+*/
+
+`include "bsg_defines.v"
+
+module bsg_mem_1r1w_sync_mask_write_byte #(parameter `BSG_INV_PARAM(width_p)
+                                         , parameter `BSG_INV_PARAM(els_p)
+                                         // semantics of "1" are write occurs, then read
+                                         // the other semantics cannot be simulated on a hardened, non-simultaneous
+                                         // 1r1w SRAM without changing timing.
+                                         // fixme: change to write_then_read_same_addr_p
+                                         , parameter read_write_same_addr_p=0
+                                         , parameter latch_last_read_p=0 
+                                         , parameter addr_width_lp=`BSG_SAFE_CLOG2(els_p)
+                                         , parameter harden_p=0
+                                         , parameter disable_collision_warning_p=0
+                                         , parameter write_mask_width_lp = width_p>>3
+                                         , parameter enable_clock_gating_p=0
+                                         )
+   (input   clk_i
+    , input reset_i
+
+    , input                     w_v_i
+    // for each bit set in the mask, a byte is written
+   ,input [`BSG_SAFE_MINUS(write_mask_width_lp, 1):0] w_mask_i
+    , input [addr_width_lp-1:0] w_addr_i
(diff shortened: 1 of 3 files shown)


dpetrisko:
Looks good!

- Do we think regular code should fail or warning?
- maybe we should also add 1r1w byte mask which presumably has the same mapping problem? https://github.com/bespoke-silicon-group/basejump_stl/pull/542
- thinking about how to combine BRAM and URAM inference https://github.com/bespoke-silicon-group/basejump_stl/pull/564/files. we could do something similar to our ASIC flow and have a switch about how which ram to instantiate. I.e. 

- generate a bsg_mem_1rw_sync.v based on JSON file like our ASIC flow
- which based on size will then call
- hard/ultrascale_plus/bsg_mem_1r1w_sync_bram.v
or
- hard/ultrascale_plus/bsg_mem_1r1w_sync_uram.v

gaozihou:
Thanks Dan!

1. Yeah I think it should be warning..
2. I think 1r1w byte mask will have the same problem, which will require hardened module for ultrascale plus
3. I agree that we should combine BRAM and URAM implementations of byte mask mem, but having bsg_mem_1r1w_sync_bram.v and bsg_mem_1r1w_sync_uram.v would not solve the problem (bsg_mem_1r1w_sync_bram.v will have the some problem as bsg_mem_1rw_sync.v). I guess we could choose between bsg_mem_1r1w_sync_mask_write_byte_bram.v and bsg_mem_1r1w_sync_mask_write_byte_uram.v to combine BRAM and URAM implementations.

drichmond:
This is not the first time we've run into this. The bitmasked rams have an even worse problem.

dpetrisko:
> Yeah I think it should be warning..

How about this @taylor-bsg ?
```
`define BSG_VIVADO_SYNTH_WARNING \
  $warning("Vivado mapping may not be optimal for module %m; check if there exists hard/ultrascale_plus/%m.v");
  
```

> I think 1r1w byte mask will have the same problem, which will require hardened module for ultrascale plus
If we merge #542 first, I think we should add a hardened variant of the 1r1w to this PR as well.  If this merges first, then we'll need to add the variant to #542.

>I guess we could choose between bsg_mem_1r1w_sync_mask_write_byte_bram.v and bsg_mem_1r1w_sync_mask_write_byte_uram.v to combine BRAM and URAM implementations.

Yeah this was what I was suggesting i.e.
```
project/hard/v/bsg_mem_1rw_sync.v:
...
if (width == 64 && els == 1024)
  bsg_mem_1rw_sync_uram
....
else if (width == 32 && els == 512)
  bsg_mem_1rw_sync_bram
 else
  bsg_mem_1rw_sync_bram
```

I think let's merge as is but leave a comment
```
// This implementation will use BRAM inference. We also can support URAM inference (https://github.com/bespoke-silicon-group/basejump_stl/pull/564/files) if we provide a hardened switch file which can choose between BRAM and URAM inference based on depth and width parameterizations.
```

gaozihou:
> ```
> // This implementation will use BRAM inference. We also can support URAM inference (https://github.com/bespoke-silicon-group/basejump_stl/pull/564/files) if we provide a hardened switch file which can choose between BRAM and URAM inference based on depth and width parameterizations.
> ```

Done, thanks Dan!

gaozihou:
@dpetrisko Added 1r1w implementation (and fixed bug in 1rw..)

Interestingly, Vivado is automatically choosing between URAM and BRAM.. Like in the example below (1rw shown, same for 1r1w), the smaller 32x1024 (32k) core_dmem is mapped to BRAM (36k), while the larger 128x512 (64k) cache_dmem is mapped to URAM..
```
Ultra RAM: Preliminary Mapping  Report (see note below)
+---------------+------------+------------------------+---+---+------------------------+---+---+------------------+---------+--------------+--------------------------+-------------+
|Module Name    | RTL Object | PORT A (Depth x Width) | W | R | PORT B (Depth x Width) | W | R | Ports driving FF | URAM288 | Matrix Shape | FF provided for Pipeline | FF absorbed | 
+---------------+------------+------------------------+---+---+------------------------+---+---+------------------+---------+--------------+--------------------------+-------------+
|cache/data_mem | nz.mem_reg | 512 x 128              | W | R |                        |   |   | Port A           | 2       | 1x1          | 0                        | 0           | 
+---------------+------------+------------------------+---+---+------------------------+---+---+------------------+---------+--------------+--------------------------+-------------+

Block RAM: Preliminary Mapping  Report (see note below)
+--------------------------------------------------+------------+------------------------+---+---+------------------------+---+---+------------------+--------+--------+-----------------+
|Module Name                                       | RTL Object | PORT A (Depth x Width) | W | R | PORT B (Depth x Width) | W | R | Ports driving FF | RAMB18 | RAMB36 | Cascade Heights | 
+--------------------------------------------------+------------+------------------------+---+---+------------------------+---+---+------------------+--------+--------+-----------------+
|bsg_mem_1rw_sync_synth:                           | nz.mem_reg | 1 K x 46(WRITE_FIRST)  | W | R |                        |   |   | Port A           | 1      | 1      | 1,1             | 
|bsg_mem_1rw_sync_mask_write_byte__parameterized0: | nz.mem_reg | 1 K x 32(NO_CHANGE)    | W | R |                        |   |   | Port A           | 0      | 1      |                 | 
+--------------------------------------------------+------------+------------------------+---+---+------------------------+---+---+------------------+--------+--------+-----------------+
```
I guess if we need to force URAM implementation, we may add `(* ram_style = “ultra” *)` to the code..

dpetrisko:
Yeah we need to explicitly specify it. The original motivation for this was for BP, with ~32 small 1KB tag rams being inferred as 16kB BRAMs each. So Xilinx definitely makes a “locally optimal, globally poor” decision in these cases

gaozihou:
If the tag ram is byte-masked, I guess the new code will fix the "~32 small 1KB tag rams being inferred as 16kB BRAMs"? (This was the original inefficient mapping I found in mc pod)

gaozihou:
> Yep, it definitely helps! Basically the first layer is what you noticed -- byte mask == 8x the RAMS, then the second layer is that a lot of times URAM is underutilized because Vivado prioritizes it for large RAMs. With designs with a lot of small RAMs, this sometimes means that it ends up as 100% BRAM 0% URAM and you run out of space even though the FPGA has the resources
> 
> original: 8N BRAM new: N BRAM + X LUTs next: N BRAM + M URAM

Got it makes sense to me, thanks Dan!

Btw I've verified both 1rw and 1r1w new models are synthesizing correctly in Vivado.

dpetrisko:
Great, I would say the only remaining task would be BSG_VIVADO_SYNTH warning if we can get a suggestion from @taylor-bsg if this is sufficient:

```
`define BSG_VIVADO_SYNTH_WARNING \
  $warning("Vivado mapping may not be optimal for module %m; check if there exists hard/ultrascale_plus/%m.v");
```

hard/ultrascale_plus/bsg_mem/bsg_mem_1r1w_sync_mask_write_byte.v
```diff
@@ -0,0 +1,84 @@
...
+
+  if (width_p == 0)
+  begin: z
+    wire unused0 = &{clk_i, w_v_i, w_mask_i, w_addr_i, w_data_i, r_v_i, r_addr_i};
+    assign r_data_o = '0;
+  end
+  else
+  begin: nz
+
+  /* WARNING: This implementation will use BRAM/URAM inference.
```
taylor-bsg, line 58: comments on lines 58 and lines 60 seem contradictory?

hard/ultrascale_plus/bsg_mem/bsg_mem_1r1w_sync_mask_write_byte.v
```diff
@@ -0,0 +1,84 @@
...
+   *
+   * We also can support URAM inference
```
taylor-bsg, line 60: I don't believe that URAM supports byte masks? Seems like it would be very wasteful?

hard/ultrascale_plus/bsg_mem/bsg_mem_1r1w_sync_mask_write_byte.v
```diff
@@ -0,0 +1,84 @@
...
+   * (https://github.com/bespoke-silicon-group/basejump_stl/pull/564/files)
+   * if we provide a hardened switch file which can choose between
+   * BRAM and URAM inference based on depth and width parameterizations.
+   */
+
+    logic [width_p-1:0] mem [els_p-1:0];
+
+    for(genvar i = 0; i < write_mask_width_lp; i++)
+      begin: write
+        always_ff @(posedge clk_i)
```
taylor-bsg, line 70: did we confirm that this synthesizes into a byte masked BRAM? Perhaps a comment that indicates the Xilinx doc that says this is how you synthesize a byte masked BRAM?

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_byte.v
```diff
@@ -0,0 +1,77 @@
...
+  begin: z
+    wire unused0 = &{clk_i, v_i, w_i, addr_i, data_i, write_mask_i};
+    assign data_o = '0;
+  end
+  else
+  begin: nz
+
+  /* WARNING: This implementation will use BRAM/URAM inference.
+   *
+   * We also can support URAM inference
```
taylor-bsg, line 51: URAMs do not support byte masking, correctly?

hard/ultrascale_plus/bsg_mem/bsg_mem_1r1w_sync_mask_write_byte.v
gaozihou: I confirmed from the netlist (Vivado design checkpoint) that it got synthesized into either byte-masked BRAM or byte-masked UARM, depending on the memory sizes.. Smaller ones get synthesized into BRAM, larger ones get synthesized into URAM.

gaozihou: In Vivado Design Suite User Guide: Synthesis (UG901), Byte Write Enable (Block RAM) (https://docs.xilinx.com/v/u/2019.1-English/ug901-vivado-synthesis), page 130-137, there are example codes that shows the coding techniques to let Vivado infer BRAM/URAM in synthesis. Indeed, I had to follow the example code line-to-line to make Vivado synthesize correctly.. Very inflexible.

hard/ultrascale_plus/bsg_mem/bsg_mem_1rw_sync_mask_write_byte.v
gaozihou: > URAMs do not support byte masking, correctly?

Hi prof, I double checked the Vivado synthesis netlist (1rw byte masked URAM), there is a byte write enable port (BWE_A[8:0], shown in the image below). I guess that is for byte masking?

Screenshot 2022-06-27 120423

hard/ultrascale_plus/bsg_mem/bsg_mem_1r1w_sync_mask_write_byte.v
gaozihou: Comments added to the files. Thanks prof!

taylor-bsg:
Thanks Paul. The comments are an improvement. But can you do a sweep on the
headers so that it makes it very clear what RAMs each module can map to?

M

gaozihou:
> Thanks Paul. The comments are an improvement. But can you do a sweep on the headers so that it makes it very clear what RAMs each module can map to? M
Thanks for the suggestion prof! Just to clarify, you meant sweeping the Vivado header files or sweeping the netlist?

taylor-bsg:
The header comments of the source files you are pull requesting.

gaozihou:
> The header comments of the source files you are pull requesting.

Oh got it, sure will do!

gaozihou:
Header comments updated!

dpetrisko (approved):
Can confirm this is working on vivado 2020.1 and 2022.1

dpetrisko:
@taylor-bsg @gaozihou @drichmond @tommydcjung 

Anyone have any last feedback or can we merge?

PR #575, opened 2022-06-29
Async fifo with cleaner CDC

dpetrisko:
This PR fixes an apparent problem with the async fifo when is connected to a module which unconditionally latches the output data (without considering the valid).

```
We're seeing a timing violation by connecting an async fifo to bsg_serial_in_parallel_out:
https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_async/bsg_async_fifo.v
https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_dataflow/bsg_serial_in_parallel_out.v
What's happening is the async fifo memory is getting updated with the write clock. Then, the data is asynchronously read by the reader. Note: the rvalid is low at this point, as expected. However, the bsg_serial_in_parallel_out updates its data every cycle, regardless of valid. Therefore, the data gets corrupted. It doesn't affect correctness as the valid signal comes a cycle after the data has settled. 
```

See header comment for further details

### DIFF
--- a/bsg_async/bsg_async_fifo.v
+++ b/bsg_async/bsg_async_fifo.v
@@ -19,7 +19,14 @@ module bsg_async_fifo #(parameter `BSG_INV_PARAM(  lg_size_p )
                         // the data bits to allow for better control optimization.
                         // control_width_p is how many of the width_p bits are control bits;
                         // these bits should be at the top of the array
-                        , parameter   control_width_p = 0)
+                        , parameter   control_width_p = 0
+                        // When the fifo is empty, the rdata output of this fifo changes based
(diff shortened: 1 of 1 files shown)


gaozihou (approved):
I confirm that this PR fixes the possible dynamic timing violations in simulation & actual silicon.

Alternatively we could use bsg_dff_en_bypass to hold previously read value, but that would require extra flops, maybe we could add support to that in the future..

dpetrisko:
>Alternatively we could use bsg_dff_en_bypass to hold previously read value, but that would require extra flops, maybe we could add support to that in the future..

Yeah that's interesting, could save dynamic power but it would depend how frequently the fifo is used. Let's defer for later

PR #580, opened 2022-07-21
Extra case for XSIM

dpetrisko:
Not sure what a default is here that works in all cases.  We're running into problems with BlackParrot+XSIM, so raising this for tracking


dpetrisko:
Side note: there's a macro for vivado sim, but not for synth...

taylor-bsg:
Maybe the best approach is just to file a big report with Xilinx and play the long game…

dpetrisko:
I think it's worth it to support vivado in the mean time. There are a lot of users that don't have paid commercial tools, but don't want to use Verilator. Let alone FGPA developers

PR #556, opened 2022-02-27
Cache to DRAM Ctrl Update

dpetrisko:
This PR removes an unnecessary arbitration in the bsg_cache_to_dram_ctrl. The DRAM controller can handle pipelined requests, but always returns in-order. Therefore, having a straight transduction from bsg_cache_dma_pkt to bsg_cache_to_dram_ctrl is the more flexible solution. If arbitration is handled by concentration in the network, all links can be serialized into bsg_wormhole_to_cache_dma_stream. If not, all links can fanout with bsg_wormhole_tocache_dma_fanout and be arbitrated before getting into bsg_cache_to_dram_ctrl.

It also removes a manycore specific hash function in bsg_cache_to_dram_ctrl, instead exposing app_add_raw. Hashing can be done outside of this module, depending on usecase.

PR #581, opened 2022-07-22
Adding Word Tracking to BSG Cache

farzamgl:
This PR adds the capability to bsg_cache to service store requests without fetching the cache-line from DRAM by tracking the written words to the cache. Changes include adding this feature to the cache pipeline, FSMs, and updating the various bsg_cache interface converters with the new DMA interface that includes a write-mask for evicting cache-line with partially valid data.

### DIFF
--- a/bsg_cache/bsg_cache.v
+++ b/bsg_cache/bsg_cache.v
@@ -32,7 +32,7 @@ module bsg_cache
     ,parameter dma_data_width_p=data_width_p // default value. it can also be pow2 multiple of data_width_p.
 
     ,localparam bsg_cache_pkt_width_lp=`bsg_cache_pkt_width(addr_width_p,data_width_p)
-    ,localparam bsg_cache_dma_pkt_width_lp=`bsg_cache_dma_pkt_width(addr_width_p)
+    ,localparam bsg_cache_dma_pkt_width_lp=`bsg_cache_dma_pkt_width(addr_width_p, block_size_in_words_p)
     ,localparam burst_size_in_words_lp=(dma_data_width_p/data_width_p)
 
     ,parameter debug_p=0
@@ -58,7 +58,6 @@ module bsg_cache
     ,output logic dma_data_ready_o
 
     ,output logic [dma_data_width_p-1:0] dma_data_o
-    ,output logic [burst_size_in_words_lp-1:0] dma_wmask_o
     ,output logic dma_data_v_o
     ,input dma_data_yumi_i
 
@@ -543,7 +542,6 @@ end
     ,.dma_data_ready_o(dma_data_ready_o)
     
     ,.dma_data_o(dma_data_o)
-    ,.dma_wmask_o(dma_wmask_o)
     ,.dma_data_v_o(dma_data_v_o)
     ,.dma_data_yumi_i(dma_data_yumi_i)
 
--- a/bsg_cache/bsg_cache.v
+++ b/bsg_cache/bsg_cache.v
@@ -303,7 +303,8 @@ end
       ,data_v_r
       ,valid_v_r
       ,lock_v_r
-      ,tag_v_r} <= '0;
+      ,tag_v_r
+      ,track_data_v_r} <= '0;
     end
     else begin
       if (v_we) begin
@@ -437,6 +438,7 @@ end
   logic sbuf_empty_lo, tbuf_empty_lo;
   logic [lg_ways_lp-1:0] chosen_way_lo;
   logic select_snoop_data_r_lo;
+  logic miss_track_data_we_lo;
 
   bsg_cache_miss #(
     .addr_width_p(addr_width_p)
@@ -468,6 +470,8 @@ end
     ,.dma_way_o(dma_way_lo)
     ,.dma_addr_o(dma_addr_lo)
     ,.dma_done_i(dma_done_li)
+
+    ,.track_data_we_o(miss_track_data_we_lo)
 
     ,.stat_info_i(stat_mem_data_lo)
 
@@ -525,6 +529,8 @@ end
     ,.dma_way_i(dma_way_lo)
     ,.dma_addr_i(dma_addr_lo)
     ,.done_o(dma_done_li)
+
+    ,.track_data_we_i(miss_track_data_we_lo)
 
     ,.snoop_word_o(snoop_word_lo)
     
--- a/bsg_cache/bsg_cache.v
+++ b/bsg_cache/bsg_cache.v
@@ -22,6 +22,7 @@ module bsg_cache
     ,parameter `BSG_INV_PARAM(block_size_in_words_p)
     ,parameter `BSG_INV_PARAM(sets_p)
     ,parameter `BSG_INV_PARAM(ways_p)
+    ,parameter `BSG_INV_PARAM(word_tracking_p)
 
     // Explicit size prevents size inference and allows for ((foo == bar) << e_cache_amo_swap)
     ,parameter [31:0] amo_support_p=(1 << e_cache_amo_swap)
@@ -30,8 +31,9 @@ module bsg_cache
     // dma burst width
     ,parameter dma_data_width_p=data_width_p // default value. it can also be pow2 multiple of data_width_p.
 
-    ,parameter bsg_cache_pkt_width_lp=`bsg_cache_pkt_width(addr_width_p,data_width_p)
-    ,parameter bsg_cache_dma_pkt_width_lp=`bsg_cache_dma_pkt_width(addr_width_p)
+    ,localparam bsg_cache_pkt_width_lp=`bsg_cache_pkt_width(addr_width_p,data_width_p)
+    ,localparam bsg_cache_dma_pkt_width_lp=`bsg_cache_dma_pkt_width(addr_width_p)
+    ,localparam burst_size_in_words_lp=(dma_data_width_p/data_width_p)
 
     ,parameter debug_p=0
   )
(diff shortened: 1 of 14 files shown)


bsg_cache/bsg_cache_miss.v
```diff
@@ -379,35 +407,61 @@ module bsg_cache_miss
...
+        stat_mem_v_o = dma_done_i & st_tag_miss_op;
+        stat_mem_w_o = dma_done_i & st_tag_miss_op;
+        stat_mem_data_out.dirty = {ways_p{1'b1}};
+        stat_mem_data_out.lru_bits = chosen_way_lru_data;
+        stat_mem_w_mask_out.dirty = chosen_way_decode;
+        stat_mem_w_mask_out.lru_bits = chosen_way_lru_mask;
+
+        // set the tag and the valid bit to 1'b1 for the chosen way on store tag miss.
+        tag_mem_v_o = dma_done_i & st_tag_miss_op;
+        tag_mem_w_o = dma_done_i & st_tag_miss_op;
```
tommydcjung, line 427: tag_mem_w_o always 1'b1?

bsg_cache/bsg_cache_tbuf.v
```diff
@@ -0,0 +1,180 @@
...
+    .clk_i(clk_i)
+    ,.data_i({way_i, addr_i})
+    ,.el0_en_i(el0_enable)
+    ,.el1_en_i(el1_enable)
+    ,.mux0_sel_i(mux0_sel)
+    ,.mux1_sel_i(mux1_sel)
+    ,.el0_snoop_o({el0_way, el0_addr})
+    ,.el1_snoop_o({el1_way, el1_addr})
+    ,.data_o({way_o, addr_o})
+  );
```
tommydcjung, line 138: I think this sbuf_queue and its control logic can be refactored and be used in both tbuf and sbuf

bsg_cache/bsg_cache_dma.v
```diff
@@ -66,6 +69,9 @@ module bsg_cache_dma
     ,output logic [ways_p-1:0][dma_data_width_p-1:0] data_mem_data_o
     ,input [ways_p-1:0][dma_data_width_p-1:0] data_mem_data_i
 
+    ,input track_miss_i
+    ,input [ways_p-1:0][block_size_in_words_p-1:0] track_mem_data_i
```
tommydcjung, line 73: In SEND_EVICT_DATA, data_mem can be selectively read for the words with valid track bit to save energy? For invalid track bit, zero data can be sent instead.

bsg_cache/bsg_cache_to_axi.v
```diff
@@ -12,6 +12,7 @@ module bsg_cache_to_axi
   #(parameter `BSG_INV_PARAM(addr_width_p)
     ,parameter `BSG_INV_PARAM(block_size_in_words_p)
     ,parameter `BSG_INV_PARAM(data_width_p)
+    ,parameter `BSG_INV_PARAM(mask_width_p)
```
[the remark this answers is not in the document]
farzamgl: The `block_size_in_words_p` in this module is meant to be `cache_line_width/dma_data_width`. `mask_width_p` is `cache_line_width/data_width`.

bsg_cache/bsg_cache.v
```diff
@@ -245,6 +246,38 @@ module bsg_cache
...
+  // track_mem
+  //
+  logic track_mem_v_li;
+  logic track_mem_w_li;
+  logic [lg_sets_lp-1:0] track_mem_addr_li;
+  logic [ways_p-1:0][block_size_in_words_p-1:0] track_mem_data_li;
+  logic [ways_p-1:0][block_size_in_words_p-1:0] track_mem_w_mask_li;
+  logic [ways_p-1:0][block_size_in_words_p-1:0] track_mem_data_lo;
+
+if (word_tracking_p) begin
```
tommydcjung, line 259: Need to label the if block

bsg_cache/bsg_cache_to_axi.v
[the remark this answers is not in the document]
tommydcjung: Okay. maybe it should be updated so that data_width_p -> dma_data_width_p, addr_width_p -> dma_addr_width_p, block_size_in_words_p -> dma_burst_len_p, mask_width_p -> dma_mask_width_p for clarity.

dpetrisko:
We'll need to reconcile this one eventually: https://github.com/bespoke-silicon-group/basejump_stl/pull/556/files. I'm fine with merging word tracking ahead of time though, because we'll be taping out the old version of #556 regardless.

QQ: Do we have a testbench setup where we test dma_data_width_p > data_width_p? I think with this change, it'll be important to have that validated, as it would be very easy to make a mistake on masking data words vs dma words

bsg_cache/bsg_cache.v
```diff
@@ -245,6 +246,38 @@ module bsg_cache
     ,.data_o(data_mem_data_lo)
   );
 
+
+  // track_mem
```
dpetrisko, line 250: Just wondering, is it required to have track mem separate from stat mem? Stat mem and track mem are in an awkward size range where hardening is not very efficient, but together would be in that range for large caches. Is it possible to combine them or is there a structural hazard I'm not considering?

dpetrisko: I see because it's read at the beginning of pipeline but written at the end. The other possibility would be combining as 1r1w which is a much more murky tradeoff. Never mind!

bsg_cache/bsg_cache.v
```diff
@@ -245,6 +246,38 @@ module bsg_cache
...
+  //
+  logic track_mem_v_li;
```
dpetrisko, line 252: Resolved my other comment because I thought about this a bit more clearly. With the tbuf to resolve the hazard, it seems possible to combine stat mem and track mem. Either simply make the whole thing 1r1w (which may be efficient for these small sizes) or:
- Read stat_track_mem on every cycle before TL
- write from TBUF opportunistically which happens either:
-  1. during true bubble
-   2. Use latch_last_read to avoid rereading stats for the same cache line
-  3. force a bubble if neither is true and tbuf is full. This last point may kill the idea as random access patterns could suffer performance hits

1r1w seems enticing, though

dpetrisko: Combining with tag has about the same considerations, though provides slightly smaller benefits as tag already tends to be larger

farzamgl:
> QQ: Do we have a testbench setup where we test dma_data_width_p > data_width_p? I think with this change, it'll be important to have that validated, as it would be very easy to make a mistake on masking data words vs dma words

I believe the wormhole testbench sweep tests for different dma data ratios.

bsg_cache/bsg_cache.v
farzamgl: I'm not sure if I understand the proposal. The difference between stat and track mem is that, outside miss handling,  stat mem is written at TV stage(for all commands), and track mem is written at the tbuf output(for full word stores) and read at input stage(for ld/AMO/partial stores). How can we circumvent this by using 1r1w?

dpetrisko:
> I believe the wormhole testbench sweep tests for different dma data ratios.

Nice, looks like it does!

bsg_cache/bsg_cache_miss.v
```diff
@@ -45,12 +49,17 @@ module bsg_cache_miss
...
+    // from track buffer
+    ,input tbuf_empty_i
+
     // to dma engine
     ,output bsg_cache_dma_cmd_e dma_cmd_o
     ,output logic [lg_ways_lp-1:0] dma_way_o
     ,output logic [addr_width_p-1:0] dma_addr_o
     ,input dma_done_i
 
+    ,output logic track_data_we_o
```
tommydcjung, line 61: I think there should be a comment here why track_data_we_o is necessary. This is for flopping the track_mem output to be used by dma_engine, so that we are not creating a new critical path including the clk-to-q of the track_mem.

tommydcjung (approved):

bsg_cache/bsg_cache.v
taylor-bsg: image (9)
image (8)

Yeah I grilled Farzam about this a while back but he convinced me that it did not make sense. They live in different places in the pipeline logically. The area overhead generally is about 9%, and the track memory is about 1/2 the size of the tag memory for a 32-bit architecture, 8-word lines, 4-way set associative cache, not taking up too much space in the floorplan.

dpetrisko: It's double that for 8-way though, which is the BP use-case. And for 8-bank 8-way 128 set, like an example BP memory system it starts to be a decent chunk of system area. 

Farzam, 1R1W would allow the memory to always be read in TL and always read/written in TV, thus obviating the need for the TBUF entirely, no?

Let me do a bit of back of envelope area analysis...

farzamgl: > Farzam, 1R1W would allow the memory to always be read in TL and always read/written in TV, thus obviating the need for the TBUF entirely, no?

I see what you mean. I think it's possible to do the 1R1W however we won't get rid of all the TBUF functionalities. For example, the read at the input stage can include stale data because the commands at TL and TV can write to that mem after/during the read, so we need to have some forwarding logic there. I've to think a bit more about the possible control flow changes if we want to do this.

tommydcjung: A write to stat_mem by a store in TV, and a write to track_mem by tbuf can go to different addresses on the same cycle... so they are merged together,  it might actually need to be 1r2w?

farzamgl:
@taylor-bsg @dpetrisko I think the PR is ready. Unless there are any more comments, can someone with authorization merge?

PR #582, opened 2022-07-22
bsg_cache_word_tracking wormhole change

tommydcjung:
- Adding read, masked write, and non-masked write to the cache dma wormhole packet.
- Updating cache wormhole and  test_dram related modules.
- Making change in miss handling logic to treat SM with all the mask bits as 1'b1 like SW.


farzamgl (approved):
Looks good! Can you remove the commented code lines and merge the commits into 1 with a message before merging?

Issue #400, opened 2021-02-16
Add subbanked SRAM wrapper


dpetrisko:
```
bsg_mem_1rw_sync_subbanked
#(parameter width_p="inv"
   , parameter els_p="inv"
   , parameter latch_last_read_p=0

   // 1 = bit
   // 8 = byte
   // width = no mask, mask ignored
   // We set this as a separate parameter so that we can substitute out the type of ram appropriately
   , parameter mask_granularity_p=1
   , parameter mask_width_lp = width_p/mask_granularity_p
   
    , parameter num_subbank_p=1
    // Don't support depth subbanks due to conflicts
    )
  (input clk_i
    , input reset_i

    , input [num_subbank_p-1:0] v_i
    , input [num_subbank_p-1:0] w_i
    , input [num_subbank_p-1:0][mask_width_lp-1:0] w_mask_i
    , input [num_subbank_p-1:0][width_p-1:0] data_i

    , output [num_subbank_p-1:0][width_p-1:0] data_o
    );
```

dpetrisko:
```
bsg_mem_1rw_sync_subbanked
#(parameter width_p="inv"
   , parameter els_p="inv"
   , parameter latch_last_read_p=0

  , localparam lg_els_lp = `BSG_SAFE_CLOG2(els_p)

   // 1 = bit
   // 8 = byte
   // width = no mask, mask ignored
   // We set this as a separate parameter so that we can substitute out the type of ram appropriately
   , parameter mask_granularity_p=1
   , parameter mask_width_lp = width_p/mask_granularity_p
   
    , parameter num_subbank_p=1
    // Don't support depth subbanks due to conflicts
    , parameter subbank_width_lp=width_p/num_subbank_p
    )
  (input clk_i
    , input reset_i

    , input [num_subbank_p-1:0] v_i
    , input [num_subbank_p-1:0] w_i
    , input [num_subbank_p-1:0][mask_width_lp-1:0] w_mask_i
    , input [num_subbank_p-1:0][subbank_width_lp-1:0] data_i
    , input [lg_els_lp-1:0] addr_i

    , output [num_subbank_p-1:0][subbank_width_lp-1:0] data_o
    );
```

EDIT 09/27: For more interesting use-cases, v_i and w_i must be able to toggle independently per subbank

PR #585, opened 2022-09-22
Code for bsg_mem_1rw_sync_subbanked SRAM

KinzaQamar:
This PR covers code that can convert small logical SRAMs to large SRAM. There are three variants of the code:
1) No mask 
2) Bit mask
3) Byte mask

Since, it is a 1rw SRAM reads and writes from small SRAMs would not happen simultaneously. The future work would be to add 1r1w sync backing SRAM to read and write simultaneously from small SRAMs .
The code is up and running on Verilator.


bsg_mem/bsg_mem_1rw_sync_mask_write_bit_subbanked.v
```diff
@@ -0,0 +1,71 @@
...
+  	, input w_i
+  	, input [num_subbank_p-1:0][subbank_width_lp-1:0] w_mask_i
+  	, input [num_subbank_p-1:0][subbank_width_lp-1:0] data_i
+  	, input [els_lp-1:0] addr_i
+  	, output logic [num_subbank_p-1:0][subbank_width_lp-1:0] data_o
+  );
+
+	wire [width_p-1:0] w_mask_expand;
+
+	bsg_expand_bitmask #(
```
dpetrisko, line 40: I think this is incorrect, it's just taking the first bit of the mask and using it for all bits in the mask. You should just directly assign w_mask_i to the RAM

Can you check why this bug is not detected in your testbench?
The sequence
```
v_i = 1 w_i = 0 -> data_o = 0
v_i = 1 w_i = 1 w_mask_i = 111110 data_i = 1
v_i = 1 w_i = 0 -> data_o = 1
```
should fail with the current code

dpetrisko:
Hi, thanks for the PR.  There's a bug in the code/testbench I've described.  This module is also not quite what we're looking for, since it's just a static wrapper.  I've updated the issue https://github.com/bespoke-silicon-group/basejump_stl/issues/400 to show the desired interface (independent v_i/w_i, same addr for every bank). By manipulating the masks, it should be possible to support the new interface with a single backing SRAM as well, though it is not as trivial

bsg_mem/bsg_mem_1rw_sync_mask_write_bit_subbanked.v
KinzaQamar: Since the width_p is 1. I was getting a warning while assigning the whole mask. Then it shouldn't be 1

dpetrisko: @KinzaQamar try with width_p = 2, the current code will fail as well

bsg_mem/bsg_mem_1rw_sync_mask_write_bit_subbanked.v
```diff
@@ -0,0 +1,99 @@
...
+ *
+ */
+
+`include "bsg_defines.v"
+
+module bsg_mem_1rw_sync_mask_write_bit_subbanked
+  #(parameter `BSG_INV_PARAM(width_p)
+    , parameter `BSG_INV_PARAM(els_p)
+    , parameter latch_last_read_p = 0
+    , parameter num_subbank_p = 4
```
dpetrisko, line 21: use BSG_INV_PARAM here as well

bsg_mem/bsg_mem_1rw_sync_mask_write_bit_subbanked.v
```diff
@@ -0,0 +1,99 @@
...
+
+      // Don't support depth subbanks due to conflicts
+    , localparam subbank_width_lp = width_p/num_subbank_p
+    , localparam els_lp = `BSG_SAFE_CLOG2(els_p)
+  )
+  (   input clk_i
+  	, input reset_i
+   	, input [num_subbank_p-1:0] v_i
+  	, input w_i
```
dpetrisko, line 30: Please add this comment

// For subbanked SRAMs, all subbanks must either read or write at once. There could be an implementation
//   to support independent subbank r/w by using a 1r1w backing SRAM. This may make sense in an FPGA
//   environment, but we leave this to future work

bsg_mem/bsg_mem_1rw_sync_mask_write_bit_subbanked.v
```diff
@@ -0,0 +1,99 @@
...
+      end // (latch_last_read_p):llr
+      
+      else begin: no_llr
+        assign data_o = data_lo;
+      end
+    
+    end // for (genvar i = 0; i < num_subbank_p; i++)
+
+    always@(*) begin
+      assert (`BSG_IS_POW2(width_p) && `BSG_IS_POW2(els_p));
```
dpetrisko, line 95: Rather than simulation time, this should be an elaboration time assertion.

if (`BSG_IS_POW2)...
  $error()

not in an always block. This will cause the build to fail, which is the desired behavior. Please add a descriptive $error message as well

dpetrisko: applies to other modules as well

bsg_mem/bsg_mem_1rw_sync_mask_write_bit_subbanked.v
dpetrisko: applies to other modules as well

bsg_mem/bsg_mem_1rw_sync_mask_write_bit_subbanked.v
dpetrisko: applies to other modules as well

bsg_mem/bsg_mem_1rw_sync_subbanked.v
```diff
@@ -0,0 +1,90 @@
...
+   	, input [num_subbank_p-1:0] v_i
+  	, input w_i
+  	, input [num_subbank_p-1:0][subbank_width_lp-1:0] data_i
+  	, input [els_lp-1:0] addr_i
+  	, output logic [num_subbank_p-1:0][subbank_width_lp-1:0] data_o
+  );
+
+    logic [num_subbank_p-1:0][subbank_width_lp-1:0] data_lo;
+
+    bsg_mem_1rw_sync #(
```
dpetrisko, line 38: This breaks for subbank > 1, try writing to only 1 subbank and reading from it

dpetrisko: The solution is to instantiate a byte-mask RAM for subbank > 1

KinzaQamar: Since this is a no mask SRAM, it works only for sub bank == 1. There is an assertion below which would fail if the subbank_p is other than 1.

We already have a byte-mask RAM. You're saying to instantiate it inside this module ?

dpetrisko: Yes. There is an overhead to supporting byte-mask on a hardened SRAM. For designs which do not need it (subbank == 1), we do not wish to incur the overhead. When subbank > 1, we have no choice but to use a byte mask RAM unless we want to instantiate multiple hardened RAMs

dpetrisko (approved):
LGTM!

dpetrisko:
@taylor-bsg does this look good to you? The basic idea is explained in #400. It's the inverse operation of bsg_mem_1rw_sync_banked and can be used to "fuse" together several logical subbanks into a single physical bank. 

The restriction is that each subbank can only be read or written at the same time, which is a common access pattern. It could be expanded to support arbitrary read-write patterns, but that's future work

bsg_mem/bsg_mem_1rw_sync_mask_write_bit_subbanked.v
```diff
@@ -0,0 +1,104 @@
...
+
+module bsg_mem_1rw_sync_mask_write_bit_subbanked
+  #(parameter `BSG_INV_PARAM(width_p)
+    , parameter `BSG_INV_PARAM(els_p)
+    , parameter `BSG_INV_PARAM(latch_last_read_p)
+    , parameter `BSG_INV_PARAM(num_subbank_p)
+
+      // Don't support depth subbanks due to conflicts
+    , localparam subbank_width_lp = width_p/num_subbank_p
+    , localparam els_lp = `BSG_SAFE_CLOG2(els_p)
```
tommydcjung, line 29: els_lp -> lg_els_lp

bsg_mem/bsg_mem_1rw_sync_mask_write_bit_subbanked.v
```diff
@@ -0,0 +1,104 @@
...
+      ,.w_i(|w_i)
+      ,.addr_i(addr_i)
+      ,.data_i(data_i)
+      ,.w_mask_i(w_mask_lo)
+      ,.data_o(data_lo)
+    );
+
+    wire [num_subbank_p-1:0] read_en;
+
+    for (genvar i = 0; i < num_subbank_p; i++) begin
```
tommydcjung, line 68: for-block should be labeled if there are module instantiations inside.

bsg_mem/bsg_mem_1rw_sync_mask_write_bit_subbanked.v
```diff
@@ -0,0 +1,104 @@
...
+      ,.w_mask_i(w_mask_lo)
+      ,.data_o(data_lo)
+    );
+
+    wire [num_subbank_p-1:0] read_en;
+
+    for (genvar i = 0; i < num_subbank_p; i++) begin 
+      assign read_en[i] = v_i[i] & ~w_i;
+
+      if (latch_last_read_p) begin: llr
```
tommydcjung, line 71: bsg_mem_1rw_sync_mask_write_bit_synth already has bsg_dff_en_bypass for latch_last_read_p, and there is another one here?

KinzaQamar: This is because we are ORing the v_i bits and will end up reading a subbank for which v_i would be loww.

bsg_mem/bsg_mem_1rw_sync_subbanked.v
```diff
@@ -0,0 +1,125 @@
...
+        ,.v_i(|v_i)
+        ,.w_i(|w_i)
+        ,.addr_i(addr_i)
+        ,.data_i(data_i)
+        ,.data_o(data_lo)
+      );
+
+    end //no_mask_sram
+
+    if (num_subbank_p > 1) begin: byte_mask_sram
```
tommydcjung, line 63: instead of "if (num_subbank_p > 1)", "else" should be enough?

bsg_mem/bsg_mem_1rw_sync_subbanked.v
```diff
@@ -0,0 +1,125 @@
...
+
+      logic [num_subbank_p-1:0][mask_width_lp-1:0] w_mask_lo;
+
+      for (genvar i = 0; i < num_subbank_p; i++) begin
+        for (genvar j = 0; j < mask_width_lp; j++) 
+          assign w_mask_lo[i][j] = w_mask_i[i][j] & v_i[i];
+      end // for (genvar i = 0; i < num_subbank_p; i++)
+  
+      bsg_mem_1rw_sync_mask_write_byte #(
```
tommydcjung, line 72: What if  width_p=15, and num_subbank_p=3? Couldn't it support that by using bsg_mem_1rw_sync_mask_write_bit instead?

bsg_mem/bsg_mem_1rw_sync_mask_write_bit_subbanked.v
tommydcjung: Okay I see.

bsg_mem/bsg_mem_1rw_sync_mask_write_bit_subbanked.v
```diff
@@ -0,0 +1,104 @@
...
+        );
+      end // (latch_last_read_p):llr
+      
+      else begin: no_llr
+        assign data_o = data_lo;
+      end
+    
+    end // for (genvar i = 0; i < num_subbank_p; i++)
+
+  if (!(`BSG_IS_POW2(width_p) && `BSG_IS_POW2(els_p)))
```
tommydcjung, line 98: Does width_p have to be power of 2? num_subbank_p can be non power of 2.

bsg_mem/bsg_mem_1rw_sync_subbanked.v
KinzaQamar: I see. A check is missing for a byte size sub bank (minimum)

bsg_mem/bsg_mem_1rw_sync_mask_write_bit_subbanked.v
KinzaQamar: If num_subbank_p is a non-power of 2, it will break the check below (width_p% num_subbank_p==0)

bsg_mem/bsg_mem_1rw_sync_subbanked.v
KinzaQamar: Yeah, it would work but then we have to implement a check for num_subbank_p!=0. Otherwise our code will break.

bsg_mem/bsg_mem_1rw_sync_mask_write_bit_subbanked.v
```diff
@@ -0,0 +1,104 @@
+/*
+ *  bsg_mem_1rw_sync_mask_write_bit_subbanked.v
+ *
+ *  This module has the same interface/functionality as
+    bsg_mem_1rw_sync_mask_write_bit.
```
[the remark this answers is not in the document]
KinzaQamar: This is the opposite of bsg_mem_1rw_sync_mask_write_bit_banked, where we are breaking the big SRAM into N small banks based on width and depth. Here, we want a big SRAM as a result of N logical banks.

bsg_mem/bsg_mem_1rw_sync_subbanked.v
tommydcjung: Well, right now, if num_subbank_p is set to 0, wouldn't it just silently fail?

KinzaQamar: Yes, it would be failing on 0.

KinzaQamar:
@tommydcjung have been made!

tommydcjung (approved):

bsg_mem/bsg_mem_1rw_sync_mask_write_bit_subbanked.v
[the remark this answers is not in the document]
tommydcjung: Can you put that info on the comments? where this module should be used, etc.

dpetrisko:
Thanks for the review @tommydcjung !

PR #584, opened 2022-09-10
Update DMC to latest tapeout code

dpetrisko:
This is the version of the DMC which was taped out in TSMC28. Main features from the previous code:
- Dly line is handled internally with a synthesizable variant
- Tag configuration is done internally and standardized to a tag structure
- Added TCALR parameter which allows for period read refreshes to tune the delay line
- Added trace replay capability for bringup of device

PR #587, opened 2022-10-08
Add range to packed arrays

flaviens:
Hi!

Some commercial tool complains about packed dimensions not specifying a range. Please double check that the widths are correct before merging!

Thanks!
Flavien


dpetrisko (approved):
LGTM, seems to work with the existing testbench.

Noting for myself that this may be fixed by #584, but is good to have on its own

PR #588, opened 2022-10-08
Make mobile_ddr_mcp usable out of the box

flaviens:
Maybe replace comments with ifdefs, as looks the best for you.

Thanks!
Flavien

### DIFF
--- a/testing/bsg_dmc/lpddr_verilog_model/mobile_ddr_mcp.v
+++ b/testing/bsg_dmc/lpddr_verilog_model/mobile_ddr_mcp.v
@@ -47,7 +47,21 @@ module mobile_ddr_mcp (
     Dm          
 );
 
-    `include "mobile_ddr_parameters.vh"
+    `ifdef den128Mb
+        `include "128Mb_mobile_ddr_parameters.vh"
+    `elsif den256Mb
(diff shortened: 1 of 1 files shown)


dpetrisko:
```

`ifdef den128Mb
    `include "128Mb_mobile_ddr_parameters.vh"
`elsif den256Mb
    `include "256Mb_mobile_ddr_parameters.vh"
`elsif den512Mb
    `include "512Mb_mobile_ddr_parameters.vh"
`elsif den1024Mb
    `include "1024Mb_mobile_ddr_parameters.vh"
`elsif den2048Mb
    `include "2048Mb_mobile_ddr_parameters.vh"
`else
    // NOTE: Intentionally cause a compile fail here to force the users
    //       to select the correct component density before continuing
    ERROR: You must specify component density with +define+den____Mb.
`endif

```

Yeah, if you wouldn't mind switching to this scheme from: https://github.com/bespoke-silicon-group/basejump_stl/blob/a93d79639c0bab878a674ca176d31176e65c6164/testing/bsg_dmc/lpddr_verilog_model/mobile_ddr.v#L96, would appreciate it

flaviens:
Updated as you suggested!

dpetrisko (approved):
LGTM

PR #589, opened 2022-10-08
Remove unused parameter from tiehi

flaviens:
The param was not used. This could break some backward compatibility, though, but it looks that at least it is never used. Or it is used as a blackbox somewhere and we should keep it.


flaviens:
See #590

PR #590, opened 2022-10-08
Remove unused param from tielo

flaviens:
The param was not used. This could break some backward compatibility, though, but it looks that at least it is never used. Or it is used as a blackbox somewhere and we should keep it.


dpetrisko:
Hi, the harden_p parameter is there for some backend flows, which use it to determine whether to substitute a hardened cell rather than synthesizing. (For tielo, this would be used to ensure correct sizing). You’ll find this parameter throughout basejump_stl and should ignore it from a frontend perspective

flaviens:
Hi Dan, sounds good! I'll set up my suppress flags to ignore it then. Thanks!

PR #591, opened 2022-10-08
Fix division by zero

flaviens:
Some commercial tool gives an unsuppressible error as it probably first does the divisions before assigning tck. Maybe there is a more elegant solution, though.


dpetrisko:
would doing:

```
real tck = 32'hdeadbeef;
```

as an initialization value help?

flaviens:
The same tool now says that these errors are suppressible, and I cannot remember how I got them unsuppressible yesterday ^^'
So let's keep it as it is now, unfortunately `real tck = 32'hdeadbeef;` does not seem to help.
Thank you for your response!

dpetrisko:
Even better! Tool-specific fixes tend to outstay their welcome...

PR #595, opened 2022-10-12
Fix uncompliant assignment

flaviens:
Hi there!

This code was not compliant Verilog.


taylor-bsg (approved):
Legal SystemVerilog, but non synthesizeable.

PR #600, opened 2022-10-13
Fix fatal syntax

flaviens:
Hi there!

The fatal syntax here was incorrect and made the Morty tool unhappy (cf. [this issue](https://github.com/pulp-platform/morty/issues/46)). This should fix it.


dpetrisko:
Interesting, seems some tools are more lax. If the spec requires an error code, I think that should be alright. We'll want to test it out in EDA playground first though

flaviens:
Tested now, it compiles with a commercial simulator :+1:

PR #601, opened 2022-10-18
standardizing harden_p for bsg_mem

dpetrisko:
Standardizing the harden_p scheme for memories. 

This means 
1) removing harden_p from _synth memories 
2) Adding harden_p to all other mems 
3) In hardened directories default harden_p = 1
4) In non-hardened directories default harden_p = 0

### DIFF
--- a/bsg_mem/bsg_mem_1r1w_sync_banked.sv
+++ b/bsg_mem/bsg_mem_1r1w_sync_banked.sv
@@ -34,6 +34,7 @@ module bsg_mem_1r1w_sync_banked
     , parameter `BSG_INV_PARAM(els_p)
     , parameter read_write_same_addr_p=0
     , parameter latch_last_read_p=0
+    , parameter harden_p=0
 
     , parameter num_width_bank_p=1
     , parameter num_depth_bank_p=1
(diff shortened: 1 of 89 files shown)


dpetrisko:
Doubtful that this will break anything as synth memories should never be explicitly instantiated anywhere, and not-synth mems will not be setting this unused parameter

taylor-bsg:
Great suggested change. Do we have an error or warning that fires when if harden_p=1 and it is going to a non-hardened module?

taylor-bsg:
If it is non-leaf, then it should pass harden_p along, rather than swallowing it. This way we don't have to clone the code for intermediate modules that would not be hardened, but whose children are hardened.

dpetrisko:
>Do we have an error or warning that fires when if harden_p=1 and it is going to a non-hardened module?

Can add!

>If it is non-leaf, then it should pass harden_p along, rather than swallowing it. This way we don't have to clone the code for intermediate modules that would not be hardened, but whose children are hardened.

Not sure what you mean here. Are you proposing
- bsg_mem/bsg_mem_1rw_sync.v has harden and $warns if it is set
- bsg_mem/bsg_mem_1rw_sync_synth.v has harden and $warns if set?

Aside $warn vs $error?

dpetrisko:
@taylor-bsg clarification on the above?

dpetrisko:
Moving discussion to #724

PR #602, opened 2022-10-19
Add bsg_fifo_1r1w_rolly, bsg_fifo_1r1w_store_and_forward

Yuan-Mao:
This PR adds modules related to bsg_fifo_1r1w_rolly and bsg_fifo_1r1w_store_and_forward.
This PR also adds random tests with covergroup for bsg_fifo_1r1w_rolly and bsg_fifo_1r1w_store_and_forwardin under testing/.

### DIFF
--- a/bsg_dataflow/bsg_fifo_1r1w_rolly.sv
+++ b/bsg_dataflow/bsg_fifo_1r1w_rolly.sv
@@ -1,0 +1,47 @@
+
+`include "bsg_defines.v"
+
+module bsg_fifo_1r1w_rolly
+  #(parameter `BSG_INV_PARAM(width_p)
+    , parameter `BSG_INV_PARAM(lg_size_p)
+    , parameter ready_THEN_valid_p = 0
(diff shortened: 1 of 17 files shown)


taylor-bsg (commented):
Seems like you have three modules where you could have 1?  I am a little unclear on hardened vs. unhardened. Seems like you are using a synchronous 1r1w in one case and an async in the other case?  Can you just pass in a parameter and have a single code base or are there a lot of differences?

dpetrisko:
I think the hardened vs unhardened is inspired by:
https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_dataflow/bsg_fifo_1r1w_small_hardened.v
https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_dataflow/bsg_fifo_1r1w_small_unhardened.v

I don't remember the tradeoffs @gaozihou considered when making these two modules. When would it be preferable to have "unhardened" as the synchronous RAM can either be hardened or soft?

PR #604, opened 2022-10-25
[bsg_cache] fix track_mem_data_r X when word_tracking_p = 0

tommydcjung:
- basically, when word_tracking_p = 0, track_data_we_i will always be zero in bsg_cache_dma. This means that track_mem_data_r will always be 'X.  Then data_mem_v_o will also become X, and so will tl_ready at the top.
- added word_tracking_p switch to regression_v2
- Added asserting to check if ready_o or v_o ever goes X in testbench, because the testbench didn't fail even when there was x.

### DIFF
--- a/bsg_cache/bsg_cache_dma.v
+++ b/bsg_cache/bsg_cache_dma.v
@@ -303,7 +303,11 @@ module bsg_cache_dma
             // so the counter is incremented to 1.
             counter_clear = 1'b1;
             counter_up = 1'b1;
-            data_mem_v_o = (|track_bits_offset_picked);
+            // When word_tracking_p is off, track_data_we_i is always zero, which means track_mem_data_r will always be 'X;
+            // So if word_tracking_p is off, we tie this to constant 1.
+            data_mem_v_o = word_tracking_p 
+              ?(|track_bits_offset_picked)
+              : 1'b1;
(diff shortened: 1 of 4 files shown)


dpetrisko:
- Can we hardcode track_mem_data_r to zero instead of not using its output? Having Xs hanging around seems dangerous in sim even though synthesis will (probably) optimize it out correctly
- When word_tracking_p == 1, is track_data_we_i guaranteed to be raised before track_mem_data_r is used?

tommydcjung:
> * Can we hardcode track_mem_data_r to zero instead of not using its output? Having Xs hanging around seems dangerous in sim even though synthesis will (probably) optimize it out correctly

I thought about doing that but then we are just hoping that the synthesis tool will do the right thing. as long as track_mem_data_r is not being used, it's safe? I'm fine with either way.

tommydcjung:
> * When word_tracking_p == 1, is track_data_we_i guaranteed to be raised before track_mem_data_r is used?

It has to be, otherwise it will show up as a bug

tommydcjung:
a part of reason why this bug got through is that even if we could replicate the bug in the regression, the testbench could not detect it. So that part has been fixed by adding assertion to check for X

farzamgl (approved):
Looks good to me.

PR #606, opened 2022-10-31
Adding Verilator Timing Check to bsg_nonsynth_clock_gen

stdavids:
This also includes some help macros to bsg_defines.v to do some logical and and logical or operations to defined macros.

### DIFF
--- a/bsg_misc/bsg_defines.v
+++ b/bsg_misc/bsg_defines.v
@@ -124,4 +124,67 @@
 `define rpgroup(x)
 `endif
 
+// verilog preprocessing -> if defined(A) && defined(B) then define C
+`define BSG_DEFIF_A_AND_B(A,B,C) \
+    `undef C \
+    `ifdef A \
+        `ifdef B \
+            `define C \
+        `endif \
+    `endif
+
+// verilog preprocessing -> if defined(A) && !defined(B) then define C
+`define BSG_DEFIF_A_AND_NOT_B(A,B,C) \
+    `undef C \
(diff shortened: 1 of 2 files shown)


bsg_misc/bsg_defines.v
```diff
@@ -124,4 +124,67 @@
...
+// verilog preprocessing -> if !defined(A) && !defined(B) then define C
+`define BSG_DEFIF_NOT_A_AND_NOT_B(A,B,C) \
+    `undef C \
+    `ifndef A \
+        `ifndef B \
+            `define C \
+        `endif \
+    `endif
+
+// verilog preprocessing -> if defined(A) || defined(B) then define C
```
dpetrisko, line 157: `ifdef B \
         `ifndef C \
           `define C \
         `endif
     `endif


Otherwise A && B will define C twice?

dpetrisko: Same with BSG_DEFIF_A_OR_B and BSG_DEFIF_A_OR_NOT_B

dpetrisko: Retracted; seems simulators are fine with this behavior

dpetrisko (approved):
LGTM!

bsg_misc/bsg_defines.v
stdavids: Yes, for posterity I have confirmed this is fine with both Verilator (v5.002) and VCS (L-2016.06-SP2-15).

PR #607, opened 2022-11-10
[idiv] fix some undeclared and unassigned signals

tommydcjung:
- Fixed undeclared signals: add_in0, add_in1.
- Fixed add2_out being 'Z, when bits_per_iter_p=1. It was fixed, because the top bit goes into idiv controller, and it's being flopped there.

### DIFF
--- a/bsg_misc/bsg_idiv_iterative.v
+++ b/bsg_misc/bsg_idiv_iterative.v
@@ -174,13 +174,13 @@ module bsg_idiv_iterative #(parameter width_p=32, parameter bitstack_p=0, parame
     bsg_nor2 #(.width_p(width_p+1)) nor_opA 
        ( .a_i( opA_xnor )
         ,.b_i({(width_p+1){~opA_clr_lo}})
-        ,.o  (add_in0)
+        ,.o  (add1_in0)
         );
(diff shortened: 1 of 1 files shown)


dpetrisko:
How'd you find these, synthesis run? I'd like to reproduce with/without fixes

tommydcjung:
> How'd you find these, synthesis run? I'd like to reproduce with/without fixes

From the vcs build. I set bits_per_iter_p = 1.

```
idiv_iterative.v, 82
  Following variable has never been set any value.
  Source info: add2_out

Lint-[IWU] Implicit wire is used
bsg_idiv_iterative.v, 177
  No type is specified for wire 'add_in0'. Default wire type is being applied
  according to the IEEE spec.
  See the Verilog LRM(1364-2005), section 4.5.

Lint-[IWU] Implicit wire is used
bsg_idiv_iterative.v, 183
  No type is specified for wire 'add_in1'. Default wire type is being applied
  according to the IEEE spec.
  See the Verilog LRM(1364-2005), section 4.5.

```

dpetrisko (approved):
Confirmed errors exist and are fixed!

PR #615, opened 2022-12-07
Clock Mux macro for Ultrascale

mysoreanoop:
Tested with 2/3/4:1 muxes with balanced_p and harden_p -- each synthesizes, FPGA-simulates.

### DIFF
--- a/hard/ultrascale_plus/bsg_misc/bsg_mux.v
+++ b/hard/ultrascale_plus/bsg_misc/bsg_mux.v
@@ -34,9 +34,11 @@ module bsg_mux #(parameter `BSG_INV_PARAM(width_p)
     ,input [lg_els_lp-1:0] sel_i
     ,output [width_p-1:0] data_o
     );
-   
+  
+  // the following macro is for clock muxing only
   if((harden_p == 1) && (balanced_p == 1)
-        && els_p != 1)
+        && (width_p == 1)
+        && (els_p > 1) && (els_p < 5))
     begin: macro
       if(els_p == 2)
         begin: mux2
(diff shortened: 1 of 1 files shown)


hard/ultrascale_plus/bsg_misc/bsg_mux.v
```diff
@@ -0,0 +1,107 @@
...
+    ,input [lg_els_lp-1:0] sel_i
+    ,output [width_p-1:0] data_o
+    );
+  
+  // the following macro is for clock muxing only
+  if((harden_p == 1) && (balanced_p == 1)
+        && (width_p == 1)
+        && (els_p > 1) && (els_p < 5))
+    begin: macro
+      if(els_p == 2)
```
dpetrisko, line 43: let's get rid of the first clause and just do

if (harden_p && balanced_p && els_p == 2)
  begin : macro
else if (harden_p && balanced_p && els_p == 3)
  begin : macro
else if (harden_p && balanced_p && els_p == 4)
  begin : macro
else
  begin : notmacro

hard/ultrascale_plus/bsg_misc/bsg_mux.v
```diff
@@ -0,0 +1,107 @@
+// Xilinx recommends for 3/4:1 clock muxes use 2 2:1 muxing instances
+// https://docs.xilinx.com/r/2021.1-English/ug1387-acap-hardware-ip-platform-dev-methodology/Clock-Multiplexing
+
+`include "bsg_defines.v"
+
+module bsg_macro_clk_mux_2_1
```
dpetrisko, line 6: let's make this an actual `define to avoid introducing more hierarchy. Some synthesis tools unavoidably insert buffers for feedthroughs

dpetrisko (changes requested):
Looks good! minor changes requested

hard/ultrascale_plus/bsg_misc/bsg_mux.v
mysoreanoop: Do you mean something like https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_async/bsg_sync_sync.v ?
Would that avoid the feedthrough buffers? We would still elaborate the macro and instantiate the module within, right?

dpetrisko:
LGTM!

dpetrisko (approved):

PR #618, opened 2022-12-13
Add bsg_tag_bitbang

Yuan-Mao:
This PR adds bsg_tag_bitbang. This module can be used to write data into a bsg tag master.

### DIFF
--- a/bsg_tag/bsg_tag_bitbang.v
+++ b/bsg_tag/bsg_tag_bitbang.v
@@ -1,0 +1,64 @@
+
+`include "bsg_defines.v"
+
+/*
+ * When a new data arrives, tag_clk_r_o will toggle once and then be held
+ *   high until the next new data comes in.
+ *
+ *                -----+
+ * v_i:                |
+ *                     +-----------------------------
+ *
+ *                     +-----+     +-----+     +-----
+ * clk_i:              |     |     |     |     |
+ *                -----+     +-----+     +-----+
+ *
+ *                -----+           +-----------------
+ * tag_clk_r_o:        |           |
+ *                     +-----------+
+ *
+ *                     +-----------------------+
+ * tag_data_r_o:       |                       |
+ *                -----+                       +-----
+ *
+ */                                        
+
+module bsg_tag_bitbang(
+    input clk_i
+  , input reset_i
+  , input data_i
+  , input v_i
+  , output logic ready_then_o // throughput: clk_i / 2
+
+  , output logic tag_clk_r_o
+  , output logic tag_data_r_o
+);
+
+  logic tag_data_valid_r;
+  logic tag_data_r;
+  wire tag_data_valid_n = ~tag_data_valid_r | ~v_i;
+
+  bsg_dff_reset #(
+     .width_p(1)
+    ,.reset_val_p(1)
+  ) tag_data_valid_reg (
+     .clk_i(clk_i)
+    ,.reset_i(reset_i)
+    ,.data_i(tag_data_valid_n)
+    ,.data_o(tag_data_valid_r)
+  );
+
+  bsg_dff_en #(.width_p(1)
+  ) tag_data_reg (
+     .clk_i(clk_i)
+    ,.data_i(data_i)
+    ,.en_i(~tag_data_valid_n)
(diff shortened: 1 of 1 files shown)


Yuan-Mao:
Done!
This test tests bsg_tag_bitbang module with bsg_tag_master and some bsg_tag_clients.
The data width in tag (width_p) and the number of tag client (els_p) can be configured in top.sv.

dpetrisko:
@taylor-bsg seems fine to me, look good to you?

bsg_tag/bsg_tag_bitbang.v
```diff
@@ -0,0 +1,72 @@
+
+`include "bsg_defines.v"
+
+/*
+ * This module takes 1 bit of data and send it out along with a clock pulse.
+ *
+ * When this module receives a bit, tag_clk_r_o becomes low and the
+ *   received data is put on tag_data_r_o on the current cycle. On next cycle,
+ *   tag_clk_r_o will become high and generate a posetive clock edge. The clock
```
taylor-bsg, line 9: typo

bsg_tag/bsg_tag_bitbang.v
```diff
@@ -0,0 +1,72 @@
...
+ */                                        
+
+module bsg_tag_bitbang(
+    input clk_i
+  , input reset_i
+  , input data_i
+  , input v_i
+  , output logic ready_then_o // throughput: clk_i / 2
+
+  , output logic tag_clk_r_o
```
taylor-bsg, line 41: // tag clock is a synchronously generated variable frequency clock that toggles
// once for every data bit that is transmitted

bsg_tag/bsg_tag_bitbang.v
```diff
@@ -0,0 +1,72 @@
...
+  , output logic tag_data_r_o
+);
+
+  logic tag_data_valid_r;
+  logic tag_data_r;
+  wire tag_data_valid_n = ~tag_data_valid_r | ~v_i;
```
taylor-bsg, line 47: BaseJump STL says that we should avoid low true signals in our code.

taylor-bsg (changes requested):
@Yuan-Mao

bsg_tag/bsg_tag_bitbang.v
Yuan-Mao: v_i and tag_data_valid_r are true when it is 1'b1, so they should not be low true signals?

dpetrisko: Yeah it's a little confusing but they're not active low. The v_r signal here doubles the tag clock, while the v_i signal is the trigger. So v_r is default high. When a v_i comes in, it goes low. When it's low, it goes back high. This automatic low->high transition is the rising edge of the tag clock

dpetrisko: Maybe a clarification could be renaming the register to tag_clk_r?

taylor-bsg: tag_data_valid_n does not seem to mean that data is actually valid

Yuan-Mao: I see your point. Perhaps like Dan suggest tag_clk_r could be a better name.

bsg_tag/bsg_tag_bitbang.v
```diff
@@ -0,0 +1,74 @@
...
+
+  // tag clock is a synchronously generated variable frequency clock that toggles
+  //   once for every data bit that is transmitted.
+  , output logic tag_clk_r_o
+  , output logic tag_data_r_o
+);
+
+  logic tag_clk_r;
+  logic tag_data_r;
+  wire tag_clk_n = ~ready_and_o | ~v_i;
```
dpetrisko, line 49: nit, but may be more clear as 
```
// Lower tag clock for one cycle upon a successful handshake
wire tag_clk_n = ~(ready_and_o & v_i)
```

dpetrisko:
@taylor-bsg I think this is good to go unless you have other thoughts. Looks very clean now

dpetrisko (approved):

taylor-bsg (approved):

PR #626, opened 2022-12-20
Updates handshaking conventions for bsg_noc

BrendenPage:
Clarifies the interface of the handshaking conventions for the modules in bsg_noc and their associated testbenches.


dpetrisko (approved):
LGTM!

PR #628, opened 2022-12-20
Clock gate hard macro for Xilinx Ultrascale

mysoreanoop:
+ fixes for bsg_mux (argument reordering)

### DIFF
--- a/hard/ultrascale_plus/bsg_misc/bsg_clkgate_optional.v
+++ b/hard/ultrascale_plus/bsg_misc/bsg_clkgate_optional.v
@@ -1,0 +1,53 @@
+`include "bsg_defines.v"
+
+module bsg_clkgate_optional #(parameter harden_p = 0)
+    
+            (input  clk_i
+             ,input  en_i
+             ,input  bypass_i
+             ,output gated_clock_o
+             );
+
+  if(harden_p)
+    begin: macro
+      // BUFGCE: General Clock Buffer with Clock Enable
+      //         UltraScale
+      // Xilinx HDL Language Template, version 2021.1
+
+      BUFGCE #(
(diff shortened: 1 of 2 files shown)


hard/ultrascale_plus/bsg_misc/bsg_clkgate_optional.v
```diff
@@ -0,0 +1,53 @@
...
+      BUFGCE_inst (
+         .O(gated_clock_o),
+         .CE(en_i),
+         .I(clk_i)
+      );
+
+      // End of BUFGCE_inst instantiation
+    end
+
+`ifndef SYNTHESIS
```
dpetrisko, line 32: I would only put the BUFGCE option (no if-else). The other will always be broken in FPGA synthesis

mysoreanoop:
Requesting review

hard/ultrascale_plus/bsg_misc/bsg_clkgate_optional.v
```diff
@@ -0,0 +1,30 @@
+`include "bsg_defines.v"
+
+module bsg_clkgate_optional
+    #( parameter is_ce_inverted = 1'b0
```
dpetrisko, line 4: Let's leave out these parameters since they don't exist in the "synth" bsg_clk_gate_optional

dpetrisko (changes requested):
Looks good with small fix

hard/ultrascale_plus/bsg_misc/bsg_clkgate_optional.v
mysoreanoop: Updated

dpetrisko (approved):
LGTM!

PR #629, opened 2022-12-20
Bsg misc ready handshake update

BrendenPage:
Updates the ready handshake naming conventions for bsg_misc and associated usage in testbenches.


bsg_misc/bsg_scheduler_dataflow.v
```diff
@@ -105,7 +105,7 @@ module bsg_scheduler_dataflow_entry #(`BSG_INV_PARAM(tag_width_p)
         end
     end
   
-  assign ready_o = ~(| src_tags_v_r);
+  assign busy_o = ~(| src_tags_v_r);
```
dpetrisko, line 108: This should be inverted right?

BrendenPage: Yep! Good catch.

dpetrisko (approved):
LGTM!

PR #631, opened 2022-12-20
[DO NOT MERGE] Removes deprecated ramulator modules

BrendenPage:
bsg_nonsynth_ramulator_hbm was deprecated when #257 removed the bsg_nonsynth_test_dram_channel.v file and module, making these files unusable.

This simply removes the ramulator module and its associated testbench.


dpetrisko:
Hi if we're doing this, can we remove the ramulator import/ as well?

@mrutt92 is ramulator support completely deprecated? Or are there any usages we want to maintain?

@taylor-bsg is it worth moving these modules to a legacy/ folder or just let them go?

mrutt92:
we don't have a use for it although we have a grant where ramulator may become relevant.

dpetrisko:
Oh interesting, okay. Can you recommend the easiest path to un-deprecate the ramulator modules? Can we use the same Dramsim3 Verilog/DPI calls with a ramulator C++ library or are the interfaces different?

PR #636, opened 2023-01-11
Bsg mesosync ready handshake update

BrendenPage:
The modules in bsg_mesosync_io have an outdated interface with respect to their ready-handshake protocols, this commit updates the naming convention to more accurately specify the type of handshake each signal performs, if any.

There are a few signals with the ready name (line_ready_i in mesosync_core, ready_o in mesosync_output) that do not appear to actually be handshake signals at all, and so they have been left as the old convention for lack of a better name.


dpetrisko (approved):
LGTM!

PR #638, opened 2023-01-14
Bsg ready param update

BrendenPage:
Several ready-valid handshake interfaces within `bsg_dataflow` are dependent on parameters to determine the type of handshake is represented by the I/O of the module, this will update all instances of modules with a `ready_THEN_valid_p` parameter to represent the ready signals as `ready_param_o/i` to make the users more aware of the choice of handshake they make.

[Issue #1142 in Black Parrot](https://github.com/black-parrot/black-parrot/issues/1142)


bsg_dataflow/bsg_fifo_1r1w_small_hardened.v
```diff
@@ -1,141 +1,141 @@
-//
```
dpetrisko, line 1: can you see why this is showing up as a huge diff? Maybe line ending incompatibility?

dpetrisko:
LGTM, if we can squash the line ending diffs

PR #646, opened 2023-02-09
Add xilinx hardened bsg_mux and bsg_launch_sync_sync

Yuan-Mao:
This PR adds hardened bsg_mux and bsg_launch_sync_sync for Xilinx.

bsg_mux uses BUFGMUX_CTRL to select between different clocks sources without glitches whereas bsg_launch_sync_sync uses (* ASYNC_REG *) to prevent the tool from optimizing the launch flop and the 2 sync flops.

bsg_mux with els_p = 4 and bsg_launch_sync_sync have been tested on FPGA (pynq Z2).

### DIFF
--- a/hard/xilinx/7_series/bsg_misc/bsg_mux.v
+++ b/hard/xilinx/7_series/bsg_misc/bsg_mux.v
@@ -1,0 +1,71 @@
+`include "bsg_defines.v"
+
+(* KEEP_HIERARCHY = "TRUE" *)
+module bsg_mux #(
+    parameter `BSG_INV_PARAM(width_p)
+  , els_p=1
+  , harden_p = 0
+  , balanced_p = 0
+  , lg_els_lp=`BSG_SAFE_CLOG2(els_p)
+)
+(
+    input  [els_p-1:0][width_p-1:0] data_i
+  , input  [lg_els_lp-1:0] sel_i
+  , output [width_p-1:0] data_o
+);
+
+if ((harden_p == 1) && (balanced_p == 1)) begin : fi
+  for (genvar j = 0; j < width_p; j=j+1) begin: rof
+    if (els_p == 2) begin : els2
+      BUFGMUX_CTRL bufgmux_ctrl (
+        .I0(data_i[0][j])
+        ,.I1(data_i[1][j])
+        ,.O(data_o[j])
+        ,.S(sel_i)
+      );
+    end
+    else if (els_p == 4) begin : els4
+      logic [1:0][width_p-1:0] data_int;
+      BUFGMUX_CTRL bufgmux_ctrl1 (
+        .I0(data_i[0][j])
+        ,.I1(data_i[1][j])
+        ,.O(data_int[0][j])
+        ,.S(sel_i[0])
+      );
+      BUFGMUX_CTRL bufgmux_ctrl2 (
+        .I0(data_i[2][j])
+        ,.I1(data_i[3][j])
+        ,.O(data_int[1][j])
+        ,.S(sel_i[0])
+      );
+      BUFGMUX_CTRL bufgmux_ctrl3 (
+        .I0(data_int[0][j])
+        ,.I1(data_int[1][j])
+        ,.O(data_o[j])
+        ,.S(sel_i[1])
+      );
+    end
+  end
+  // synopsys translate_off
+  if(els_p != 2 && els_p != 4)
+    $error("%m error: unsupported els_p for hardened balanced bsg_mux");
+  // synopsys translate_on
+end
+else begin : nofi
+  if (els_p == 1) begin
+    assign data_o = data_i;
+    wire unused = sel_i;
+  end
+  else begin
+    assign data_o = data_i[sel_i];
+  end
+
+  // synopsys translate_off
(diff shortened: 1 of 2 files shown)


dpetrisko:
For the synchronizers, follow the pattern in TSMC28
https://github.com/bespoke-silicon-group/basejump_stl/blob/master/hard/tsmc_28/bsg_async/bsg_sync_sync_unit.v
This makes it easier to grab instances with get_cells in the backend

dpetrisko:
I would also keep in ultrascale+ directory instead of 7-series, but @gaozihou and @taylor-bsg may have opinions

gaozihou:
> I would also keep in ultrascale+ directory instead of 7-series, but @gaozihou and @taylor-bsg may have opinions

Yeah we could keep the modules in ultrascale+ directory for this PR (for simplicity). Both BUFGMUX_CTRL and ASYNC_REG would work in ultrascale+ as well.

hard/xilinx/common/bsg_async/bsg_launch_sync_sync.v
```diff
@@ -0,0 +1,358 @@
...
+   , output [bits-1:0] oclk_data_o                                      \
+   );                                                                   \
+                                                                        \
+ `rp_group    (blss_bank)                                               \
+ `rp_place    (hier blss_launch_1 0 0)                                  \
+ `rp_place    (hier blss_1   1 0)                                       \
+ `rp_place    (hier blss_2   2 0)                                       \
+ `rp_endgroup (blss_bank)                                               \
+                                                                        \
+   (* ASYNC_REG = "TRUE" *)                                             \
```
gaozihou, line 52: I would suggest applying DONT_TOUCH instead of ASYNC_REG on the launch flop.

According to UG912 (https://docs.xilinx.com/r/en-US/ug912-vivado-properties/ASYNC_REG), ASYNC_REG modifies the register to output the last known value even though a timing violation occurs. ASYNC_REG also improve mean time between failure (MTBF) for registers that can go metastable. Applying this property on launch flop could hide timing violations in simulation, and the MTBF improvement is unnecessary (launch flop should never go metastable)

hard/xilinx/7_series/bsg_misc/bsg_mux.v
```diff
@@ -0,0 +1,71 @@
+`include "bsg_defines.v"
+
+(* KEEP_HIERARCHY = "TRUE" *)
+module bsg_mux #(
```
gaozihou, line 4: Shouldn't we rename this module as bsg_clk_mux or something similar? The BUFGMUX_CTRL in the FPGA is within the clock distribution network and cannot be used by fabric logics. We are already using bsg_mux in many of our designs, I guess hardening bsg_mux in Vivado will cause other problems..

hard/xilinx/7_series/bsg_misc/bsg_mux.v
```diff
@@ -0,0 +1,71 @@
...
+    if (els_p == 2) begin : els2
+      BUFGMUX_CTRL bufgmux_ctrl (
+        .I0(data_i[0][j])
+        ,.I1(data_i[1][j])
+        ,.O(data_o[j])
+        ,.S(sel_i)
+      );
+    end
+    else if (els_p == 4) begin : els4
+      logic [1:0][width_p-1:0] data_int;
```
gaozihou, line 28: Should be `logic [1:0] data_int;`, or we could move it outside the for loop

hard/xilinx/7_series/bsg_misc/bsg_mux.v
```diff
@@ -0,0 +1,71 @@
...
+      BUFGMUX_CTRL bufgmux_ctrl3 (
+        .I0(data_int[0][j])
+        ,.I1(data_int[1][j])
+        ,.O(data_o[j])
+        ,.S(sel_i[1])
+      );
+    end
+  end
+  // synopsys translate_off
+  if(els_p != 2 && els_p != 4)
```
gaozihou, line 50: Maybe we could still `assign data_o = data_i[sel_i]` even though els_p is unsupported?

hard/xilinx/7_series/bsg_misc/bsg_mux.v
dpetrisko: I guess hardened && balanced implies clock mux in our designs? But I agree it probably makes sense to have a separate module to be more clear about the intention. @taylor-bsg want to weigh in?

hard/xilinx/7_series/bsg_misc/bsg_mux.v
Yuan-Mao: I think one benefit of doing this is that it forces users to fix this problem. Otherwise people might easily miss the warning.

dpetrisko: Yeah I agree with @Yuan-Mao, the user must have selected hardened/balanced intentionally, so failing silently is not ideal. Would also take an $error assertion

Yuan-Mao:
Now this PR only adds bsg_launch_sync_sync to ultrscale+.

dpetrisko (approved):
LGTM!

PR #648, opened 2023-02-21
Adding SURELOG to list of synthesis tools

dpetrisko:
We could add this flag in preparation for -DSURELOG, or wait for it to be added upstream. Won't break anything in the meantime so I'd suggest merging early.


dpetrisko:
Update: Surelog has the -DSURELOG macro, let's merge

taylor-bsg (approved):

PR #651, opened 2023-03-17
Adding tag constraints

dpetrisko:
This PR adds a generic routine for constraining bsg_tag interfaces.

In addition to this constraint, users must add CDC between bsg_tag and other domains. But I believe this is outside of the scope of these constraints

### DIFF
--- a/hard/common/bsg_tag/tcl/bsg_tag.constraints.tcl
+++ b/hard/common/bsg_tag/tcl/bsg_tag.constraints.tcl
@@ -1,0 +1,40 @@
+
+# This process constraints a bsg_tag interface, which typically has three pins:
+#    bsg_tag_clk : The source synchronous clock
+#    bsg_tag_data: Center-aligned data
+#    bsg_tag_en  : Out-of-band signal enabling the tag interface. It is illegal to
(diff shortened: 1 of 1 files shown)


stdavids (commented):
Is there a reason that we call it "attach" in the code and "en" in comments?

dpetrisko:
I think the reason was to also include all “slow” tag signals like tag_node_id_offset

taylor-bsg:
@dpetrisko note that en_i has been removed from the interface. Since clk_i/data_i is a source-synchronous interface, it should have constraints in both directions, constraining how much clk_i and data_i may shift relative to each other.

dpetrisko:
ACK on removing en_i, though I think having side channels in the constraints is still a good idea.

I pulled this from other tapeouts (https://github.com/bespoke-silicon-group/hb_bigblade/blob/e79dc29bd24be619abf87a73db6d4d48ca371bd9/bigblade_clk_gen/tcl/bsg_tag.constraints.tcl#L4), I think the justification was that we're overconstraining so much that treating the pad as a system synchronous source is okay. But @stdavids or @gaozihou might know better why we did it like this?

PR #652, opened 2023-03-23
Add en_i port to bsg_tag_master_decentralized

dpetrisko:
This could be useful for independently configuring multiple identical IP blocks, among other things


dpetrisko:
@gaozihou can you take a quick look at this? Thanks!

dpetrisko:
We removed en_i from bsg_tag_master, so this is no longer relevant

PR #653, opened 2023-04-13
Added an optimized popcount TDM generator to bsg_misc

klundblad:
This is a python script that generates a popcount .v file. It attempts to do a more optimized adder tree using the Three Dimensional Method (TDM). The following PPA analysis compares the original bsg_popcount.v to the generated popcount_tdm.v for widths 8, 32, 64, 128, and 1024. 

image (3)

The full PPA spreadsheet can be found here: [https://docs.google.com/spreadsheets/d/1FjDbuVNZ45XHxjTXhP-kz8AH1CuP96d7sxJkHXT-KGA/edit?usp=sharing](url)

### DIFF
--- a/bsg_misc/bsg_generate_popcount.py
+++ b/bsg_misc/bsg_generate_popcount.py
@@ -1,0 +1,259 @@
+#!/usr/bin/env python3
+import math
+
+fid = None
+twire_ctr = 0
+
+
+
+############################################################
+# BIT
+#
+class BIT:
+
+  def __init__(self, name, delay):
+    self.name = name
+    self.delay = delay
+
+  def __str__(self):
+    return f'{self.name}'
+  
+  def __repr__(self):
+    return f'{self.name}'
+
+
+
+############################################################
+# WIRE initializes an array logic for a wire 
+#
+def WIRE( name=None, dim=(1,), inst=True, delay=0.0 ):
+  global twire_ctr             # wire counter
+  if not name:
+    name = f't{twire_ctr}_n'   # creates wire variable in verilog
+    twire_ctr += 1           
+
+  if inst:
+    fid.write(f'wire ')        
+    for d in dim:
+      fid.write(f'[{d-1}:0]')  # initializes size of logic inverilog
+    fid.write(f' {name};\n')
+
+  if len(dim) == 1:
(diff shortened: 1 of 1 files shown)


bsg_misc/bsg_generate_popcount.py
```diff
@@ -0,0 +1,259 @@
...
+  num_columns_p = (math.ceil(math.log2(width_p+1)))                                   # number of columns
+  (x_i,s_o,c_o,sum_o) = print_begin_module( 'popcount_tdm', width_p, num_columns_p )  # creates and writes "popcount_tdm.v"
+  columns = [[] for i in range(num_columns_p)]                                        # create empty columns
+  bits_lp = WIRE("x_i", dim=(width_p,), inst=False)                                   # initialize array for input binary
+  columns[0] = [bits_lp[i] for i in range(width_p)]                                   # set elements of input binary to first column
+  TDM(columns, s_o, c_o, sum_o)                                                       # call TDM 
+  print_end_module( 'popcount_tdm', s_o, c_o)                                         # calculate sum in verilog file
+
+if __name__ == '__main__':
+  popcount_gen(8) # generates a "popcount_tdm.v" file for width of 8
```
stdavids, line 258: Let's expose the width as a command line argument

stdavids (commented):
Let's rename the script `bsg_popcount_tdm.py` and make the generated module name also `bsg_popcount_tdm`.

The round robin arb does something similar, we can take some inspiration from:
https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_misc/bsg_round_robin_arb.py
https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_misc/bsg_round_robin_arb.v

The arb script  takes the max size as an argument and then from 1 to the max size it generates a single module that uses generate-if statements to select the generated module. Rather than sweeping all widths from 1 to max size, maybe we just take all of the widths on the command line and generate those -- so the usage would be something like `python3 ./bsg_popcount_tdm.py 32 64 1000 > bsg_popcount_tdm.v` and it would generate a single modules with 3 generate-if statements.

taylor-bsg:
Let's try to get to the bottom of why we are not winning on frequency, especially for popcount <= 32.

taylor-bsg:
You can compare the two netlists and also the STA reports.

PR #655, opened 2023-05-24
bsg_rom_param

dpetrisko:
This PR adds a ROM module where the contents are set by parameter. This is useful for dynamically generated ROMs where the parameterization is derived from Verilog


dpetrisko:
Closing for staleness

PR #662, opened 2023-09-03
bsg_mem_1r1w_one_hot_write_mask_bit/byte intial commit

RicoLi424:
add write mask bit/byte feature to 1r1w one hot mem

### DIFF
--- a/bsg_mem/bsg_mem_1r1w_one_hot_mask_write_bit.v
+++ b/bsg_mem/bsg_mem_1r1w_one_hot_mask_write_bit.v
@@ -1,0 +1,88 @@
+// 1 read-port, 1 write-port ram with a onehot address scheme
+// write mask bits
+// reads are asynchronous
+//
+
+`include "bsg_defines.v"
+
+module bsg_mem_1r1w_one_hot_mask_write_bit #(parameter `BSG_INV_PARAM(width_p)
+                                           , parameter `BSG_INV_PARAM(els_p)
+
+                                           , parameter safe_els_lp=`BSG_MAX(els_p,1)
+                                           )
+   (input   w_clk_i
+    // Currently unused
+    , input w_reset_i
+
+    // one or zero-hot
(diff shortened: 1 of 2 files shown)


bsg_mem/bsg_mem_1r1w_one_hot_mask_write_bit.v
```diff
@@ -0,0 +1,88 @@
...
+
+  wire unused0 = w_reset_i;
+
+  logic [safe_els_lp-1:0][width_p-1:0] data_r;
+
+  logic [`BSG_SAFE_CLOG2(safe_els_lp)-1:0] w_addr_lo;
+
+  logic [width_p-1:0] data_n;
+  
+  bsg_encode_one_hot #(.width_p(safe_els_lp)
```
dpetrisko, line 36: don't need the encoder here, the enable signal in the mem_array for loop will handle the selection. We can just nest the write mask loop in the mem_array loop

bsg_mem/bsg_mem_1r1w_one_hot_mask_write_byte.v
```diff
@@ -0,0 +1,73 @@
...
+
+  logic [width_p-1:0] w_mask_expanded_lo;
+  
+  bsg_expand_bitmask #(.in_width_p(mask_width_lp)
+                      ,.expand_p(8)
+                      ) mask_expand
+                      (.i(w_mask_i)
+                      ,.o(w_mask_expanded_lo)
+                      );
+
```
dpetrisko, line 38: Preferred to use this strategy of replicating non-masked rams: https://github.com/bespoke-silicon-group/basejump_stl/blob/master/bsg_mem/bsg_mem_1r1w_sync_mask_write_byte_synth.v

RicoLi424:
Used the previously unused reset input port in bsg_mem_1r1w_one_hot and bsg_mem_1r1w_one_hot_mask_write_bit/byte

PR #664, opened 2023-09-08
fix the issue that max_val_p cannot be set to 2^n

RicoLi424:


taylor-bsg (approved):

taylor-bsg:
I mixed this in the main branch, but used BSG_WIDTH instead of BSG_CLOG2(x+1)

PR #669, opened 2024-01-18
Update ascii to rom to work with python 2/3

dpetrisko:
Currently the script will only work with python 2, not 3. This change allows it to work with both versions


tommydcjung (approved):

dpetrisko:
Thanks!

PR #675, opened 2024-03-11
bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.sv: Fix variable order

### LINKED ISSUE #674: bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.sv: Variable used before declaration
The variable `addr_li` is referenced before declaration in bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.sv. While in principle an HDL shouldn't have any problem with this, some compilers don't like it very much. Causes compile errors in Questa/Modelsim and Synopsys DC (and possibly others as well).

https://github.com/bespoke-silicon-group/basejump_stl/blob/2a7310619aaf47362be74fcd34125a99b82cd388/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.sv#L65-L69

The easy solution is to move L69 up to before L65. This resolves errors in both tools.

infinitymdm:
Reorder variable declarations so that addr_li is not referenced before its declaration. Fixes #674

### DIFF
--- a/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.sv
+++ b/bsg_mem/bsg_mem_1rw_sync_mask_write_bit_from_1r1w.sv
@@ -62,11 +62,10 @@ module bsg_mem_1rw_sync_mask_write_bit_from_1r1w #(
     );
 
   wire [addr_width_lp-1:0]  w_addr_r;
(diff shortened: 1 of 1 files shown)


dpetrisko (approved):
LGTM

PR #676, opened 2024-03-11
bsg_dmc/bsg_dmc_dly_line_v3.sv: Fix missing include

infinitymdm:
Add a missing include to bsg_dmc/bsg_dmc_dly_line_v3.sv. Without this line we get errors when trying to run simulation since the file uses some of the macros defined in bsg_misc/bsg_defines.sv.


dpetrisko:
LGTM

dpetrisko (approved):

PR #689, opened 2024-08-22
Removed all synopsys translate_off/on pragmas

sifferman:
Hello!

According to the IEEE 1364.1-2005 spec 6.3.2, the use of `translate_off`/`translate_on` is discouraged as opposed to `` `ifndef SYNTHESIS ``. Plus, some newer SystemVerilog frontends such as sv2v, Verible, and PySlint do not have any extra AST support for `translate_off`, so they may get confused by unsupported unsynthesizable constructs.

Therefore, I changed 200 occurrences of `translate_off` to `` `ifndef SYNTHESIS ``.

I would be happy to hear another perspective if you feel that `translate_off` is in fact the better option.

Thanks!

### DIFF
--- a/bsg_mem/bsg_mem_2rw_sync_mask_write_bit_synth.sv
+++ b/bsg_mem/bsg_mem_2rw_sync_mask_write_bit_synth.sv
@@ -74,6 +74,7 @@ module bsg_mem_2rw_sync_mask_write_bit_synth #( parameter `BSG_INV_PARAM(width_p
         else
             b_addr_r <= 'X;
 
+`ifndef BSG_HIDE_FROM_SYNTHESIS
         // if addresses match and this is forbidden, then nuke the read address
 
         if (a_addr_i == b_addr_i && a_v_i && b_v_i && (a_w_i || b_w_i) && !read_write_same_addr_p)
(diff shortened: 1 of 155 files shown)


dpetrisko:
Seems reasonable to me. @taylor-bsg 

I would suggest to future-proof we use:

ifndef BSG_HIDE_FROM_SYNTHESIS

where in bsg_defines.sv

ifdef SYNTHESIS
'define BSG_HIDE_FROM_SYNTHESIS 
endif 

Which will let us have per-tool logic in the future if we need to

taylor-bsg:
@sifferman thanks for your contribution Ethan.  Would you be able to modify the pull request as suggested by @dpetrisko? @dpetrisko would you be able to verify that BP and bsg_manycore are cool with this?

bsg_mem/bsg_mem_2rw_sync_mask_write_bit_synth.sv
```diff
@@ -74,6 +74,7 @@ module bsg_mem_2rw_sync_mask_write_bit_synth #( parameter `BSG_INV_PARAM(width_p
         else
             b_addr_r <= 'X;
 
+`ifndef BSG_HIDE_FROM_SYNTHESIS
```
sifferman, line 77: I saw that this file was missing the ifndef SYNTHESIS, so I added it

sifferman:
Just made the changes!

taylor-bsg (approved):

taylor-bsg:
TY!

sifferman:
Thanks!

PR #693, opened 2024-09-06
Fix Verilator Warnings in bsg_mem and bsg_misc

infinitymdm:
Non-functional changes in bsg_mem and bsg_misc to fix Verilator lint warnings and 1 syntax error.

- `bsg_mem/bsg_mem_multiport_latch_write_banked_bypassing.sv`: Fix syntax error due to mismatched `end`. Looks like it should have been commented out with the rest of the block above it.
- `bsg_misc/bsg_mul.sv`: Fix incorrect port name. Fixes Verilator `PINMISSING` warning.
- `bsg_mem/bsg_mem_2rw_sync.sv`: Fix misleading indentation of `assert ... else` blocks. Fixes Verilator `MISINDENT` warnings.
- `bsg_mem/bsg_mem_1r1w.sv`: Cast logic to int before comparison with parameter. Fixes Verilator `WIDTHEXPAND` warning.
- `bsg_misc/bsg_counter_overflow_en.sv`: Cast logic to int before comparison with parameter. Fixes Verilator `WIDTHEXPAND` warning.

### DIFF
--- a/bsg_mem/bsg_mem_1r1w.sv
+++ b/bsg_mem/bsg_mem_1r1w.sv
@@ -46,7 +46,7 @@ module bsg_mem_1r1w #(parameter `BSG_INV_PARAM(width_p)
    always_ff @(negedge w_clk_i)
      if (w_v_i===1'b1)
        begin
-         assert ((w_reset_i === 'X) || (w_reset_i === 1'b1) || (w_addr_i < els_p) || (els_p <= 1))
+         assert ((w_reset_i === 'X) || (w_reset_i === 1'b1) || (int'(w_addr_i) < els_p) || (els_p <= 1))
             else $error("Invalid address %x to %m of size %x (w_reset_i=%b, w_v_i=%b)\n", w_addr_i, els_p, w_reset_i, w_v_i);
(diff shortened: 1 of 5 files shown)


taylor-bsg:
The Verilator warnings are an interesting case. I am not sure about casting the values to ints since they are 32-bit signed.

taylor-bsg:
ChatGPT:

2. **Type Conversion and Overflow Logic**:
   - The line `assign overflow_o = (int'(count_o) == max_val_p);` uses `int'()` type casting, which is likely not necessary and could be replaced with `ptr_width_lp'()` to match the width of `count_o`:
     ```systemverilog
     assign overflow_o = (count_o == ptr_width_lp'(max_val_p));
     ```
   - This avoids unnecessary type conversions and potential mismatches.

infinitymdm:
I'm thinking the clearest way to do this will be to apply the `BSG_WIDTH` macro for casting in both files where there's an affected comparison. That seems clearer to me than using `max_val_p`, at least (though I admit it's a little less efficient for the preprocessor), and provides a way to do this in both files with the `WIDTHEXPAND` warning. Expect an updated PR shortly.

infinitymdm:
The latest changes are in line with your suggestions, but I worry a bit about the safety of shrinking `els_p` rather than expanding `w_addr_i` in bsg_mem_1r1w. Especially with the new UNSIGNED warning.

PR #698, opened 2024-10-18
2-D Torus NoC

tommydcjung:


taylor-bsg:
Seems like you should be able to have bsg_virtual_channel_router, and then torus should just be an outer level wiring thing. Then we can reuse this for all things virtual channel.

PR #699, opened 2024-10-23
SDR Token Clock Timing

gaozihou:
Add token boundary cells to improve token clock driving strength
Update token clock timing constraints, add token clock driving cell


dpetrisko (approved):
LGTM! These changes work in genus as well

PR #703, opened 2024-11-11
Adding static library target

dpetrisko:
For some use cases, static libraries are better


dpetrisko:
closing for lack of interest

PR #705, opened 2024-11-15
Update bsg_dlatch.sv

dpetrisko:
Latches are recommended to use blocking assignments as they follow similar rules to always_comb.  This doesn't affect the functionality here, but disable warnings in Verilator and Xcelium


dpetrisko:
I've found literature arguing both ways, so closing until someone tells me what to think :)

PR #707, opened 2024-11-18
Port bsg_cache to verilator

dpetrisko:
Only minor changes now that verilator supports timing statements. This PR only fixes basic_32, but would be trivial to port other tests.

- resolve blocking / non-blocking mixture in checker. Should all be non-blocking
- Add USE_VERILATOR flag to switch between VCS/Verilator
- Move to bsg_trace_replay from deprecated bsg_fsb_node_trace_replay
- Adding -DWAVE definition to allow sims to be compiled w/wo dump for speed
- rename testbench.checker to testbench._checker to avoid keyword conflict (checker is a SV keyword. VCS allows this override but Verilator does not, which is probably the sane thing)
- Workaround for Verilator 5.030 regression https://github.com/verilator/verilator/pull/5616

### DIFF
--- a/testing/bsg_cache/regression_v2/testbench.sv
+++ b/testing/bsg_cache/regression_v2/testbench.sv
@@ -34,14 +34,21 @@ module testbench();
 
   integer status;
   integer wave;
-  string checker;
+  string _checker;
   initial begin
     status = $value$plusargs("wave=%d",wave);
-    status = $value$plusargs("checker=%s",checker);
-    $display("checker=%s", checker);
+    status = $value$plusargs("checker=%s",_checker);
+    $display("checker=%s", _checker);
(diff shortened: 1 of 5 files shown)


testing/bsg_cache/regression_v2/testbench.sv
```diff
@@ -34,16 +34,23 @@ module testbench();
 
   integer status;
   integer wave;
-  string checker;
+  string _checker;
```
dpetrisko, line 37: "checker" is systemverilog keyword

dpetrisko:
@tommydcjung  can you take a quick look at this? Thanks!

tommydcjung (approved):
Looks good thanks!

PR #708, opened 2024-11-22
Making bsg_fifo_1r1w_narrowed harden-able

dpetrisko:
Adding parameter to outer module for usage with inner FIFO


dpetrisko:
@stdavids can you take a quick look at this? Thanks!

stdavids (approved):

PR #717, opened 2025-02-18
Update memory generator script to Python 3

LRitzdorf:
This didn't actually require any substantive code changes.
The original so-called Python 2 code was already using the `print()` function (as opposed to Python 2's `print` keyword), so we don't even have to touch that!

Pairs with https://github.com/bsg-external/ee477-hammer-cad/pull/3


LRitzdorf:
Looks like Python 3 does some string formatting that prints numbers with decimal points, which is very much not allowed in Verilog module names. One moment...

LRitzdorf:
Dang Python 2 floor division...

dpetrisko:
I think probably we should not specify #!/usr/bin/env python3. ubuntu by default calls its python version "python" despite it being python3

LRitzdorf:
Uhh, where are you seeing that? I just tested both Debian and Ubuntu systems, and both call their Python 3s, well, `python3` (without even a symlink for `python`). This also matches my testing on our AlmaLinux 8 servers.

In addition, the general guidance is that you use `python` for programs that run under both Python 2 and 3, or explicitly specify a version when the program does care.

Specifying `python3` works with this guidance, while also ensuring that the script runs on our workstations.

dpetrisko:
Whoops, remembered it backwards. Ubuntu removed “python” and only supports python3, so that change should be fine

buhraian:
I have a commit up on my fork that updates bsg_ascii_to_rom and bsg_cache_dma_trace_gen to work with python3, but I realized that this PR points to a branch that already contains an updated bsg_ascii_to_rom. I was wondering if uw_e477_wi23 was going to be merged into master and what I should do with my small fix (i.e. keep it local or try to PR it in).

dpetrisko:
Oh, I didn't see this was to non-master, I'll go ahead and merge it. Looks like we have this change on master in #647 

>I was wondering if uw_e477_wi23 was going to be merged into master

It's unlikely that it'd be done directly, but if there are any changes you think would be general, feel free to open a PR to master

dpetrisko:
moving fast and breaking things, I merged that one too. This comment is free license for EE477 TAs to yell at me about it

LRitzdorf:
fwiw, I got away with using these changes for the second half(ish) of last quarter, so we should be in the clear :)

PR #722, opened 2025-03-28
Add Verilator support to Makefile.sim

buhraian:
Small changes to enable Verilator as an option in Makefile.sim, checked with `Verilator 5.034 2025-02-24 rev v5.034-47-gac3f30ed6`. Included guards for `define in various test_bsg.sv files since Verilator and VCS have different behaviors when trying to override macro definitions using +define+ in the command line.

### DIFF
--- a/testing/Makefile.sim
+++ b/testing/Makefile.sim
@@ -19,7 +19,19 @@ BSG_IP_CORES_SIM = ./simv
 
 BSG_IP_CORES_SIM_DEFINE_PARAM = +define+
 BSG_IP_CORES_SIM_INCLUDE_DIRS = +incdir+$(TOP)/bsg_test +incdir+$(TOP)/bsg_misc +incdir+$(TOP)/bsg_dataflow
-BSG_IP_CORES_SIM_CLEAN_FILES  = simv*
+BSG_IP_CORES_SIM_CLEAN_FILES  = simv* csrc/ ucli.key
+#
+##########
+else ifeq ($(SIMULATOR),verilator)
+##########
+# settings for Verilator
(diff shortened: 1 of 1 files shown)


dpetrisko:
Hi, thanks for the pull. What’s the define behavior for Verilator, it always takes the file value? Is that a Verilator bug or does the SV spec not take about precedence?

buhraian:
So Verilator actually does throw the below warning (https://verilator.org/guide/latest/warnings.html?highlight=redefmacro#cmdoption-arg-REDEFMACRO) when it encounters the file's macro definition. I couldn't find anything in the 2023 SV spec regarding macro definition precedence, but I did test this with VCS and Questa too and both seem to silently ignore the redefinition within the file and maintain the command line definition, FWIW.

```
%Warning-REDEFMACRO: test.sv:2:23: Redefining existing define: 'TEST_MACRO', with different value: '10'
                     ... Location of previous definition, with value: '99'
                     ... For warning description see https://verilator.org/warn/REDEFMACRO?v=5.035
                     ... Use "/* verilator lint_off REDEFMACRO */" and lint_on around source to disable this message.
```

buhraian:
Just asked this question in the Verilator repo as well, linking for visibility: https://github.com/verilator/verilator/issues/5900.

buhraian:
Just merged in a change to Verilator to allow macro overrides using ``+define+``, so I updated the PR to remove the unnecessary `` `define `` guards now.

dpetrisko:
awesome, thanks!

dpetrisko (approved):

PR #733, opened 2025-06-20
miniblade rc2 vcache code

RicoLi424:
@gaozihou @tommydcjung


dpetrisko:
Hi @RicoLi424 @tommydcjung curious what the final status of this was? Is there still more validation to be done or is it mostly code cleanup left? Was there anything miniblade-specific that we should trim out before merging?

RicoLi424:
Hi @dpetrisko I had run unit testing suite and regressions, and the tested configurations had passed, including notifications disabled.

Some upstream integration/cleanup remains to do. There was also a small fix to the notification _**way_id**_ metadata that I haven't pushed. That fix already passed targeted tests and the full mini suite as well.

The miniblade-specific part is the FPGA shadow-directory protocol (_**write_validate, way_id, evict_pending**_, and the associated **flush behavior**). I can split that into a separate PR while retaining the generic uncached-access support if you think it'd be better

dpetrisko:
Thanks for the quick response! Yes if it would be simpler to merge a smaller PR with just the uncached changes, that would be ideal for me as I’m most interested in those

taylor-bsg:
Agree!

PR #741, opened 2026-03-26
Fixing ununsized concat in BSG_WIDTH

dpetrisko:
```
credit_counter_width_p=`BSG_WIDTH(32)
...
xmvlog: *E,NONOWD (bsg_manycore/v/bsg_manycore_endpoint_standard.sv,37|44): Illegal use of a constant without an explicit width specification [4.1.14(IEEE)].
```

Basically '32' is an integer literal which is usually evaluated to 32b width but by strict IEEE is unsized. This could be patched in all usages of BSG_WIDTH by doing e.g. 32'd32, but this solution should fix the compilation error and maintain the overflow property from 0e8460329e1af3fac7536c24187b6886165c4aa6

### DIFF
--- a/bsg_misc/bsg_defines.sv
+++ b/bsg_misc/bsg_defines.sv
@@ -47,7 +47,7 @@
 // maps 0/1 --> 1 instead of to 0
 `define BSG_SAFE_CLOG2(x) ( (((x)==1) || ((x)==0))? 1 : $clog2((x)))
 `define BSG_IS_POW2(x) ( (1 << $clog2(x)) == (x))
-`define BSG_WIDTH(x) ( ((x)==0)? 1 : $clog2(x) + (`BSG_IS_POW2(x) ? 1 : 0))
+`define BSG_WIDTH(x) ( $clog2(x) + (`BSG_IS_POW2(x) ? 1 : 0))
 `define BSG_SAFE_MINUS(x, y) (((x)<(y))) ? 0 : ((x)-(y))
 
 // these "SAFE" shift functions handle a common problem that shifts default to 32 bits wide even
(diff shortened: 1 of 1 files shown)


dpetrisko:
Err, seems like this works in xcelium but not vcs. Will need some more investigation...

dpetrisko:
Okay, this seems to work in all three VCS, Xcelium and Verilator

```
`define BSG_WIDTH(x) (((x) == 0) ? 1 : ($clog2(x) + ((((x) & ((x) - 1)) == 0) ? 1 : 0)))

module test;

initial begin
    int a = 0;
    int b = 1;
    int c = 2;
    int d = 3;
    int e = 32'hffffffff;

    $display("a: %x %d\n", a, `BSG_WIDTH(a));
    $display("b: %x %d\n", b, `BSG_WIDTH(b));
    $display("c: %x %d\n", c, `BSG_WIDTH(c));
    $display("d: %x %d\n", d, `BSG_WIDTH(d));
    $display("e: %x %d\n", e, `BSG_WIDTH(e));
end

endmodule
```

...
```
a: 00000000           1
b: 00000001           1
c: 00000002           2
d: 00000003           2
e: ffffffff          32
```

dpetrisko:
confirmed same behavior
```
a: 00000000           1
b: 00000001           1
c: 00000002           2
d: 00000003           2
e: ffffffff          32
```

in verilator, vcs and xcelium

dpetrisko:
@taylor-bsg

taylor-bsg:
@dpetrisko seems okay, but since this fix is about fixing compatibility with xcelium, we should not change the semantics of BSG_WIDTH

dpetrisko:
@taylor-bsg whoops, thought 0->1 was the original behavior. This should be good to go then:

```
`define BSG_WIDTH(x) ( $clog2(x) + (`BSG_IS_POW2(x) ? 1 : 0))
...
a: 00000000           0

b: 00000001           1

c: 00000002           2

d: 00000003           2

e: ffffffff          32

```

dpetrisko:
This also fixes (non-fatal) warnings in Verilator and VCS, xcelium just has upgraded severity

taylor-bsg (approved):

PR #746, opened 2026-05-08
bsg_dmc: fix CALR ldst_tick using wrong bank state

flaviens:
Hi there!

CALR always reads bank 0, row 0, but its ldst_tick setup used the bank/row from the user-command AFIFO head. If a non-zero-bank user command was pending when calibration fired, CALR could issue ACT to an already-open bank 0 or READ to inactive bank 0.

This PR splits the CALR and LDST setup paths so CALR checks open_bank[0] / open_row[0], matching its actual target.

Thanks!
Flavien


dpetrisko:
Seems reasonable! Did you find this with one of the existing testbenches? Would like to add a test case to avoid regression.

I don't think this is related to https://github.com/bespoke-silicon-group/basejump_stl/pull/387 but could you see if your change is compatible with it?

PR #748, opened 2026-07-07
Add links_credit_late_o output to wormhole concentrator

dpetrisko:
Adding unused port to suppress warnings


dpetrisko:
can also expose it to the top-level if we feel that's useful

PR #749, opened 2026-09-06
Fix syntax error in bsg_mem_multiport_latch_write_banked_bypassing.sv

kotelone:
Fix syntax error (missing comment for `end` keyword)


taylor-bsg:
Thank you!

PR #750, opened 2026-09-08
Add BSG_SAFE_WIDTH and fix zero-maximum counter elaboration

taylor-bsg:
`bsg_counter_clear_up` currently derives a zero-bit width when `max_val_p=0`, causing zero-size casts to fail elaboration. This occurs in `bsg_cache_to_test_dram` when one DRAM request covers a cache line: its counter maximum is `num_req_lp-1=0`.

Add `BSG_SAFE_WIDTH(x)` as `max(1, BSG_WIDTH(x))` and use it for the counter's default width. This preserves `BSG_WIDTH(0)=0` and its full-range maximum handling, consistent with the semantic-compatibility decision in #741. The counter's sequencing is unchanged.

Validation on Verilator 5.050:

- Added a self-checking regression for maxima 0–8, `32'h7fffffff`, `32'h80000000`, `32'hffffffff`, and `64'hffffffffffffffff`. All 13 pass width, reset, hold, increment and clear checks; the test also checks both macros' distinct zero behavior.
- The same test with the original counter fails elaboration on zero-size casts.
- HammerBlade's 2×2, 4-way, 8-word-line HBM profile now builds execution and profiling models. Cold/warm vector_add N=4096 passes in both modes after a separate Replicant channel-count correction; the counter fix alone does not repair that profile's host DMA mapping mismatch.

Run the unit regression from `testing/bsg_misc/bsg_counter_clear_up` with:

```sh
make SIMULATOR=verilator SHELL=/bin/bash '.SHELLFLAGS=-e -o pipefail -c'
```

VCS, Xcelium and Design Compiler were not run.

### DIFF
--- a/bsg_misc/bsg_counter_clear_up.sv
+++ b/bsg_misc/bsg_counter_clear_up.sv
@@ -9,8 +9,10 @@ module bsg_counter_clear_up #(parameter `BSG_INV_PARAM(max_val_p)
 			      // this originally had an "invalid" default value of -1
 			      // which is a bad choice for a counter
 			     ,parameter init_val_p   = `BSG_UNDEFINED_IN_SIM('0)
+                             // Maximum zero still needs one bit for a legal register and casts.
+                             // The safe helper preserves BSG_WIDTH's full-range maximum handling.
                              ,parameter ptr_width_lp =
-                             `BSG_WIDTH(max_val_p)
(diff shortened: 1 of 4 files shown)


taylor-bsg:
@dpetrisko BSG_SAFE_WIDTH

